PR #11, opened 2017-06-14
Compile the js as a CommonJS module, npm loves this!

Sambego:
Npm requires all packages in it's registry to be in the `CommonJS` format. Npm is originally a registry for node modules, and node knows how to handle them. You can read more about it in the "npm-registry" section here: https://docs.npmjs.com/all

This library is currently not compiled to any module type, which causes errors when you try to unit test a component which includes this lib for example. Jest or any other test runner run in a node environment which only knows how to handle CommonJS.

Why doesn't it throw an error when I just include them in my react app? You're probably using babel, which knows how to handle `import` statements.

By removing the `{ "modules" : false }` babel will compile it to the default module standard which is CommonJS (https://babeljs.io/docs/plugins/preset-es2015/).

This should not break any changes, since babel knows how to handle them. Unless the library gets used somewhere without babel or another compiler which does not know how to handle CommonJS (do they even exist?). 

### Changed
- Switch from using no module standard to CommonJS


.babelrc
```diff
@@ -1,11 +1,9 @@
...
   "presets": [
-    ["es2015", { "modules" : false }],
-    "react",
-    "stage-1"
-  ],
-  "plugins": [
-    "transform-runtime"
+    "transform-runtime",
+    "es2015",
+    "stage-0",
```
maartenscholz: Q: are there consequences to switch to stage-0?

Sambego: Some more support for future ES features. Stage 0 includes stage 1. https://babeljs.io/docs/plugins/preset-stage-0/

nickwaelkens: I too prefer `stage-0` over `stage-1`, function binding syntax is dope.

nickwaelkens:
I see your (extremely) valid point. My understanding was tree shaking in Webpack@2 only occurred when `{ "modules" : false }` was specified in the `.babelrc` file. Have you got any idea if this will still work now, or do you think it doesn't matter as long as the parent app is compiled with that flag turned off?

Sambego:
@nickwaelkens as far as I know treeshaking happens when you use webpack with the `--optimize-minimize` flag. You can also use the production flag `-p` which is equal to `--optimize-minimize` `--optimize-occurrence-order`.

 I think using `{ "modules" : false }` in the final app is enough to transform everything to ES2015 modules, but I'm not an expert in these kind of things :)

https://webpack.js.org/guides/tree-shaking/

.babelrc
tcoopman: Be careful with using stage-0, the syntax/proposal/... can completely change: https://tc39.github.io/process-document/

I'm not saying you shouldn't use it, just know that it can have an impact later

.babelrc
```diff
@@ -1,11 +1,9 @@
 {
   "presets": [
-    ["es2015", { "modules" : false }],
-    "react",
-    "stage-1"
-  ],
-  "plugins": [
-    "transform-runtime"
+    "transform-runtime",
```
tcoopman: what's the difference between this on the presets vs in plugins?

PR #20, opened 2017-08-21
Component/radio

driesd:


components/radio/Radio.js
```diff
@@ -0,0 +1,23 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+const Radio = ({ checked, onMouseDown, theme, ...other }) => (
+  <div
+    data-react-toolbox="radio"
```
nickwaelkens: ...

components/radio/RadioButton.js
```diff
@@ -0,0 +1,112 @@
+import React, { Component } from 'react';
+import PropTypes from 'prop-types';
+import classnames from 'classnames';
+import { themr } from 'react-css-themr';
+import { RADIO } from '../identifiers';
+import Radio from './Radio';
+
+const factory = (Radio) => {
```
nickwaelkens: Are we still using `factory`?

components/radio/RadioButton.js
```diff
@@ -0,0 +1,112 @@
+import React, { Component } from 'react';
+import PropTypes from 'prop-types';
+import classnames from 'classnames';
```
nickwaelkens: Let's use `cx` like is the case everywhere else

components/index.js
```diff
@@ -20,5 +21,6 @@ export {
   Overlay,
   PopoverHorizontal,
   PopoverVertical,
+  Radio,
```
nickwaelkens: Why not import `RadioGroup`?

driesd:
The labels are already clickable

nickwaelkens:
I'm kind of confused what this component looks like. There is no `label`, just the radio button?

lowiebenoot:
@nickwaelkens RadioButton contains the label as well. Radio is just the circle.

PR #25, opened 2017-09-13
Added package ui-typography and our new typography component

driesd:
schermafdruk 2017-09-13 13 11 21

**Note:** ignore the purple link for now. _Link_ will be a separate component


components/typography/Monospaced.js
```diff
@@ -0,0 +1,42 @@
...
+const factory = (defaultElement) => {
+  class Monospaced extends PureComponent {
+    static propTypes = {
+      children: PropTypes.node,
+      className: PropTypes.string,
+      element: PropTypes.node,
+    };
+
+    static defaultProps = {
+      className: '',
```
nickwaelkens: Don't think you need to specify this empty string as default prop, undefined will also work, so you can remove it. (But check it first!)

components/typography/Heading.js
```diff
@@ -0,0 +1,43 @@
...
+const factory = (type, defaultElement) => {
+  class Heading extends PureComponent {
+    static propTypes = {
+      children: PropTypes.node,
+      className: PropTypes.string,
+      element: PropTypes.node,
+    };
+
+    static defaultProps = {
+      className: '',
```
nickwaelkens: Don't think you need to specify this empty string as default prop, undefined will also work, so you can remove it. (But check it first!) Same with element?

components/index.js
```diff
@@ -13,15 +24,24 @@ export {
...
+  Heading2,
+  Heading3,
+  Heading4,
   Menu,
+  Monospaced,
   IconMenu,
   MenuItem,
   MenuDivider,
   LoadingMolecule,
+  OldStyle,
```
nickwaelkens: Not too fond of `OldStyle` as name, it's not obvious enough. `OldStyleNumber`? Something less vague at least 👍

PR #27, opened 2017-09-13
Added a basic Link component

driesd:


components/link/Link.js
```diff
@@ -0,0 +1,54 @@
...
+    children: PropTypes.node,
+    className: PropTypes.string,
+    external: PropTypes.bool,
+    onClick: PropTypes.func,
+    url: PropTypes.string,
+  };
+
+  static defaultProps = {
+    className: '',
+    external: false,
```
nickwaelkens: I know I commented on previous PRs that having falsey values as `defaultProps` is kinda redundant, but I'm having second thoughts about it now.

WDYT @maartenscholz @lowiebenoot @svenlombaert ?

components/link/Link.js
```diff
@@ -0,0 +1,54 @@
...
+import cx from 'classnames';
+import s from './theme.css';
+
+class Link extends PureComponent {
+  static propTypes = {
+    children: PropTypes.node,
+    className: PropTypes.string,
+    external: PropTypes.bool,
+    onClick: PropTypes.func,
+    url: PropTypes.string,
```
maartenscholz: NTH: add `isRequired` for the required props

components/link/Link.js
maartenscholz: I don't mind it, especially for optional boolean props

lowiebenoot: Not really strong opinionated about it. I don't mind having them, but i also don't mind not having them.

components/link/Link.js
```diff
@@ -0,0 +1,54 @@
...
+      ...rest
+    } = this.props;
+
+    const classNames = cx(
+      s['link'],
+      className,
+    );
+
+    const target = external ? '_blank' : undefined;
+    const rel = external ? 'noopener noreferrer' : undefined;
```
lowiebenoot: Not really fond of limiting target and rel to these values. I would just not define them as props in this component, and let them sit in the `rest` so you can still pass them to the component, with any value you like.

lowiebenoot:
2 general comments/questions:

* doesn't it need more styling/options for colors? Or how are we going to style them? Also, it now looks different when you use a `button` than when you use a `a`. The `button` isn't using proxima nova, but system-ui.

* how can we now use the `<Link>` component from react-router (as we discussed yesterday).

components/link/Link.js
nickwaelkens: Maybe combine the `target` and `rel` with potential `this.props.target` and `this.props.rel`?

components/link/Link.js
```diff
@@ -0,0 +1,45 @@
...
+  static propTypes = {
+    children: PropTypes.node.isRequired,
+    className: PropTypes.string,
+  };
+
+  static defaultProps = {
+    className: '',
+  };
+
+  static use (LinkComponentToUse) {
```
lowiebenoot: This `Link.use(...)` implementation makes it impossible to have different types of `Link` components within the same component. For example: a React router link, and a normal link in the same component. I think this should be possible.

nickwaelkens: We also came to that conclusion yesterday. I just tested the following, so maybe that would work with importing Link components too;

```js
// fileA.js
export const foo = "some value";

// fileB.js
import { foo as bar, foo as baz } from './fileA';
console.log(bar); // "some value"
console.log(baz); // "some value"
```

lowiebenoot: Your test actually proofs nothing and is quite useless :D (it just proofs that you can import the same thing twice under a different name, but it will still be the same think in the end, I guess it's just a reference or something).

I just tried it out, and it doesn't work. Here's some pseudo code 💅 

```js
import { Heading1, Heading2 } from 'HeadingShizzle';
import { Link as Link1, Link as Link2 } from 'LinkShizzle';

Link1.use(Heading1);
Link2.use(Heading2);

...

<Link1 /> // will use a Heading2 (while we expected a Heading1)
<Link2 /> // will use a Heading2 (which is what we expected)
```

lowiebenoot: A factory is probably the solution?

PR #30, opened 2017-09-14
Removed react-css-themr from the components

lowiebenoot:
We decided to remove react-css-themr, because we don't need it (our designers promised us) and it was only adding extra complexity.


lowiebenoot:
Yes indeed, this will be bumped to a different minor version. But I will do this in a separate PR, once we will actually publish. It's not really a part of this PR imo.

PR #31, opened 2017-09-14
StatusLabel component

driesd:
schermafdruk 2017-09-14 11 05 03

### DIFF
--- a/components/statusLabel/StatusLabel.js
+++ b/components/statusLabel/StatusLabel.js
@@ -1,0 +1,78 @@
+import React, { Component } from 'react';
+import PropTypes from 'prop-types';
+import cx from 'classnames';
+import omit from 'lodash.omit';
+import theme from './theme.css';
+
+class StatusLabel extends Component {
+  static propTypes = {
+    children: PropTypes.node,
+    className: PropTypes.string,
+    mint: PropTypes.bool,
+    teal: PropTypes.bool,
+    violet: PropTypes.bool,
+    ruby: PropTypes.bool,
+    gold: PropTypes.bool,
+    aqua: PropTypes.bool,
+    dark: PropTypes.bool,
+    small: PropTypes.bool,
+  };
+
+  getColor () {
+    const colors = [
+      'mint',
+      'violet',
+      'ruby',
+      'gold',
+      'aqua',
+      'teal',
+    ];
+
+    for (var i = 0; i < colors.length; i++) {
+      const color = colors[i];
+      if (this.props[color]) {
+        return color;
+      }
+    }
+  }
+
+  getSize () {
+    return this.props.small ? 'small' : 'medium';
+  }
+
+  render () {
+    const {
+      children,
+      className,
+      ...others
+    } = this.props;
+
+    const color = this.getColor();
+    const size = this.getSize();
+
+    const classes = cx(
+      theme.label,
+      theme[color],
+      theme[size],
+      className,
+    );
+
+    const rest = omit(others, [
+      'mint',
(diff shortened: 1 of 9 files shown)


components/statusLabel/StatusLabel.js
```diff
@@ -0,0 +1,77 @@
...
+      'aqua',
+      'teal',
+    ];
+
+    for (var i = 0; i < colors.length; i++) {
+      const color = colors[i];
+      if (this.props[color]) {
+        return color;
+      }
+    }
```
svenlombaert: Doesn't something like this work?

```js
color = colors.filter(color => this.props.hasOwnProperty(color))
return color[0]
```

nickwaelkens: It's starting to get copy/pasted all over the place too, time to extract? :)

lowiebenoot: I actually don't link the filter, because it can't return "early", it keeps evaluating even after it found a match.
Extracting it probably wouldn't hurt indeed.

components/statusLabel/StatusLabel.js
```diff
@@ -0,0 +1,73 @@
...
+      'mint',
+      'violet',
+      'ruby',
+      'gold',
+      'aqua',
+      'teal',
+    ];
+
+    const color = colors.filter(color => this.props.hasOwnProperty(color));
+    return color[0];
```
lowiebenoot: As I already commented somewhere else, i don't like the filter because you can't return "early", it keeps evaluating the rest even if we already found a match.

Also, `hasOwnProperty` will return `true`, even if the prop is `false` 👎

components/statusLabel/theme.css
```diff
@@ -0,0 +1,57 @@
...
+  line-height: calc(1.8 * var(--unit));
+  padding: var(--spacer-smallest) var(--spacer-smaller);
+}
+
+.small {
+  font-size: calc(1.2 * var(--unit));
+  font-variant: small-caps;
+  line-height: calc(1.8 * var(--unit));
+  padding: 0 var(--spacer-smaller);
+  text-transform: uppercase;
```
lowiebenoot: I actually don't see the `text-transform` and `font-variant` in the design files.

components/statusLabel/StatusLabel.js
svenlombaert: Fair point, we could use `array.find` which returns on the first true evaluation.

And I guess the conditional changes to `this.props.hasOwnProperty(color) && this.props.color`

We could also ditch the hasOwnProperty check if we set default propTypes to false. It just that I had to triple look at the if to finally grasp what was going on.

svenlombaert: Implementation would change to:

`return colors.find(color => this.props.hasOwnProperty(color) && this.props.color);`

lowiebenoot: Yours won't be correct though (the `this.props.color`) and I think we can just drop the hasOwnProperty, because undefined evaluates to false.

`return colors.find(color => this.props[color]);`

components/statusLabel/theme.css
lowiebenoot: I talked to Stephen, and you already discussed this apparently and decided to go with the uppercase and small-caps. So ignore my comment 👍

driesd: The design in Zeplin has been updated by @stephen-tl. But @lowiebenoot you're right about the text-transform which isn't needed in this case. We will just use the **font-variant: small-caps;** to get the right result.

PR #35, opened 2017-09-15
Counter component

driesd:
schermafdruk 2017-09-19 10 30 36

**Out of scope:** Tooltip component

### DIFF
--- a/components/counter/Counter.js
+++ b/components/counter/Counter.js
@@ -33,7 +33,7 @@ class Counter extends PureComponent {
       theme[color],
       theme[size],
       {
-        [theme.rounded]: count,
+        [theme.rounded]: typeof (count) !== 'undefined',
       },
       className,
     );
--- a/components/counter/Counter.js
+++ b/components/counter/Counter.js
@@ -1,7 +1,6 @@
 import React, { PureComponent } from 'react';
 import PropTypes from 'prop-types';
 import cx from 'classnames';
-import omit from 'lodash.omit';
 import { Monospaced } from '../typography';
 import theme from './theme.css';
 
@@ -38,13 +37,8 @@ class Counter extends PureComponent {
       className,
     );
 
-    const rest = omit(others, [
-      'color',
-      'size',
-    ]);
-
     return (
-      <span className={classes} {...rest} data-teamleader-ui="counter">
+      <span className={classes} {...others} data-teamleader-ui="counter">
         <Monospaced>{children}</Monospaced>
       </span>
     );
--- a/components/counter/Counter.js
+++ b/components/counter/Counter.js
@@ -1,7 +1,6 @@
 import React, { PureComponent } from 'react';
 import PropTypes from 'prop-types';
 import cx from 'classnames';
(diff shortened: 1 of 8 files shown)


components/counter/theme.css
```diff
@@ -0,0 +1,35 @@
...
+}
+
+.medium {
+  min-height: 9px;
+  min-width: 9px;
+}
+
+.small {
+  min-height: 6px;
+  min-width: 6px;
```
nickwaelkens: **PP** Can we use css4 variables here instead? 
Example here: https://github.com/teamleadercrm/teamleader-ui/blob/button-v2/components/button/theme.css

components/counter/Counter.js
```diff
@@ -0,0 +1,54 @@
...
+      children,
+      className,
+      color,
+      size,
+      ...others
+    } = this.props;
+
+    const classes = cx(
+      theme.counter,
+      theme.monospaced,
```
lowiebenoot: Seems like a leftover? The class isn't doing anything as far as I can see (although it actually has to be monospaced).

components/counter/Counter.js
```diff
@@ -0,0 +1,54 @@
...
+      theme[color],
+      theme[size],
+      {
+        [theme.circular]: !children,
+        [theme.rounded]: children,
```
lowiebenoot: I would just make the css rounded by default, and only add the `circular` "variant" when necessary. This way you can drop the rounded class, and you end up with just circular being a variant.

components/counter/Counter.js
```diff
@@ -0,0 +1,49 @@
...
+      size,
+      ...others
+    } = this.props;
+
+    const classes = cx(
+      theme.counter,
+      theme[color],
+      theme[size],
+      {
+        [theme.rounded]: count,
```
lowiebenoot: If the count is `0`, it will not show the rounded variant.

lowiebenoot:
I'm also having some doubts about the color prop. 
Is the color functional as in e.g. active/not-active? Or is it really just as a color, and we're limiting it to 2 colors? If it's the 2nd one, we can keep it as a color prop, if it has some functionality, I would change it so active/non-active (or a better suitable name).

stephen-tl:
I see this as functional color: active/Inactive. One thing that I will add to the spec in zeplin is a negative version of the inactive state. This can then be used in things like the sidenav. you can expect an update in 15 minutes

stephen-tl:
After discussing this, we decided to make it possible to use multiple colors as a background. This to ensure flexibility later on. I'll update the spec later today.

components/counter/Counter.js
```diff
@@ -0,0 +1,57 @@
...
+      theme[color],
+      theme[size],
+      {
+        [theme.dark]: dark,
+        [theme.rounded]: typeof (count) !== 'undefined',
+      },
+      className,
+    );
+
+    const rest = omit(others, [
```
lowiebenoot: `dark` is not in `others`, so you can ditch the `omit`.

PR #41, opened 2017-09-22
Button & IconButton component adjustments, ButtonGroup added

driesd:


components/button/theme.css
```diff
@@ -1,208 +1,406 @@
...
 }
 
-.bordered {
-  border: 1px solid var(--button-border-color);
+/* Button colors */
+.iconButton,
+.neutral,
+.mint,
+.outline,
+.ruby {
```
lowiebenoot: Q: do you need these classes? Because these classes just resemble all the possible button types? So they seem obsolete to me, and I think you can just put the content under `.button`.

driesd: You are right, I could just put the content under .button!

PR #45, opened 2017-09-26
Avatar component

driesd:


components/avatar/Avatar.js
```diff
@@ -0,0 +1,41 @@
...
+      theme[size],
+      className,
+    );
+
+    return (
+      <span className={classes} {...others} data-teamleader-ui="avatar">
+        <span className={theme.image}>
+          <img src={image} />
+        </span>
+      </span>
```
nickwaelkens: Why `<span />`? IMO something like `<figure />` makes more sense, ARIA wise, remember 😉

components/counter/Counter.js
```diff
@@ -6,25 +6,29 @@ import theme from './theme.css';
 
 class Counter extends PureComponent {
   static propTypes = {
+    borderColor: PropTypes.oneOf([ 'neutral', 'mint', 'aqua', 'violet', 'teal', 'gold', 'ruby', 'tealdark' ]),
```
lowiebenoot: Can't we use a `border` prop that is either `true` or `false`? I guess we're never going to use a gold counter with a ruby border?

lowiebenoot: ignore my comment, i'm wrong 👍

build/webpack.config.js
```diff
@@ -71,8 +72,10 @@ webpackConfig.plugins = [
 if (__DEV__) {
   debug('Enable plugins for live development (HMR, NoErrors).');
   webpackConfig.plugins.push(
+    new CopyWebpackPlugin([{ from: paths.client('static') }]),
     new webpack.HotModuleReplacementPlugin(),
     new webpack.NoEmitOnErrorsPlugin()
+
```
nickwaelkens: **Q:** How is adding this plugin different from the [existing code](https://github.com/teamleadercrm/teamleader-ui/blob/component/avatar/bin/compile.js#L18-L20)?

components/avatar/Avatar.js
```diff
@@ -0,0 +1,56 @@
...
+      className,
+    );
+
+    const imageClasses = cx(
+      theme.image,
+      theme[`border-${borderColor}`],
+    );
+
+    return (
+      <span className={avatarClasses} {...others} data-teamleader-ui="avatar">
```
nickwaelkens: **PP:** Still not convinced about using `<span>`s here, it just doesn't feel right... Why was that again?

components/avatar/AvatarStack.js
```diff
@@ -0,0 +1,63 @@
...
+    return (
+      <div
+        data-teamleader-ui="avatar-stack"
+        className={classes}
+        {...others}
+      >
+        { overflowAmount > 0 &&
+          <span className={theme.overflow} onClick={onOverflowClick}>{`+${overflowAmount}`}</span>
+        }
+        { childrenToDisplay.map(child => child) }
```
nickwaelkens: **Q:** No idea what this is doing here, haven't tested it, but AFAIK this does nothing.

```js
[0, 1, 2, 3].map(number => number); // [0, 1, 2, 3]
```

Unless this is some "fancy" way of doing something in React, get rid of it and replace with just something like
```jsx
{ childrenToDisplay }
```

components/avatar/theme.css
```diff
@@ -0,0 +1,148 @@
...
+    height: var(--size);
+    width: var(--size);
+  }
+}
+
+/* AvatarGroup */
+/*.group {
+  display: inline-block;
+  position: relative;
+}*/
```
nickwaelkens: Not used anymore?

driesd: whoops, I've removed this now.

components/avatar/AvatarStack.js
driesd: Just me being stupid, I've corrected that...

components/avatar/Avatar.js
driesd: There's no need to use a figure-tag, so which tag should I use? A div-tag maybe?

build/webpack.config.js
driesd: It enables me to serve the 'static' folder on localhost, so I could load an image from http://localhost/avatars/1.png

nickwaelkens: That's the whole point of the existing code, works perfectly here...

components/avatar/Avatar.js
```diff
@@ -0,0 +1,52 @@
+import React, { PureComponent } from 'react';
+import PropTypes from 'prop-types';
+import cx from 'classnames';
+import theme from './theme.css';
+
+class Avatar extends PureComponent {
+  static propTypes = {
+    borderColor: PropTypes.oneOf([ 'neutral', 'mint', 'aqua', 'violet', 'teal', 'gold', 'ruby', 'tealdark' ]),
```
Sambego: In most component you use a separate 'dark' prop to create a darker version of some colors. Here you use 'tealdark`. Maybe it's better to keep all components consistent?

components/avatar/Avatar.js
```diff
@@ -0,0 +1,52 @@
...
+
+    const imageClasses = cx(
+      theme.image,
+      theme[`border-${borderColor}`],
+    );
+
+    return (
+      <div className={avatarClasses} {...others} data-teamleader-ui="avatar">
+        <div className={imageClasses}>
+          <img src={image} />
```
Sambego: It would be nice to have an `alt` tag here, the name of the avatar or something like that.

driesd:
@lowiebenoot I was porting the spec to a story as you were commenting on this PR. So here it is!

components/counter/Counter.js
```diff
@@ -6,28 +6,32 @@ import theme from './theme.css';
...
+    const { borderColor, className, color, count, inactive, maxCount, size, ...others } = this.props;
 
     const classes = cx(
       theme.counter,
       theme[color],
       theme[size],
+      theme[`border-${borderColor}`],
       {
-        [theme.dark]: dark,
+        [theme.inactive]: inactive,
```
nickwaelkens: **B:** PHPStorm doesn't allow clicking through to this class definition. Can you use array notation here? Makes it easier to refactor too, if necessary.

build/webpack.config.js
```diff
@@ -73,6 +73,7 @@ if (__DEV__) {
   webpackConfig.plugins.push(
     new webpack.HotModuleReplacementPlugin(),
     new webpack.NoEmitOnErrorsPlugin()
+
```
lowiebenoot: Can you remove this?

components/counter/Counter.js
driesd: Agreed @nickwaelkens! I'll do it in a separate PR.

components/avatar/AvatarStack.js
```diff
@@ -0,0 +1,41 @@
...
+    direction: 'horizontal',
+    displayMax: 0,
+    inverse: false,
+    size: 'medium',
+  };
+
+  render() {
+    const { children, className, direction, displayMax, inverse, onOverflowClick, size, ...others } = this.props;
+
+    const classes = cx(theme.stack, theme[direction], theme[size], inverse ? [theme.light] : [theme.dark], className);
```
lowiebenoot: use `classNames` as we agreed on in another PR

components/avatar/Avatar.js
```diff
@@ -0,0 +1,39 @@
...
+    borderColor: 'neutral',
+    size: 'medium',
+  };
+
+  render() {
+    const { borderColor, className, counter, image, imageAlt, size, ...others } = this.props;
+
+    const avatarClasses = cx(theme.avatar, theme[size], className);
+
+    const imageClasses = cx(theme.image, theme[`border-${borderColor}`]);
```
lowiebenoot: use classNames here as well (so `avatarClassNames` and `imageClassNames`)

components/counter/Counter.js
lowiebenoot: I guess you mean `theme['inactive']`? That does give me the ability to click through, but to the wrong things...
Also, I don't think we should start changing things because PhpStorm doesn't support it.
screen shot 2017-10-17 at 11 02 58

driesd: **edit:** I did it in the smae PR 😉

PR #46, opened 2017-10-03
Experiment/postcss simple vars

driesd:
This PR is the IE fallback for using css vars in the cascade (selector based state variation). Read more about this [known issue](https://css-tricks.com/issue-preprocessing-css-custom-properties/).

Once the [browser support](http://caniuse.com/#search=css%20var) is completely covered for **Css vars**, all the lines containing a dollar sign can be removed.

### DIFF
--- a/components/button/theme.css
+++ b/components/button/theme.css
@@ -1,6 +1,86 @@
 @import '@teamleader/ui-colors';
 @import '@teamleader/ui-typography';
 
+$iconbutton-border: 0;
+$iconbutton-background-color: transparent;
+$iconbutton-background-color-hover: color(var(--color-teal-darkest) a(5%));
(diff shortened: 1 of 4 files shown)


components/button/theme.css
```diff
@@ -34,9 +114,9 @@
   white-space: nowrap;
   vertical-align: middle;
 
-  .children {
+ /* .children {
     padding-top: var(--button-children-padding-correction);
-  }
+  }*/
```
nickwaelkens: **Q** Can we just not get rid of this block if it's commented out?

driesd: Sorry, I have these comments removed in the next PR: https://github.com/teamleadercrm/teamleader-ui/pull/47/files

PR #59, opened 2017-10-04
Buttons: converted to crossbrowser supported css

driesd:


components/button/theme.css
```diff
@@ -85,261 +76,315 @@
...
+        box-shadow: none;
       }
     }
+
+    &.processing {
+      color: transparent;
+    }
+
+    .spinner:before {
+      border: 2px solid color(var(--color-teal-darkest) a(20%));
```
nickwaelkens: So you define 
```css
:root {
  --spinner-border-width: 2px;
}
```

but then you decide not to use it. Please be consistent; I know it's going to be a *visual* code mess when we use variables for everything, but future Dries (or someone else who will work on this) will be thankful.

PR #69, opened 2017-10-10
Feature/react storybook

Sambego:
### Added
- React storybook ([https://storybook.js.org/](https://storybook.js.org/))

### Removed 
- Manually created specs

----

When adding new components, a new story should be added for them in the `/stories` folder.
There are problems with some components, mostly because some "not so clean" things are happening in them. I've left out stories for the `popover` component, because of this. 

The storybook gives an overview of all components, and their states. When clicking the "show info" button in the top right corner, you can get an overview of the following things:
- The JSX to recreate the component with the current state
- An overview of all proptypes per component, and their current value
- The value types of all properties 
- Whether a property is required or not 
- The default values of all properties 

There is also a log of all action happening with the components when interacting with them, and an accessibility tester.


package.json
```diff
@@ -4,8 +4,14 @@
   "version": "0.1.5-beta",
   "author": "Teamleader <development@teamleader.eu>",
   "betterScripts": {
+    "storybook": {
+      "command": "start-storybook -p 3000",
```
lowiebenoot: Q: can you make it possible to change the port? I run this on a different port most of the times.

svenlombaert: Is it possible to use the port of the config? I don't want to close my stuff running on 3000

stories/counter.js
```diff
@@ -0,0 +1,28 @@
+import React from 'react';
+import { storiesOf } from '@storybook/react';
+import { checkA11y } from 'storybook-addon-a11y';
+import { withInfo } from '@storybook/addon-info';
+import Counter from '../components/counter';
+
+const colors = ['neutral', 'mint', 'violet', 'ruby', 'gold', 'aqua', 'white'];
```
lowiebenoot: There's no white counter.

package.json
Sambego: You can now use `yarn start -- --port 1337` to change the port.

stories/counter.js
Sambego: Removed the white color.

stories/radio.js
```diff
@@ -0,0 +1,32 @@
+import React from 'react';
+import { storiesOf } from '@storybook/react';
+import { action } from '@storybook/addon-actions';
+import { checkA11y } from 'storybook-addon-a11y';
+import { withInfo } from '@storybook/addon-info';
+import { RadioGroup, RadioButton } from '../components';
+
+let stringValue = 'thewalkingdead';
+const changeValue = value => {
```
lowiebenoot: The onchange doesn't seem to work

stories/loading.js
```diff
@@ -0,0 +1,31 @@
...
+  .add('small', () => (
+    <div style={{ height: '300px' }}>
+      <LoadingMolecule basePath={basePath} startColor="#00ACA9" stopColor="#1F7F79" type="small" />
+    </div>
+  ))
+  .add('large', () => (
+    <div style={{ height: '300px' }}>
+      <LoadingMolecule basePath={basePath} startColor="#00ACA9" stopColor="#1F7F79" type="large" />
+    </div>
+  ));
```
lowiebenoot: It would make more sense if the filename would reflect the component name more, so in thise case rename this file to `loadingMolecule.js`.

stories/menu.js
```diff
@@ -0,0 +1,21 @@
...
+    <Menu onSelect={action('select')} selectable={false} selected>
+      <MenuItem value="foo" caption="Caption" />
+      <MenuItem onClick={action('click')} value="bar" caption="Caption & Shortcut" shortcut="Ctrl + P" />
+      <MenuItem caption="Disabled ..." disabled shortcut="Ctrl + P" />
+      <MenuDivider />
+      <MenuItem caption="Caption & Icon" icon="check" />
+      <MenuItem caption="Caption, Icon & Shortcut" icon="list_bulleted" shortcut="Ctrl + P" />
+      <MenuItem caption="Disabled ..." icon="flag" shortcut="Ctrl + P" disabled />
+    </Menu>
+  ));
```
lowiebenoot: Not sure what's happening here, but it's not rendering good at all.

.storybook/config.js
```diff
@@ -0,0 +1,20 @@
...
+    stylesheet.link.topRight.zIndex = 99999999;
+    return stylesheet;
+  }
+});
+
+function loadStories() {
+  require('../stories');
+}
+
+configure(loadStories, module);
```
lowiebenoot: Do we need both this file and `stories/index.js`? Both are doing the `loadStories` and `configure`

package.json
```diff
@@ -5,7 +5,7 @@
   "author": "Teamleader <development@teamleader.eu>",
   "betterScripts": {
     "compile": {
-      "command": "node bin/compile",
+      "command": "build-storybook -o dist",
```
lowiebenoot: `dist` used to be our compiled version of the components. Now it's a compiled version of the storybooks? I guess that's now what we want to ship.

.storybook/config.js
nickwaelkens: Can this not be inlined instead of creating an (ES5) function?

package.json
```diff
@@ -155,6 +157,8 @@
     "lint:fix": "npm run lint -- --fix",
     "precommit": "npm run lint:fix",
     "prepare": "npm run build",
-    "start": "better-npm-run start"
+    "start": "better-npm-run start",
+    "storybook": "better-npm-run storybook",
```
nickwaelkens: Unused?

stories/button.js
```diff
@@ -0,0 +1,107 @@
...
+      </Button>
+      <Button onClick={action('clicked ouline button')} level="outline">
+        Outline button
+      </Button>
+      <Button onClick={action('clicked destructive button')} level="destructive">
+        Destructive button
+      </Button>
+    </ButtonGroup>
+  ))
+  .add('inverse', () => (
```
nickwaelkens: Can you change the background color of whatever is holding the `<Button>`s? There's no visual difference between a regular button and an inverse button now?

stories/loading.js
nickwaelkens: Also not rendering correctly because `basePath` returns something extremely exotic 😄

.storybook/config.js
Sambego: Removed the `index.js` and added it as an inline function.

Sambego:
The previews were rendering in Times New Roman when the component didn't have a font-family defined. It was not possible to add some default styling to storybook so I wrote a little decorator for it 😛 https://www.npmjs.com/package/@sambego/storybook-styles

nickwaelkens:
Cool, but in what universe is this a scalable solution? 
Does this mean there is no option to simply import a component from the suite and expect it to work "out of the box"?

Maybe we should really reconsider CSS-in-JS; cfr https://github.com/teamleadercrm/guild-frontend/issues/22

( summoning @stivni )

Sambego:
The components work out of the box, it is a lack of styling in the components which are causing the styling problems. Sections for example just use the default font-family used the document.

Either we:
- Add all necessary styling to the components
- Accept that some styling might vary depending on the document styles

stories/button.js
Sambego: Done.

screen shot 2017-10-11 at 11 02 25

stories/menu.js
Sambego: The problem with this icon menu items are that they rely on an icon font for the icons. This icon font is not included in the component styles, but needs to be loaded separate.

This icon font should be replaced with svg icons in the near future, so I would rather not include them in the storybook at the moment :)

Sambego:
The version is now also visible on the storybook page.

screen shot 2017-10-11 at 12 28 39

stivni:
nice move!

bin/compile.js
```diff
@@ -1,34 +0,0 @@
...
-    .then(() => {
-      debug('Compilation completed successfully.');
-    })
-    .catch((err) => {
-      debug('Compiler encountered an error.', err);
-      process.exit(1);
-    });
-};
-
-compile();
```
nickwaelkens: How are we going to build and publish our actual components now? 🤔

stivni:
sorry, clicked the wrong button while commenting ;)

bin/compile.js
Sambego: If I'm not mistaken, the components get build using the `npm run build` script which doesn't use those bin scripts but gulp.

nickwaelkens: @Sambego I'm sorry, you are right. I'm confusing different projects and their setups. Can you also make sure you remove the (now unused) packages? Thanks! 👍 

 :ok: by @nickwaelkens on 11/10/2017 13:12:45

Sambego: Done!

stories/menu.js
lowiebenoot: Ok. The menu will indeed have to change to using our icon components instead of the font.

bin/compile.js
nickwaelkens: Cool, I will make a follow-up PR that cleans out even more stuff (out of scope for now)

package.json
lowiebenoot: I was just checking again, and I'm a bit confused. We have a gulpfile as well which copies stuff to `lib`? I think I'm always publishing just by running `npm run deploy:prod`. 
@nickwaelkens do you remember this?

lowiebenoot: We really need to start documenting better 😢

nickwaelkens: I do remember we switched to an extremely simple gulpfile to speed up our build process. Webpack was jumping through all kinds of hoops, because it's a copy/paste of the setup for marketplace/core/etc. Documentation is something we need to work on, I added it to my todo list!

.storybook/config.js
```diff
@@ -0,0 +1,25 @@
...
+  source: true,
+  propTablesExclude: [],
+  styles: stylesheet => {
+    stylesheet.link.topRight.zIndex = 99999999;
+    return stylesheet;
+  }
+});
+setOptions({
+  name: `Version ${process.env.__VERSION__}`,
+  url: 'http://teamleader.design'
```
nickwaelkens: `https`? ✏️

Sambego: Good catch!

stories/button.js
```diff
@@ -0,0 +1,110 @@
...
+      <Button onClick={action('clicked button 1')}>button 1</Button>
+      <Button onClick={action('clicked button 2')}>button 2</Button>
+    </ButtonGroup>
+  ))
+  .add('with icon', () => (
+    <ButtonGroup>
+      <Button onClick={action('clicked button 1')}>
+        <IconAddMediumOutline />
+      </Button>
+      <Button onClick={action('clicked button 1')}>
```
nickwaelkens: Should be `clicked button 2`. #mierenneukerij 🙈

stories/menu.js
stivni: resolved?

.storybook/config.js
stivni: :ok: by @stivni on 12/10/2017 11:01:06

.storybook/webpack.config.js
```diff
@@ -0,0 +1,39 @@
+const webpack = require('webpack');
```
svenlombaert: Not used?

.storybook/webpack.config.js
```diff
@@ -0,0 +1,39 @@
...
+const config = require('../config');
+const pkg = require('../package.json');
+const argv = require('yargs').argv;
```
svenlombaert: not used?

.storybook/webpack.config.js
```diff
@@ -0,0 +1,39 @@
...
+  __VERSION__: JSON.stringify(pkg.version)
+};
+const cssModulesLoader = [
+  'css-loader?sourceMap&-minimize',
+  'modules',
+  'importLoaders=1',
+  'localIdentName=[name]__[local]___[hash:base64:5]'
+].join('&');
+
+const path = require('path');
```
svenlombaert: Also, this seems unused

nickwaelkens: I will make sure everything unused (old Webpack setup) will be removed in a follow-up PR.

(cfr https://github.com/teamleadercrm/teamleader-ui/pull/69#discussion_r143980511)

.storybook/webpack.config.js
Sambego: Removed.

.storybook/webpack.config.js
Sambego: Removed.

.storybook/webpack.config.js
Sambego: Removed.

nickwaelkens:
As this will become the landing page of our `teamleader.design` page, a proper title and favicon would be nice!

Sambego:
I haven't found a way to customize the title. It's possible to create a custom version of storybook, but that's beyond the scope of this pr I think.

nickwaelkens:
Such flexibility...

andreascreten:
Let me know if I need to change the deploy script for this.

Currently I have:

```
git pull origin master
npm install
npm run deploy:prod
```

PR #70, opened 2017-10-11
Component/tooltip

driesd:
schermafdruk 2017-10-11 17 06 00

schermafdruk 2017-10-11 17 04 11

**ATTENTION:** When the [StoryBook PR](https://github.com/teamleadercrm/teamleader-ui/pull/69) is merged, we should port the spec to StoryBook stories.


components/tooltip/Tooltip.js
```diff
@@ -0,0 +1,271 @@
...
+        tooltipDelay: defaultDelay,
+        tooltipHideOnClick: defaultHideOnClick,
+        tooltipIcon: defaultIcon,
+        tooltipPosition: defaultPosition,
+        tooltipShowOnClick: defaultShowOnClick,
+        tooltipSize: defaultSize,
+      };
+
+      state = {
+        active: false,
```
nickwaelkens: **PP** Prefer using `isCondition` signature for boolean values, eg. `isActive`.

components/tooltip/Tooltip.js
```diff
@@ -0,0 +1,271 @@
...
+          tooltipDelay, // eslint-disable-line no-unused-vars
+          tooltipHideOnClick, // eslint-disable-line no-unused-vars
+          tooltipIcon,
+          tooltipPosition, // eslint-disable-line no-unused-vars
+          tooltipShowOnClick, // eslint-disable-line no-unused-vars
+          tooltipSize,
+          ...other
+        } = this.props;
+
+        const _className = cx(theme.tooltip, theme[tooltipColor], theme[tooltipSize], {
```
nickwaelkens: We're never using underscores, I think you mean 
```js
const classNames = cx(...);
```
to be consistent with the rest of the codebase?

components/tooltip/Tooltip.js
driesd: As we recently discussed, we'd prefer to pass the props without the is prefix. But, functions that return a boolean value we should always prefix with is, has, ...

components/tooltip/Tooltip.js
```diff
@@ -0,0 +1,271 @@
...
+          this.setState({ visible: false });
+        }
+      };
+
+      getPosition(element) {
+        const { tooltipPosition } = this.props;
+
+        if (tooltipPosition === POSITION.HORIZONTAL) {
+          const origin = element.getBoundingClientRect();
+          const { width: ww } = getViewport();
```
Sambego: What does `ww` stand for? It's a bit vague :)

components/tooltip/Tooltip.js
```diff
@@ -0,0 +1,271 @@
...
+          const toRight = origin.left < ww / 2 - origin.width / 2;
+
+          return toRight ? POSITION.RIGHT : POSITION.LEFT;
+        } else if (tooltipPosition === POSITION.VERTICAL) {
+          const origin = element.getBoundingClientRect();
+          const { height: wh } = getViewport();
```
Sambego: The same goes for `wh`.

components/tooltip/Tooltip.js
```diff
@@ -0,0 +1,271 @@
...
+                ref={node => {
+                  this.tooltipNode = node;
+                }}
+                className={_className}
+                data-teamleader-ui="tooltip"
+                style={{ top, left }}
+              >
+                <div className={theme.tooltipInner}>
+                  <div className={theme.tooltipArrow} />
+                  {tooltipIcon && <div className={theme.tooltipIcon}>{tooltipIcon}</div>}
```
Sambego: Why do you have a separate `tooltipIcon` property? An icon can also be inside the tooltip element rendered the line below.

components/tooltip/Tooltip.js
```diff
@@ -0,0 +1,271 @@
...
+          };
+        } else if (position === POSITION.RIGHT) {
+          return {
+            top: top + height / 2 + yOffset,
+            left: left + width + xOffset,
+            position,
+          };
+        }
+
+        return undefined;
```
Sambego: You don't really have to explicitly return `undefined`. When a function does not return anything, it returns `undefined` by default.

components/tooltip/TooltipLabel.js
```diff
@@ -0,0 +1,35 @@
...
+  render() {
+    const { children, className, icon, inverse, ...others } = this.props;
+
+    const classes = cx(theme.label, className, {
+      [theme.inverse]: inverse,
+    });
+
+    return (
+      <span data-teamleader-ui="tooltip-label" className={classes} {...others}>
+        {icon}
```
Sambego: Same remark as above, why a separate property for icon when it can be a part of `children`.

components/tooltip/theme.css
```diff
@@ -0,0 +1,228 @@
...
+}
+
+.tooltip {
+  display: block;
+  padding: var(--tooltip-margin);
+  pointer-events: none;
+  position: absolute;
+  text-align: left;
+  transform-origin: top left;
+  transition: cubic-bezier(0.4, 0, 0.2, 1) var(--tooltip-animation-duration) transform;
```
Sambego: I've noticed that this bezier curve is being used by both the button and tooltip component. Maybe it's a good idea to store it in a variable somewhere.

driesd: You didn't now that yet, but the plan is to bundle all those animation stuff into **ui-animations**

components/tooltip/Tooltip.js
driesd: We chose to always treat the icons as a separate prop, so we could:
- create consistency over all the components
- apply css to the icon specificly
- do some checks with the prop to render something different

components/tooltip/TooltipLabel.js
driesd: Same answer as above 🤓

components/tooltip/theme.css
```diff
@@ -0,0 +1,229 @@
...
+  z-index: var(--tooltip-z-index);
+
+  &::before {
```
nickwaelkens: **PP:** I know this is the official way of writing pseudo selectors, but AFAIK we've always used single colons `:before` instead of double `::before`. My OCD is killing me, so either we change it everywhere to double or these to single 😬

Sambego: `:before` is the css2 way of writing the before pseudo element .
`::before` is the css3 way of writing the before pseudo element to distinguish pseudo classes (`:focus`, `:active`, ...) from pseudo elements. I would opt for the replace everywhere solution :)

spec/components/tooltip.js
```diff
@@ -0,0 +1,152 @@
+import React, { PureComponent } from 'react';
```
lowiebenoot: Can you make a story for it (and remove spec) so we can test it?

driesd: Done! 🎉

components/tooltip/theme.css
```diff
@@ -0,0 +1,229 @@
+@import '@teamleader/ui-colors';
+@import '@teamleader/ui-utilities';
+
+:root {
+  --unit: 10px;
```
lowiebenoot: why not just use the one form `variables.css`?

driesd: I'd suggest putting **--unit: 10px;** inside the ui-utilities and use it from there?

lowiebenoot: good for me. Do it in a later PR.

PR #71, opened 2017-10-11
Added color variations on Text & Heading components

driesd:


components/typography/Heading.js
```diff
@@ -8,17 +8,38 @@ const factory = (type, defaultElement) => {
...
       children: PropTypes.node,
       className: PropTypes.string,
+      color: PropTypes.oneOf(['white', 'neutral', 'mint', 'teal', 'violet', 'ruby', 'gold', 'aqua']),
       element: PropTypes.node,
+      soft: PropTypes.bool,
     };
 
     static defaultProps = {
       element: null,
+      soft: false,
```
nickwaelkens: **PP** Prefer using `isCondition` signature for boolean values, eg. `isSoft`.

driesd: As we recently discussed, we'd prefer to pass the props without the **is** prefix. But, functions that return a boolean value we should always prefix with **is**, **has**, ...

lowiebenoot: +1, I prefer props without `is` as well, so you can do for example `<Heading soft>` (makes more sense than `<Heading isSoft>)`.

nickwaelkens: Yeah, I was seriously confused, sorry boizzz 👍

components/typography/Heading.js
```diff
@@ -8,17 +8,38 @@ const factory = (type, defaultElement) => {
     static propTypes = {
       children: PropTypes.node,
       className: PropTypes.string,
+      color: PropTypes.oneOf(['white', 'neutral', 'mint', 'teal', 'violet', 'ruby', 'gold', 'aqua']),
```
Sambego: It looks like we're using this array of colors in a lot of places, maybe we can keep it somewhere separate and reuse it in all components. Should there ever be the need to add a new color, we have to add it in just one place.

Sambego:
The `Heading` and `Text` factories are completely the same, except the word "heading" and "text". Why no use a single factory to create these, instead of duplicating code?

screen shot 2017-10-12 at 10 44 56

svenlombaert:
I'm completely on the same track as @Sambego with this. Header is indeed a variation on the text component.

components/typography/Heading.js
driesd: Yes, we could do that, but not always the same color variations are applicable to our components.

driesd:
You are both right, but they get a different classname (.header vs .text) and if you take a look in ui-typography you'll notice that .heading & .text got slightly different styling.

PR #75, opened 2017-10-13
Refactor/menu

driesd:


components/menu/MenuItem.js
```diff
@@ -24,28 +23,30 @@ const factory = () => {
...
+        onClick(event, this);
       }
     };
 
     render() {
-      const { icon, caption, children, shortcut, selected, disabled, ...others } = this.props;
+      const { icon, caption, children, className, shortcut, selected, disabled, ...others } = this.props;
 
-      const className = cx(
+      const classNames = cx(
```
nickwaelkens: The way classnames are being used is lacking consisteny and I'm crying inside.
Here it's `classNames`, while...

1. https://github.com/teamleadercrm/teamleader-ui/blob/master/components/island/Island.js#L34  (`classes`)
2. https://github.com/teamleadercrm/teamleader-ui/blob/master/components/popover/Popover.js#L123 (`customClassNames`)
3. https://github.com/teamleadercrm/teamleader-ui/blob/master/components/radio/RadioButton.js#L65 (`_className` (which is weird that nobody, including me, didn't notice this underscore syntax before))


  Please...

lowiebenoot: I think we should just use `classNames` everywhere (that's also the reason that we import the `classNames` module as `cx` I guess).

driesd:
@lowiebenoot @nickwaelkens, I agree, we 'll use **classNames** in every component. In this PR lowie has already fixed it. Can we ship this pls?

PR #79, opened 2017-10-17
:wrench: say goodby to package-lock.json

lowiebenoot:
The `package-lock.json` is just annoying as fuck and doesn't work as we expected, so we agreed to disable it.


lowiebenoot:
Fixed merge conflict but waiting for Github to pick up the changes..................

PR #80, opened 2017-10-17
Refactor/status label

driesd:


nickwaelkens:
Good stuff, but no need to release until Friday imho. Let's reduce the amount of releases, very hard to keep track of them. When you want to test locally, `git pull && npm link` is your friend.

PR #83, opened 2017-10-18
Box component

driesd:
This component would be the solution to add spacings to any other component.

schermafdruk 2017-10-18 14 15 33

For now, as a proposal, I only implemented the **marginTop** property to our **Text** component.


components/box/Box.js
```diff
@@ -0,0 +1,40 @@
...
+import PropTypes from 'prop-types';
+import cx from 'classnames';
+import theme from './theme.css';
+
+class Box extends PureComponent {
+  static propTypes = {
+    children: PropTypes.node,
+    className: PropTypes.string,
+    element: PropTypes.node,
+    marginTop: PropTypes.number,
```
lowiebenoot: we'll probably want a `oneOff` here, so you can't use unexisting factors.

components/box/Box.js
```diff
@@ -0,0 +1,84 @@
...
+      marginRight,
+      marginTop,
+      padding,
+      paddingBottom,
+      paddingLeft,
+      paddingRight,
+      paddingTop,
+      ...others
+    } = this.props;
+
```
nickwaelkens: **Q:** Would it make sense to drop the `margin-#` classes and solely rely on the `margin-{direction}-#` classes?

This way, you could set the default values when destructuring here and remove the `margin-#` classes all together. This way we can add a little bit more sugar, eg. `marginVertical` and `marginHorizontal` (same shorthand like you would use in regular CSS)

Blatantly copied from `este`:

```js
const {
    margin,
    marginHorizontal = margin, // ES6 default values ftw?
    marginVertical = margin,
    marginBottom = marginVertical,
    marginLeft = marginHorizontal,
    marginRight = marginHorizontal,
    marginTop = marginVertical,

    padding,
    paddingHorizontal = padding,
    paddingVertical = padding,
    paddingBottom = paddingVertical,
    paddingLeft = paddingHorizontal,
    paddingRight = paddingHorizontal,
    paddingTop = paddingVertical,
} = this.props;
```

This way, you can also remove a lot of the defaultProps, as they will all fall back to `margin` 👍

components/box/Box.js
```diff
@@ -0,0 +1,84 @@
...
+        [theme[`margin-${margin}`]]: margin > 0,
+        [theme[`margin-bottom-${marginBottom}`]]: marginBottom > 0,
+        [theme[`margin-left-${marginLeft}`]]: marginLeft > 0,
+        [theme[`margin-right-${marginRight}`]]: marginRight > 0,
+        [theme[`margin-top-${marginTop}`]]: marginTop > 0,
+        [theme[`padding-${padding}`]]: padding > 0,
+        [theme[`padding-bottom-${paddingBottom}`]]: paddingBottom > 0,
+        [theme[`padding-left-${paddingLeft}`]]: paddingLeft > 0,
+        [theme[`padding-right-${paddingRight}`]]: paddingRight > 0,
+        [theme[`padding-top-${paddingTop}`]]: paddingTop > 0,
```
nickwaelkens: This is OK for now, but would like to see this refactored in a helper function of sorts. We perfectly know which "CSS like" props are allowed, so we can easily camelCase to kebab-case the names etc.

components/typography/Text.js
```diff
@@ -43,10 +45,12 @@ const factory = (baseType, type, defaultElement) => {
 
       const Element = element || defaultElement;
 
+      const rest = omit(others, ['soft']);
```
nickwaelkens: This seems a bit tedious to check every component to make sure not to send too many props down? We are checking in `<Box>` itself which props are supported, so not sure if this is necessary?

components/box/Box.js
lowiebenoot: because `paddingTop > 0` is the conditional here, you can't override the padding-top with `0`. Applies to all the others as well of course.

stories/typography.js
```diff
@@ -21,19 +22,19 @@ storiesOf('Typography', module)
   .addDecorator(checkA11y)
   .addDecorator(styles(baseStyles))
   .add('Headings', () => (
-    <div>
-      <Heading1>Heading 1</Heading1>
-      <Heading2>Heading 2</Heading2>
-      <Heading3>Heading 3</Heading3>
-      <Heading4>Heading 4</Heading4>
-    </div>
+    <Box>
```
lowiebenoot: Not really necessary to use a `Box` here right? It confused me at first, It kinda looked like you need to wrap your components into the `Box` for being able to use the margin and padding properties. So I would just change it to a div to avoid confusion.

stories/typography.js
```diff
@@ -21,19 +22,19 @@ storiesOf('Typography', module)
...
       <TextBody>Text body</TextBody>
-      <TextSmall>Text small</TextSmall>
-      <TextTiny>Text tiny</TextTiny>
-    </div>
+      <TextSmall marginTop={2}>Text small</TextSmall>
+      <TextTiny marginTop={2}>Text tiny</TextTiny>
+      <TextBody marginTop={4}><Monospaced>Monospaced number 1234567890</Monospaced></TextBody>
+      <TextBody marginTop={2}><OldStyleNumber>Old style number 1234567890</OldStyleNumber></TextBody>
+    </Box>
```
lowiebenoot: Maybe add another story to show the overriding of the general margin/padding. Something like `margin={4} margin-top={0}`.

components/box/Box.js
lowiebenoot: e.g. `padding={4} paddingTop={0}` will not work.

stories/typography.js
driesd: I fixed @nickwaelkens first remark with the default values. Now I can something like `margin={4} margin-top={0}`and it works!

 :ok: by @driesd on 19/10/2017 15:53:52

stories/typography.js
driesd: I've added some padding now, so the Box isn't just rendering a div. I did this because in [https://github.com/teamleadercrm/teamleader-ui/pull/84](this PR) I'm removing the generally set padding on the preview area.

components/box/Box.js
driesd: I fixed @nickwaelkens first remark with the default values. Now I can something like padding={4} padding-top={0} and it works!

 :ok: by @driesd on 19/10/2017 16:05:57

driesd: I implemented @nickwaelkens his solution (first remark) with the default values. Now you can perfectly override any margin/padding with a value of 0. So, right now, `padding={4} paddingTop={0}` works very well.

 :ok: by @driesd on 19/10/2017 16:18:37

stories/typography.js
driesd: I implemented @nickwaelkens his solution (first remark) with the default values. Now you can perfectly override any margin/padding with a value of 0. So, right now, padding={4} paddingTop={0} works very well.

 :ok: by @driesd on 19/10/2017 16:18:48

lowiebenoot: Yes, looks good. Tested it again and works perfectly now 💅 

 :ok: by @lowiebenoot on 19/10/2017 16:51:19

PR #86, opened 2017-10-19
Refactor/dialog

driesd:


nickwaelkens:
Is refactoring to React portals out of scope?

lowiebenoot:
@nickwaelkens I would do that in a separate PR. And first change to react v16 in another PR as well.

PR #97, opened 2017-10-23
Component/input

Sambego:
### Added
- Input component
- Label component

### DIFF
--- a/components/input/Input.js
+++ b/components/input/Input.js
@@ -156,9 +156,9 @@ export default class Input extends Component {
     }
 
     if (meta.error) {
-      return <IconWarningBadgedSmallFilled className={cx(theme['feedback-icon'], theme['error'])} />;
+      return <IconWarningBadgedSmallFilled className={cx(theme['validation-icon'], theme['error'])} />;
     } else if (meta.valid) {
-      return <IconCheckmarkSmallFilled className={cx(theme['feedback-icon'], theme['valid'])} />;
+      return <IconCheckmarkSmallFilled className={cx(theme['validation-icon'], theme['valid'])} />;
     }
   }
 
--- a/components/input/Input.js
+++ b/components/input/Input.js
@@ -130,37 +130,36 @@ export default class Input extends Component {
 
   renderValidationMessage() {
     const { meta } = this.props;
+    const props = {
+      className: theme['validation-text'],
+      marginTop: 2,
+    };
 
     if (!meta) {
       return;
     }
 
     if (meta.error) {
-      return (
-        <TextSmall marginTop={2} className={theme['error']}>
-          {meta.error}
-        </TextSmall>
-      );
+      return <TextSmall {...props}>{meta.error}</TextSmall>;
     } else if (meta.valid) {
-      return (
-        <TextSmall marginTop={2} className={theme['valid']}>
-          {meta.valid}
-        </TextSmall>
-      );
+      return <TextSmall {...props}>{meta.valid}</TextSmall>;
     }
   }
 
   renderValidationIcon() {
     const { meta } = this.props;
+    const props = {
+      className: theme['validation-icon'],
+    };
 
     if (!meta) {
       return;
     }
 
     if (meta.error) {
-      return <IconWarningBadgedSmallFilled className={theme['validation-icon']} />;
+      return <IconWarningBadgedSmallFilled {...props} />;
     } else if (meta.valid) {
-      return <IconCheckmarkSmallFilled className={theme['validation-icon']} />;
+      return <IconCheckmarkSmallFilled {...props} />;
     }
   }
 
--- a/components/input/Input.js
+++ b/components/input/Input.js
@@ -116,11 +116,13 @@ export default class Input extends Component {
   renderHelpText() {
     const { helpText, inverse } = this.props;
 
-    return helpText && (
-      <TextSmall color={inverse ? 'white' : 'neutral'} marginTop={1} soft>
-        {helpText}
-      </TextSmall>
-    );
+    if (helpText) {
+      return (
+        <TextSmall color={inverse ? 'white' : 'neutral'} marginTop={1} soft>
+          {helpText}
+        </TextSmall>
+      );
+    }
   }
 
   renderSpinnerControls() {
--- a/components/input/Input.js
+++ b/components/input/Input.js
@@ -17,6 +17,7 @@ export default class Input extends Component {
     className: PropTypes.string,
     counter: PropTypes.number,
     disabled: PropTypes.bool,
+    description: PropTypes.string,
     icon: PropTypes.oneOfType([PropTypes.func, PropTypes.element]),
     iconPlacement: PropTypes.oneOf(['left', 'right']),
     id: PropTypes.string,
@@ -112,6 +113,14 @@ export default class Input extends Component {
     }
   }
 
+  renderDescription () {
+    const { description, inverse } = this.props;
+
+    if(description) {
+      return <TextSmall color={inverse ? 'white' : 'neutral'} marginTop={1} soft>{description}</TextSmall>;
+    }
+  }
+
   renderSpinnerControls() {
     if (this.isNumberInput) {
       return (
@@ -174,7 +183,7 @@ export default class Input extends Component {
           {this.renderSpinnerControls()}
           {this.renderValidationIcon()}
         </div>
-
+        {this.renderDescription()}
         {this.renderValidationMessage()}
       </div>
     );
--- a/components/input/Input.js
+++ b/components/input/Input.js
@@ -21,6 +21,7 @@ export default class Input extends Component {
     icon: PropTypes.oneOfType([PropTypes.func, PropTypes.element]),
     iconPlacement: PropTypes.oneOf(['left', 'right']),
     id: PropTypes.string,
+    inverse: PropTypes.bool,
     meta: InputMetaPropTypes,
     onBlur: PropTypes.func,
     onChange: PropTypes.func,
@@ -35,6 +36,7 @@ export default class Input extends Component {
 
   static defaultProps = {
     iconPlacement: 'left',
+    inverse: false,
     disabled: false,
     placeholder: '',
     readOnly: false,
@@ -163,7 +165,7 @@ export default class Input extends Component {
   }
 
   render() {
-    const { className, counter, icon, iconPlacement, size, meta } = this.props;
+    const { className, counter, icon, iconPlacement, inverse, size, meta } = this.props;
     const classNames = cx(
       theme.wrapper,
       theme[`is-${size}`],
@@ -172,6 +174,7 @@ export default class Input extends Component {
         [theme['has-counter']]: counter,
         [theme['has-error']]: Boolean(meta && meta.error),
         [theme['has-validation-feedback']]: meta,
+        [theme['is-inverse']]: inverse,
         [theme['is-numeric']]: this.isNumberInput,
         [theme['is-valid']]: Boolean(meta && meta.valid),
       },
--- a/components/input/Input.js
+++ b/components/input/Input.js
@@ -18,7 +18,7 @@ export default class Input extends Component {
     className: PropTypes.string,
     counter: PropTypes.number,
     disabled: PropTypes.bool,
-    description: PropTypes.string,
+    helpText: PropTypes.string,
     icon: PropTypes.oneOfType([PropTypes.func, PropTypes.element]),
     iconPlacement: PropTypes.oneOf(['left', 'right']),
     id: PropTypes.string,
@@ -113,12 +113,12 @@ export default class Input extends Component {
   }
 
   renderDescription() {
-    const { description, inverse } = this.props;
+    const { helpText, inverse } = this.props;
 
-    if (description) {
+    if (helpText) {
       return (
         <TextSmall color={inverse ? 'white' : 'neutral'} marginTop={1} soft>
-          {description}
+          {helpText}
         </TextSmall>
       );
     }
--- a/components/input/Input.js
+++ b/components/input/Input.js
@@ -114,11 +114,15 @@ export default class Input extends Component {
     }
   }
 
-  renderDescription () {
+  renderDescription() {
     const { description, inverse } = this.props;
 
-    if(description) {
-      return <TextSmall color={inverse ? 'white' : 'neutral'} marginTop={1} soft>{description}</TextSmall>;
+    if (description) {
+      return (
+        <TextSmall color={inverse ? 'white' : 'neutral'} marginTop={1} soft>
+          {description}
+        </TextSmall>
+      );
     }
   }
 
--- a/components/input/Input.js
+++ b/components/input/Input.js
@@ -126,8 +126,8 @@ export default class Input extends Component {
   }
 
   renderSpinnerControls() {
-    if (this.isNumberInput) {
-      return (
+    return (
+      this.isNumberInput && (
         <div className={theme['spinner']}>
           <button className={theme['spinner-up']} onClick={this.handleIncreaseValue}>
             <IconChevronUpSmallOutline />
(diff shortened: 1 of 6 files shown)


components/input/Input.js
```diff
@@ -0,0 +1,161 @@
...
+      return this.setState({ value: this.state.value + this.props.step });
+    }
+
+    return this.setState({ value: this.props.step });
+  }
+
+  decreaseValue() {
+    if (this.state.value) {
+      return this.setState({ value: this.state.value - this.props.step });
+    }
```
nickwaelkens: As you know, React may batch multiple `setState()` calls into a single update for performance. Could you change this to the setState with function notation, just or my ease of mind?

```jsx
this.setState((prevState, props) => ({ value: prevState.value ? prevState.value - props.step : -props.step }));
```

(And of course everywhere else where comparing with previous state)

components/input/Input.js
```diff
@@ -0,0 +1,161 @@
...
+    step: 1,
+  };
+
+  constructor(props) {
+    super(props);
+
+    this.isNumberInput = Boolean(props.type === 'number');
+    this.handleChange = this.onChange.bind(this);
+    this.handleIncreaseValue = this.increaseValue.bind(this);
+    this.handleDecreaseValue = this.decreaseValue.bind(this);
```
nickwaelkens: We are using the `stage-0` preset, feel free to use new function bind operator `::`.

components/input/InputMetaPropTypes.js
```diff
@@ -0,0 +1,22 @@
...
+  pristine: PropTypes.bool,
+  submitting: PropTypes.bool,
+  submitFailed: PropTypes.bool,
+  touched: PropTypes.bool,
+  valid: PropTypes.bool,
+  visited: PropTypes.bool,
+  warning: PropTypes.string,
+});
+
+export default InputMetaPropTypes;
```
nickwaelkens: I don't really like React PropTypes, but this is clean! 👏

components/input/theme.css
```diff
@@ -0,0 +1,280 @@
...
+  color: var(--error-color);
+
+  svg {
+    width: 12px;
+  }
+
+  svg,
+  g,
+  path {
+    fill: var(--error-color);
```
nickwaelkens: FYI: You should be able to remove this (everywhere actually) when using `@teamleader/ui-icons@^0.0.7`, all you need to do is set the `color` property. 

(Check https://github.com/teamleadercrm/ui-icons/pull/3)

components/input/Input.js
Sambego: 🎉

components/input/Input.js
Sambego: If it eases your mind, sure!

components/input/theme.css
Sambego: That's a nice solution for this!

nickwaelkens: You can actually remove the entire `svg { color : ... }` block, it will take the parent's color so in this case you're good to go.

components/input/Input.js
```diff
@@ -55,28 +55,24 @@ export default class Input extends Component {
...
   }
 
   increaseValue() {
-    if (this.state.value) {
-      return this.setState({ value: this.state.value + this.props.step });
-    }
-
-    return this.setState({ value: this.props.step });
+    this.setState((state, props) => ({
+      value: state.value ? state.value + props.step : props.step,
```
nickwaelkens: Cool, thanks. Can you use `prevState` as variable name though, to be more explicit? You get the previous state and the current props as arguments 👍

components/input/theme.css
Sambego: ✅

components/input/Input.js
Sambego: ✅ Done.

components/input/Input.js
```diff
@@ -0,0 +1,158 @@
...
+        </div>
+      );
+    }
+  }
+
+  renderError() {
+    if (this.props.meta && this.props.meta.error) {
+      return (
+        <p className={theme['error']}>
+          <IconErrorSmallOutline /> {this.props.meta.error}
```
lowiebenoot: This is not the correct icon, it should be the filled one.

components/input/theme.css
```diff
@@ -0,0 +1,280 @@
...
+  border: 1px solid var(--input-border-color);
+  border-radius: var(--input-border-radius);
+  box-shadow: 0;
+
+  color: var(--input-text-color);
+  line-height: var(--input-height-medium);
+  font-size: var(--input-text-size);
+  font-family: var(--font-family-regular);
+
+  &:focus {
```
lowiebenoot: you also need to disable the outline here for Chrome.

components/label/theme.css
```diff
@@ -0,0 +1,17 @@
...
+  --label-font-size: calc(1.6 * var(--unit));
+  --label-line-height: 1.5;
+  --label-color: var(--color-teal-darkest);
+}
+
+.label {
+  display: block;
+
+  font-size: var(--label-font-size);
+  font-family: var(--font-family-semi-bold);
```
lowiebenoot: it doesn't have to be semi-bold

components/input/Input.js
```diff
@@ -0,0 +1,158 @@
...
+    placeholder: '',
+    readOnly: false,
+    size: 'medium',
+    step: 1,
+  };
+
+  constructor(props) {
+    super(props);
+
+    this.isNumberInput = Boolean(props.type === 'number');
```
lowiebenoot: Not really a fan of "storing" this in `this`. Now it's actually not possible to change an input from a number input to a normal one (or vice versa). We'll probably never do that, but you kind of expect that it would work.

components/input/Input.js
```diff
@@ -0,0 +1,158 @@
...
+      onChange: this.handleChange,
+      onFocus: onFocus,
+      placeholder,
+      readOnly,
+      step,
+      type,
+      value: this.state.value,
+    };
+
+    return <input {...props} />;
```
lowiebenoot: There's no way of adding extra attributes to the input field now (for example data-attributes). I would use destructuring so to pass just all props. Something like this (didn't test it):

```js
  renderInput() {
    const { bold, disabled, readOnly, ...others } = this.props;

    const classNames = cx(theme['input'], {
      [theme['is-disabled']]: disabled,
      [theme['is-read-only']]: readOnly,
      [theme['is-bold']]: bold,
    });

    const props = {
      className: classNames,
      disabled,
      onChange: this.handleChange,
      readOnly,
      value: this.state.value,
      ...others,
    };

    return <input {...props} />;
  }
```

This will also give you the same issues as I had with the `Box` props that should be removed first because they don't have to be passed onto the input field.

components/input/Input.js
```diff
@@ -0,0 +1,158 @@
...
+  render() {
+    const { className, counter, icon, iconPlacement, size, meta } = this.props;
+    const classNames = cx(theme.wrapper, theme[`is-${size}`], className, {
+      [theme[`has-icon-${iconPlacement}`]]: icon,
+      [theme['has-counter']]: counter,
+      [theme['has-errors']]: Boolean(meta && meta.error),
+    });
+
+    return (
+      <div>
```
lowiebenoot: Can you implement `Box` here as well?

components/input/theme.css
Sambego: 👍🏻

components/input/Input.js
```diff
@@ -0,0 +1,158 @@
...
+    return <input {...props} />;
+  }
+
+  renderCounter() {
+    if (this.props.counter) {
+      return <Counter className={theme.counter} count={this.props.counter} color="ruby" size="small" />;
+    }
+  }
+
+  renderSpinnerControls() {
```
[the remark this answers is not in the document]
Sambego: Unfortunately it is the name the webkit team chose to use to identify the increase/decrease buttons in a number input. I just used the same name to keep some kind of consistency.

eg:
```css
-webkit-inner-spin-button,
-webkit-outer-spin-button {
    ...
}
```

components/label/Label.js
```diff
@@ -0,0 +1,21 @@
...
+import theme from './theme.css';
+
+export default class Label extends PureComponent {
+  static propTypes = {
+    children: PropTypes.oneOfType([PropTypes.element, PropTypes.string, PropTypes.array]),
+    for: PropTypes.string,
+  };
+
+  render() {
+    const { children } = this.props;
```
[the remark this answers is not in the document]
Sambego: For is a reserved keyword, and thus can not be the name of a var :)

components/label/Label.js
```diff
@@ -0,0 +1,21 @@
...
+
+    return (
+      <Box element="label" htmlFor={this.props.for} marginBottom={3} className={theme['label']}>
```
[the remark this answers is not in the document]
Sambego: For is a reserved keyword, and thus can not be the name of a var :)

components/label/theme.css
Sambego: 👍🏻

components/input/theme.css
Sambego: Done.

components/input/Input.js
Sambego: Done

nickwaelkens:
Still having quite some errors when checking out the storybook;

```
Warning: Received `false` for a non-boolean attribute `bold`.

If you want to write it to the DOM, pass a string instead: bold="false" or bold={value.toString()}.
If you used to conditionally omit it with bold={condition && value}, pass bold={condition ? value : undefined} instead.
```

```
Warning: React does not recognize the `helpText` prop on a DOM element. If you intentionally want it to appear in the DOM as a custom attribute, spell it as lowercase `helptext` instead. If you accidentally passed it from a parent component, remove it from the DOM element.
```

```
Warning: Each child in an array or iterator should have a unique "key" prop.
```

And typing in the input field is extremely slow, or is that just me?

components/input/Input.js
```diff
@@ -0,0 +1,196 @@
...
+    this.handleChange = ::this.onChange;
+    this.handleIncreaseValue = ::this.increaseValue;
+    this.handleDecreaseValue = ::this.decreaseValue;
+
+    this.state = {
+      value: this.isNumberInput ? Number(props.value) || '' : props.value || '',
+    };
+  }
+
+  componentWillUpdate(props, state) {
```
InstaK: PP: Props is not being used here, it can be removed

components/input/Input.js
```diff
@@ -0,0 +1,196 @@
...
+  }
+
+  renderHelpText() {
+    const { helpText, inverse } = this.props;
+
+    if (helpText) {
+      return (
+        <TextSmall color={inverse ? 'white' : 'neutral'} marginTop={1} soft>
+          {helpText}
+        </TextSmall>
```
InstaK: PP:

components/input/Input.js
```diff
@@ -0,0 +1,196 @@
...
+      );
+    }
+  }
+
+  renderSpinnerControls() {
+    if (this.isNumberInput) {
```
InstaK: PP: this.isNumberInput && (
        <div className={theme['spinner']}>
          <button className={theme['spinner-up']} onClick={this.handleIncreaseValue}>
            <IconChevronUpSmallOutline />
          </button>
          <button className={theme['spinner-down']} onClick={this.handleDecreaseValue}>
            <IconChevronDownSmallOutline />
          </button>
        </div>
)
```

File: components/input/Input.js [hunk 18]
```diff
@@ -116,13 +116,11 @@ export default class Input extends Component {
...
     const { helpText, inverse } = this.props;
 
-    if (helpText) {
-      return (
-        <TextSmall color={inverse ? 'white' : 'neutral'} marginTop={1} soft>
-          {helpText}
-        </TextSmall>
-      );
-    }
+    return helpText && (
```
InstaK: PP: I think the return statement is superfluous here. If helpText is not falsy then it will return what comes after it.

components/input/Input.js
InstaK: PP: I also think here return is superfluous again, I might be wrong though because I didn't comment it like that and you still wrote 'return', could you confirm this please? @driesd

components/input/Input.js
InstaK: Nevermind, it's in a function so idd you *do* have to return it :) 

 :ok: by @InstaK on 13/02/2018 12:08:46

components/input/Input.js
InstaK: Nevermind, it's in a function so idd you *do* have to return it :) 

 :ok: by @InstaK on 13/02/2018 12:08:54

components/input/Input.js
```diff
@@ -0,0 +1,210 @@
...
+        <TextSmall color={inverse ? 'white' : 'neutral'} marginTop={1} soft>
+          {helpText}
+        </TextSmall>
+      )
+    );
+  }
+
+  renderSpinnerControls() {
+    return (
+      this.isNumberInput && (
```
nickwaelkens: **PP** 
```js
renderSpinnerControls() { 
	if (this.isNumberInput) {
		return (
			<div />
		);
	}
}
```

components/input/Input.js
```diff
@@ -0,0 +1,210 @@
...
+
+  renderValidationMessage() {
+    const { meta } = this.props;
+    const props = {
+      className: theme['validation-text'],
+      marginTop: 2,
+    };
+
+    if (meta && meta.error) {
+      return <TextSmall {...props}>{meta.error}</TextSmall>;
```
nickwaelkens: Like you're doing here 👍

components/label/Label.js
```diff
@@ -0,0 +1,67 @@
...
+
+    return (
+      <Box element="label" htmlFor={this.props.for} marginBottom={3} className={classNames}>
+        {React.Children.map(children, child => {
+          if (isComponentOfType(Input, child)) {
+            return React.cloneElement(child, childProps);
+          }
+
+          return (
+            <div>
```
nickwaelkens: Is the wrapping `<div>` really necessary here or is it just to please React? Since v16 we can return arrays, so that'd fix this if that's the case

components/input/Input.js
lowiebenoot: Can you change this as well @driesd?

components/input/Input.js
lowiebenoot: Not sure if this comment is still about the code that I see, but props is clearly used in this function?

 :ok: by @lowiebenoot on 13/02/2018 13:53:49

components/label/Label.js
```diff
@@ -0,0 +1,67 @@
...
+    for: PropTypes.string,
+    inverse: PropTypes.bool,
+    helpText: PropTypes.string,
+    required: PropTypes.bool,
+    size: PropTypes.oneOf(['small', 'medium', 'large']),
+  };
+
+  static defaultProps = {
+    inverse: false,
+    helpText: 'Optional',
```
lowiebenoot: I guess the default should be empty/false?

components/input/Input.js
driesd: Dropped the storage in `this.`.

 :ok: by @driesd on 13/02/2018 14:14:48

lowiebenoot:
I also just noticed that the height of the fields is not completely correct, seems to be 2 pixels less heigh than in the designs.

components/label/Label.js
InstaK: When adding `required` to the Label component helpText is removed. In some way I think it's logical, but then again I also think it's weird because the label is not required, but the input field... However, I think it would be way too hard to check the required prop on the Input component to change things in the Label component. 

In a way I think having it default empty is more logical, but then again, it also reminds you to specify if an input field is required or not...

I'm just listing pro's and cons here 😄

InstaK:
@lowiebenoot 2 pixels less high? Sounds like a 1px border mistake? 🤔

components/input/Input.js
lowiebenoot: We decided to leave it as it is now, and come back to this later.

 :ok: by @lowiebenoot on 15/02/2018 17:30:45

PR #104, opened 2017-10-24
Add a static method to split in the Box components

lowiebenoot:
While using the `Box` component I stumbled upon an issue where I couldn't add paddings/margins.
For example: I have a (simplified) `Toggle` component that I implemented with `Box`.

```js
class Toggle extends PureComponent {
  render() {
    const { label, ...others } = this.props;

    return (
      <Box>
        <input type="checkbox" {...others} />
        {label}
      </Box>
    );  
  }
}
```

If I would use the `Toggle` component, and I would add a padding/margin, those props would be assigned to the `input` instead of `Box`, which makes sense, because I'm destructuring them in the `input` element. I need to destructure them at that place because I want to be able to pass default checkbox props such as `disabled`, `checked`, `onChange`, ...

So I actually need a way to split off props that were meant for `Box` and other props.  I wanted to define those `Box` props only in 1 place, so I added a static method to `Box` that can split props into props for `Box` and other props.

It can then be implemented like this:

```js
class Toggle extends PureComponent {
  render() {
    const { label, ...others } = this.props;
    const { boxProps, otherProps } = Box.splitProps(others);

    return (
      <Box {...boxProps}>
        <input type="checkbox" {...otherProps} />
        {label}
      </Box>
    );  
  }
}
```

It's not the cleanest code, but I currently don't see a different way for solving this. If you know one, I would love to hear it!

### DIFF
--- a/components/box/Box.js
+++ b/components/box/Box.js
@@ -4,6 +4,22 @@ import cx from 'classnames';
 import theme from './theme.css';
 
 const spacings = [0, 1, 2, 3, 4, 5, 6, 7, 8];
+const availableProps = [
+  'margin',
+  'marginHorizontal',
+  'marginVertical',
+  'marginBottom',
+  'marginLeft',
+  'marginRight',
+  'marginTop',
+  'padding',
+  'paddingHorizontal',
+  'paddingVertical',
+  'paddingBottom',
+  'paddingLeft',
+  'paddingRight',
+  'paddingTop',
+];
 
 class Box extends PureComponent {
   static propTypes = {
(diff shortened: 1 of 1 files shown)


tcoopman:
I agree with the first part of @Sambego and would do it manually or add a method to the toggle component but not in Box. It's not the responsibility of box to split the props.
What if you want a margin both on box and the input? If Box handles it you have less control over it. You are responsible of passing the right props to Box.

lowiebenoot:
I didn't want to use proptypes because to me PropTypes are purely a thing for development mode and I don't want to rely on them in production. You might use [babel-plugin-transform-react-remove-prop-types](https://www.npmjs.com/package/babel-plugin-transform-react-remove-prop-types) as well, which will then actually break the implementation.

But I follow in the fact that it shouldn't be the responsibility of `Box`, so I'll just do it in the `Toggle` component.

PR #109, opened 2017-10-26
Font has been changes to Inter UI

driesd:


nickwaelkens:
Correct me if I'm wrong, but we will have to edit every component so it uses Inter UI. Now it seems like you are trusting the fact that the wrapper apps have it loaded..?

PR #111, opened 2017-10-27
Upgrade to React v16.0.0

driesd:
After upgrading to React v16.0.0. and when I show/hide a Toast component, or when checking a RadioButton, I get this console warning:

```
Warning: Can only update a mounted or mounting component. This usually means you called setState, replaceState, or forceUpdate on an unmounted component. This is a no-op.
```


nickwaelkens:
Is this issue present when just using the components as is, or only in a story? Not pointing any fingers, but might be caused by `@sambego/storybook-state`.

driesd:
Th fix for the warning is in `v1.0.3` of `@sambego/storybook-state`. The store needed to be unsubscribed when components were unmounted.

PR #117, opened 2017-11-07
Tag component

driesd:


components/tag/Tag.js
```diff
@@ -0,0 +1,69 @@
...
+
+  getCloseIcon() {
+    return this.props.size === 'large' ? <IconCloseMediumOutline /> : <IconCloseSmallOutline />;
+  }
+
+  render() {
+    const { className, inverse, label, onLabelClick, onRemoveClick, size, ...others } = this.props;
+
+    const classNames = cx(
+      theme.tag,
```
nickwaelkens: **PP** Use array notation here...

components/tag/Tag.js
```diff
@@ -0,0 +1,69 @@
...
+    size: PropTypes.oneOf(['small', 'medium', 'large']),
+  };
+
+  static defaultProps = {
+    inverse: false,
+    size: 'medium',
+  };
+
+  getCloseButtonColor() {
+    return this.props.inverse ? 'white' : 'neutral';
```
nickwaelkens: Kinda useless methods, just assign to a variable in `render`...

stories/tag.js
```diff
@@ -0,0 +1,82 @@
...
+const sizes = ['small', 'medium', 'large'];
+
+// const store = new Store({
+//   checked: false,
+// });
+
+// const handleOnChange = checked => {
+//   store.set({ checked });
+//   action(`onChange - checked: ${checked}`)();
+// };
```
nickwaelkens: Commented code, remove it? 🔥

components/tag/Tag.js
lowiebenoot: I prefer it without the array actually.

components/tag/theme.css
```diff
@@ -0,0 +1,114 @@
...
+  font-family: var(--font-family-medium);
+}
+
+.remove-button {
+  border-radius: 0 var(--tag-border-radius) var(--tag-border-radius) 0;
+  height: 100%;
+}
+
+.label-button {
+  border: 0 !important;
```
lowiebenoot: no other way to solve this without using `!important?`

components/tag/Tag.js
lowiebenoot: Nvm, thought you meant `cx([...])`.
Oops 🤓

components/tag/theme.css
driesd: Not realy, but I think it's allright to use `!important` for a specific css property override. Especially because we work with css-modules, so the override is isolated inside it's own component and won't affect others.

 :ok: by @driesd on 13/11/2017 16:05:25

PR #123, opened 2017-11-10
Bugfix: deleted the wierd stylesheet z-index config

driesd:
This fixes following error:

schermafdruk 2017-11-10 15 55 10


nickwaelkens:
Does this break anything? Is this the "show info" thingie in the right hand corner or what? 🤔

PR #126, opened 2017-11-14
Update/dialog

driesd:


lowiebenoot:
Approved, but I actually think we should still use the `<Portal>` hoc, which can now use react portals internally. Now we're writing the code for the portal in every component, while we could just do it once in the hoc, no?

PR #134, opened 2017-11-20
Update/tooltip

driesd:


components/tooltip/Tooltip.js
```diff
@@ -230,11 +198,6 @@ const tooltipFactory = (options = {}) => {
           ...other
```
nickwaelkens: **PP** Would really prefer using `lodash.omit` here instead of assigning useless variables.

components/tooltip/Tooltip.js
```diff
@@ -250,22 +213,33 @@ const tooltipFactory = (options = {}) => {
...
-                  {tooltipIcon && <div className={theme['icon']}>{tooltipIcon}</div>}
-                  <div className={theme['text']}>{tooltip}</div>
-                </Box>
-              </div>
-            </Portal>
+          createPortal(
+            <Transition
+              in={active}
+              onExited={() => {
+                document.body.removeChild(this.tooltipRoot);
```
nickwaelkens: Handling DOM operations in `render()` doesn't feel like the best place to do so.

components/tooltip/Tooltip.js
```diff
@@ -1,10 +1,10 @@
...
 import Box from '../box';
+import Transition from 'react-transition-group/Transition';
 import cx from 'classnames';
-import Portal from '../hoc/Portal';
+import { createPortal } from 'react-dom';
 import { getViewport } from '../utils/utils';
-import events from '../utils/events';
 import theme from './theme.css';
 
 const POSITION = {
```
nickwaelkens: `SIZES` is plural, `POSITION` isn't? 🐜

components/tooltip/Tooltip.js
lowiebenoot: I don't like it either, but it's the easiest way to do I guess. I think you can't use `componentWillUnmount` because it will remove the element before the transition happens, not sure though. Did you try this @driesd?

And also extract the function out of the render method.

components/tooltip/Tooltip.js
driesd: Valid point!

 :ok: by @driesd on 27/11/2017 14:21:35

components/tooltip/Tooltip.js
driesd: I extracted the arrow function out of render method, but I'm not able to use the `componentWillUnmount` lifecycle method instead. This is because it's fired to early and so it breaks the exit animation.

 :ok: by @driesd on 27/11/2017 14:23:56

components/tooltip/Tooltip.js
driesd: Done!

 :ok: by @driesd on 27/11/2017 14:24:08

PR #136, opened 2017-11-20
Update/avatar

driesd:


lowiebenoot:
Just wondering: what would be a usecase where you would add a classname to the image (prop `imageClassName`)?

PR #144, opened 2017-11-28
Component: DataGrid

driesd:
schermafdruk 2018-01-18 14 47 49


stories/datagrid.js
```diff
@@ -0,0 +1,113 @@
...
+        </DataGrid.HeaderRow>
+        {
+          rows.map((row, index) => {
+            return (
+              <DataGrid.Row key={index}>
+                <DataGrid.Cell>
+                  <TooltippedStatusBullet color={row.column1} tooltip={<TextTiny>Overdue</TextTiny>} tooltipColor={row.column1} tooltipSize="small" size="large"/>
+                </DataGrid.Cell>
+                <DataGrid.Cell>
+                  <TextTiny>{row.column2}</TextTiny>
```
lowiebenoot: Why are you using `TextTiny`?

lowiebenoot: Not sure, but maybe we should even add a default text style to the cell (fontsize 14px and color neutral), so we don't always have to use the typography components.

 :ok: by @lowiebenoot on 28/11/2017 11:30:24

lowiebenoot: Not sure, but maybe we should even add a default text style to the cell (fontsize 14px and color neutral), so we don't always have to use the typography components.

components/datagrid/DataGrid.js
```diff
@@ -0,0 +1,97 @@
...
+    this.handleRowSelectionChange = ::this.handleRowSelectionChange;
+    this.handleHeaderRowSelectionChange = ::this.handleHeaderRowSelectionChange;
+
+    this.state = {
+      selectedRows: [],
+    };
+  }
+
+  handleHeaderRowSelectionChange(value) {
+    const allChildrenIndexes = React.Children.map(this.props.children, child => child.key);
```
lowiebenoot: you're actually getting the `HeaderRow` key with this as well (which is undefined I guess in this case). It's probably never going to give us issues, but it feels a bit wrong.

So either we introduce extra levels into the datagrid (`Datagrid.Header` and `Datagrid.Body`?) or you can probably do a `isComponentOfType` check.

components/datagrid/DataGrid.js
```diff
@@ -0,0 +1,97 @@
...
+  render() {
+    const { children, className, selectable, ...others } = this.props;
+    const { selectedRows } = this.state;
+
+    const classNames = cx(theme['data-grid'], className);
+
+    return (
+      <Box data-teamleader-ui="data-grid" className={classNames} {...others}>
+        {React.Children.map(children, child => {
+          if (!isComponentOfType(Row, child) && !isComponentOfType(HeaderRow, child)) {
```
lowiebenoot: you can remove this if-statement and just put `return child` at the end of the function.

lowiebenoot:
Maybe we should also add some props for the alignment in cells. We'll probably use right-alignment often in cells.

stories/datagrid.js
driesd: You are right! I dropped the usage of our TextTiny component and added markup props to our Cell component.

 :ok: by @driesd on 28/11/2017 16:45:38

components/datagrid/DataGrid.js
```diff
@@ -0,0 +1,97 @@
...
+      this.setState({
+        selectedRows: [],
+      });
+    }
+  }
+
+  handleRowSelectionChange(rowIndex) {
+    const { selectedRows } = this.state;
+
+    if (selectedRows.indexOf(rowIndex) > -1) {
```
nickwaelkens: **PP** `selectedRows.includes(rowIndex)`

components/datagrid/DataGrid.js
```diff
@@ -0,0 +1,97 @@
...
+      this.setState({
+        selectedRows: selectedRows.filter(row => row !== rowIndex),
+      });
+    } else {
+      this.setState({
+        selectedRows: [...selectedRows, rowIndex],
+      });
+    }
+  }
```
nickwaelkens: **Overall PP** Would like to see the functional `setState` instead of passing an object.

```js
this.setState(prevState => {
  const selectedRows = prevState.selectedRows.includes(rowIndex)
    ? prevState.selectedRows.filter(row => row !== rowIndex)
    :  [...selectedRows, rowIndex];

    return {
      ...prevState,
      selectedRows,
     };
});
```

lowiebenoot: This is a nicer way indeed! Never thought of doing it that way.

components/datagrid/theme.css
```diff
@@ -0,0 +1,118 @@
...
+  font-family: var(--font-family-medium);
+}
+
+/* Cell widths */
+.flex-fit-content {
+  flex: 0 0 auto;
+}
+
+.flex-min-width {
+  flex: 0 0 30px;
```
lowiebenoot: Just wondering, why 30px?

lowiebenoot:
I resized to a very small window, and the flexing starts to do weird things (which is what I expected when using flex).
screen shot 2017-12-05 at 18 27 55

components/datagrid/DataGrid.js
```diff
@@ -0,0 +1,176 @@
...
+              if (isComponentOfType(HeaderRow, child)) {
+                return React.cloneElement(child, {
+                  onSelectionChange: event => this.handleHeaderRowSelectionChange(event),
+                  selected: selectedRows.length === children[1].length,
+                  selectable,
+                  sliceTo: stickyFromLeft > 0 ? stickyFromLeft : 0,
+                });
+              }
+
+              if (isComponentOfType(BodyRow, child)) {
```
lowiebenoot: use `else if`

components/datagrid/BodyRow.js
```diff
@@ -0,0 +1,40 @@
...
+    selectable: PropTypes.bool,
+    selected: PropTypes.bool,
+    sliceFrom: PropTypes.number,
+    sliceTo: PropTypes.number,
+  };
+
+  render() {
+    const { className, children, sliceFrom, sliceTo, onSelectionChange, selected, selectable, ...others } = this.props;
+
+    const childrenArray = Array.isArray(children) ? children : [children];
```
nickwaelkens: This could be written like 
```js
const childrenArray = [...children];
```

And then maybe these 2 variables (`childrenArray` and `childrenSliced`) would not be necessary and can just write it inline?

components/datagrid/Cell.js
```diff
@@ -0,0 +1,63 @@
...
+      theme[`has-border-${border}`],
+      {
+        [theme['is-soft']]: soft,
+        [theme['is-strong']]: strong,
+      },
+      className,
+    );
+
+    return (
+      <div className={classNames} data-teamleader-ui="datagrid-cell" {...others}>
```
nickwaelkens: `Box` is A-OK 😃

components/datagrid/DataGrid.js
```diff
@@ -0,0 +1,176 @@
...
+        selectedRows,
+      };
+    });
+  }
+
+  setCalculatedRowWidth() {
+    const rowDOMNodes = [];
+    let maxRowWidth = 0;
+
+    Array.from(this.rowNodes.values())
```
nickwaelkens: This could also be 
```js
[...this.rowNodes.values()]
```

components/datagrid/DataGrid.js
```diff
@@ -0,0 +1,176 @@
...
+      .filter(rowNode => rowNode != null)
+      .forEach(rowNode => {
+        const rowDOMNode = ReactDOM.findDOMNode(rowNode);
+        const totalRowChildrenWidth = Array.from(rowDOMNode.children)
```
nickwaelkens: This could also be 
```js
[...rowDOMNode.children]
```

components/datagrid/FooterRow.js
```diff
@@ -0,0 +1,33 @@
...
+    children: PropTypes.any,
+    preserveSelectableSpace: PropTypes.bool,
+    sliceFrom: PropTypes.number,
+    sliceTo: PropTypes.number,
+  };
+
+  render() {
+    const { className, children, sliceFrom, sliceTo, preserveSelectableSpace, ...others } = this.props;
+
+    const childrenArray = Array.isArray(children) ? children : [children];
```
nickwaelkens: Same comment as before

components/datagrid/HeaderRow.js
```diff
@@ -0,0 +1,40 @@
...
+    selectable: PropTypes.bool,
+    selected: PropTypes.bool,
+    sliceFrom: PropTypes.number,
+    sliceTo: PropTypes.number,
+  };
+
+  render() {
+    const { className, children, sliceFrom, sliceTo, onSelectionChange, selected, selectable, ...others } = this.props;
+
+    const childrenArray = Array.isArray(children) ? children : [children];
```
nickwaelkens: Guess what 💯

components/datagrid/BodyRow.js
driesd: This doesn't work because `children` can be either an `object` or an `array`. The spread operator can not handle mixed types.

 :ok: by @driesd on 23/01/2018 15:35:10

components/datagrid/FooterRow.js
driesd: This doesn't work because `children` can be either an `object` or an `array`. The spread operator can not handle mixed types.

 :ok: by @driesd on 23/01/2018 15:35:24

components/datagrid/HeaderRow.js
driesd: This doesn't work because `children` can be either an `object` or an `array`. The spread operator can not handle mixed types.

 :ok: by @driesd on 23/01/2018 15:35:28

PR #146, opened 2017-11-29
Tab Component

InstaK:
**Added**
Tab component. Made a TabGroup component to easily be able to invert the colors.


lowiebenoot:
The PR title should be about the tab component, not navigation 😉

components/tab/TabGroup.js
```diff
@@ -0,0 +1,36 @@
+import React, { PureComponent } from 'react';
+import PropTypes from 'prop-types';
+import Box from '../box';
+import cx from 'classnames';
+import theme from './theme.css';
+
+class TabGroup extends PureComponent {
+  static propTypes = {
+    children: PropTypes.node,
```
lowiebenoot: children should be required

components/tab/TitleTab.js
```diff
@@ -0,0 +1,43 @@
+import React, { PureComponent } from 'react';
+import PropTypes from 'prop-types';
+import Box from '../box';
+import cx from 'classnames';
+import theme from './theme.css';
+
+import { Heading4 } from '../typography';
+
+class TitleTab extends PureComponent {
```
lowiebenoot: Why is it called `TitleTab` and not just `Tab`. Or do you still have to create the IconTab? I want to see a story with icons as well.

components/tab/TitleTab.js
```diff
@@ -0,0 +1,43 @@
...
+    return (
+      <Box
+        data-teamleader-ui="title-tab"
+        className={classNames}
+        marginHorizontal={3}
+        paddingHorizontal={3}
+        paddingVertical={4}
+        {...others}
+      >
+        <Heading4>{children}</Heading4>
```
lowiebenoot: Using a `h4` isn't very semantically correct in this case. Luckily the `Heading4` component has a prop to change the element 😉  I would use a span instead.

components/tab/theme.css
```diff
@@ -0,0 +1,56 @@
...
+  &::after {
+    content: '';
+    position: absolute;
+    display: block;
+    color: var(--color-teal-darkest);
+    box-shadow: inset 0px -3px 0px 0px var(--color-mint);
+    height: 3px;
+    width: 100%;
+    bottom: 0;
+    transition: transform .25s ease-in-out;
```
lowiebenoot: can you use the transition variables from `@teamleader/ui-animations`?

 :ok: by @lowiebenoot on 29/11/2017 15:46:59

components/tab/theme.css
```diff
@@ -0,0 +1,56 @@
+@import '@teamleader/ui-colors';
+
+.title-tab {
```
lowiebenoot: Can you fix the styling when adding a `href` prop to the tab?
The same thing needs to be done when using a button element as for the tab, that looks a bit fucked up right now.
We need it to be possible to be a button for usability reasons. You can't use an anchor element without a href, because that's interactive with the keyboard (tab and enter/space I guess)

components/tab/TabGroup.js
InstaK: Argh I keep forgetting to think about this. In that case children is not required either in `ButtonGroup.js`, I might fix that (in other PR)

 :ok: by @InstaK on 29/11/2017 16:34:47

components/tab/TitleTab.js
InstaK: Indeed, an IconTab is on its way, but not yet.

 :ok: by @InstaK on 29/11/2017 16:35:49

components/tab/TitleTab.js
InstaK: You're right 👍 

 :ok: by @InstaK on 29/11/2017 16:36:04

components/tab/theme.css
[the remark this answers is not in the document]
InstaK: I tried to use the animation-duration but that is .35s, Stephen asked me to make it snappier so I made it .25s. Do I go with .35s (the variable) or .25s?

components/tab/theme.css
```diff
@@ -0,0 +1,56 @@
...
+  color: var(--color-neutral-darkest);
+  display: flex;
+  align-items: center;
+  justify-content: center;
+  outline: 0;
+  user-select: none;
+  position: relative;
+
+  &:hover {
+    background-color: rgba(130, 130, 140, 0.24);
```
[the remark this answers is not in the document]
InstaK: Didn't think of changing it to hex, saw there was no rgba values in `ui-colors` but the rgb value is the same as `82828c` hex which is `neutral-darkest`

 :ok: by @InstaK on 29/11/2017 16:42:31

components/tab/theme.css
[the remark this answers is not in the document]
lowiebenoot: Though we had multiple speeds, but apparently not. Nvm then. We might add some more speeds later in the animations library, but that's for later.

 :ok: by @lowiebenoot on 29/11/2017 17:11:35

components/tab/theme.css
InstaK: Thanks for explaining why it's also needed

 :ok: by @InstaK on 30/11/2017 11:09:19

components/tab/theme.css
```diff
@@ -0,0 +1,66 @@
...
+  &::after {
+    content: '';
+    position: absolute;
+    display: block;
+    color: var(--color-teal-darkest);
+    box-shadow: inset 0px -3px 0px 0px var(--color-mint);
+    height: 3px;
+    width: 100%;
+    bottom: 0;
+    transition: transform var(--animation-duration) var(--animation-curve-fast-out-slow-in);
```
lowiebenoot: didn't you say it was too slow?

 :ok: by @lowiebenoot on 30/11/2017 14:59:57

lowiebenoot:
Changed label to WIP since there is no IconTab yet

components/tab/theme.css
InstaK: I talked to Stephen, he said I could take this and he would talk with you about changing the variable.

 :ok: by @InstaK on 30/11/2017 18:33:15

InstaK:
Oh God this is not going to take ages again before this PR can be finished, is it 😞

lowiebenoot:
Then continue working on it...? 🚎

lowiebenoot:
It might be better to try to merge this PR without the `IconTab` component (if you feel the same way @InstaK just remove the WIP label).

InstaK:
Up for re-review, added IconTab @lowiebenoot @driesd @nickwaelkens @svenlombaert

components/tab/IconTab.js
```diff
@@ -0,0 +1,65 @@
...
+
+const iconMap = {
+  crm: IconContactsMediumOutline,
+  invoices: IconInvoiceMediumOutline,
+  products: IconProductsMediumOutline,
+  stats: IconStatsMediumOutline,
+  time: IconTimerMediumOutline,
+  deals: IconCheckmarkMediumOutline,
+  planning: IconCalendarMediumOutline,
+};
```
nickwaelkens: We do not want any specific icons here. It should be able to accept any icon as a prop; otherwise if we want to update project A with icon B, we'd have to `teamleader-ui`. Not scalable!

components/tab/IconTab.js
```diff
@@ -0,0 +1,65 @@
...
+
+  render() {
+    const { active, children, className, counter = null, icon, ...others } = this.props;
+    const classNames = cx(theme['icon-tab'], { [theme['is-active']]: active }, className);
+
+    const IconToRender = iconMap[icon.toLowerCase()];
+
+    return (
+      <Box
+        data-teamleader-ui="title-tab"
```
nickwaelkens: This should be `icon-tab`, no?

components/tab/IconTab.js
```diff
@@ -0,0 +1,65 @@
...
+        className={classNames}
+        marginHorizontal={3}
+        paddingHorizontal={3}
+        paddingVertical={4}
+        {...others}
+      >
+        <IconToRender element="span">{children}</IconToRender>
+        {counter && React.cloneElement(counter, { className: theme['counter'], koekoek: 'jongen' })}
```
nickwaelkens: Dude. `koekoek: 'jongen'`

components/tab/IconTab.js
InstaK: Good catch, copy pasteritis

 :ok: by @InstaK on 14/12/2017 13:41:12

components/tab/IconTab.js
InstaK: Desperate times call for desperate measures.

 :ok: by @InstaK on 14/12/2017 13:41:49

components/tab/IconTab.js
InstaK: I thought it was said there was no other way to solve this? I asked this before that I wanted to only import the icon that's being requested, but that didn't work. I assumed a huge iconMap was the only way out. I don't know how to do it differently -> tip?

nickwaelkens: Like I said, we should import the icon that we want in the `IconTab` in the project that needs it. eg in `core`

components/tab/IconTab.js
```diff
@@ -0,0 +1,65 @@
...
+};
+
+class IconTab extends PureComponent {
+  static propTypes = {
+    active: PropTypes.bool,
+    children: PropTypes.node.isRequired,
+    className: PropTypes.string,
+    counter: PropTypes.node,
+    element: PropTypes.node,
+    href: PropTypes.string.isRequired,
```
lowiebenoot: `href` should not be listed in the propTypes because you can now also use other elements.

stories/tab.js
```diff
@@ -0,0 +1,163 @@
...
+import { Store, State } from '@sambego/storybook-state';
+import { action } from '@storybook/addon-actions';
+import { IconTab, TabGroup, TitleTab } from '../components/tab';
+import { baseStyles, centerStyles } from '../.storybook/styles';
+
+const store = new Store({
+  items: [
+    {
+      active: false,
+      href: '#',
```
lowiebenoot: can you change the story to using buttons? When I click a tab, it somehow navigates me to a different page inside storybook which is really annoying.

lowiebenoot: Or use a preventDefault, I don't really care in this case.

components/tab/TitleTab.js
lowiebenoot: Can you add an `onClick` to the component (and on IconTab as well). If you now click on an item, it keeps the focus state, which is default behaviour, but kind of annoying and ugly. So you can trigger a blur in the onClick of the element. I did the same in the `Checkbox` component, you can use that as an example.

stories/tab.js
InstaK: Yeah I also noticed that, but if I wanted to see the active state I needed to add a href to the anchor tag, right? So once I see it works, I should change it to buttons?

lowiebenoot: Yes, or as an alternative you can use `event.preventDefault();` in the `onClick`'s

components/tab/IconTab.js
InstaK: This is idd a way better approach. 👍 

 :ok: by @InstaK on 02/01/2018 15:37:03

PR #150, opened 2017-12-06
Menu: font size change and obsolete css variables clean up

driesd:


components/menu/theme.css
```diff
@@ -11,13 +11,10 @@
   --menu-outline-border-radius: calc(0.2 * var(--unit));
   --menu-item-hover-background: var(--color-neutral-light);
   --menu-item-selected-background: transparent;
-  --menu-item-icon-font-size: calc(1.4 * var(--unit));
-  --menu-item-icon-size: calc(1.6 * var(--menu-item-icon-font-size));
   --menu-item-height: calc(2.8 * var(--unit));
   --menu-item-padding: calc(1.6 * var(--unit));
-  --menu-item-font-size: calc(1.6 * var(--unit));
+  --menu-item-font-size: calc(1.4 * var(--unit));
```
lowiebenoot: maybe we should use `.text-small` from ui-typography instead?

driesd: @lowiebenoot That was a very good idea!

 :ok: by @driesd on 06/12/2017 16:33:03

PR #158, opened 2017-12-13
Update/toast: new styleguide specs

driesd:


components/toast/Toast.js
```diff
@@ -93,17 +93,17 @@ const factory = (Link, IconButton) => {
                   {label}
                   {children}
                 </TextSmall>
-                {onClick ? (
+                {onClose ? (
                   action ? (
                     <TextSmall className={theme['action-link']} color="white">
-                      <Link onClick={onClick}>{action}</Link>
+                      <Link onClose={onClose}>{action}</Link>
```
nickwaelkens: Does `Link` have an `onClose` prop..? 🤔

driesd: aw shit... 🙈 

 :ok: by @driesd on 13/12/2017 15:03:50

components/toast/Toast.js
```diff
@@ -2,23 +2,26 @@ import React, { PureComponent } from 'react';
 import PropTypes from 'prop-types';
 import Transition from 'react-transition-group/Transition';
 import cx from 'classnames';
-import { Button, IconButton } from '../button';
+import { IconButton } from '../button';
+import { TextSmall } from '../typography';
+import Link from '../link';
```
lowiebenoot: I think Link is a bit misused here, since we want it to act like a button (so not an `<a>`). I think we need to create another component `LinkButton` which is actually a button, but looks like a link.

```js
const LinkButton = props => <Link element="button" data-teamleader-ui="link-button" {...props} />;
```

lowiebenoot: I think Link is a bit misused here, since we want it to act like a button (so not an `<a>`). I think we need to create another component `LinkButton` which is actually a button, but looks like a link.

```js
const LinkButton = props => <Link element="button" data-teamleader-ui="link-button" {...props} />;
```

driesd: @lowiebenoot I follow your opinion about this LinkButton! I also think we need that, especially if you look at de [designs](https://app.zeplin.io/project/599af68db2f4978a56dec839/screen/5a339f3e0e1ba4f27ec6a7c8) of our (yet to build) pagination component.

driesd: @lowiebenoot You can check out [this PR](https://github.com/teamleadercrm/teamleader-ui/pull/164) for the new LinkButton component.

PR #162, opened 2017-12-14
Update/message: use our ButtonGroup instead of div

driesd:
I removed the obsolete styling and re-ordered some classes in the css file.
No breaking changes.

### DIFF
--- a/components/message/Message.js
+++ b/components/message/Message.js
@@ -32,7 +32,7 @@ class Message extends PureComponent {
           {children}
           {hasAction && (
             <ButtonGroup className={theme['actions']} marginTop={4}>
-              {button && <span className={theme['button']}>{button}</span>}
+              {button && React.cloneElement(button, { className: theme['button'] })}
               {link}
             </ButtonGroup>
           )}
--- a/components/message/Message.js
+++ b/components/message/Message.js
@@ -1,6 +1,7 @@
 import React, { PureComponent } from 'react';
 import PropTypes from 'prop-types';
 import Box from '../box';
+import { ButtonGroup } from '../button';
 import cx from 'classnames';
 import theme from './theme.css';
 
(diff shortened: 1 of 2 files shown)


components/message/Message.js
```diff
@@ -30,10 +31,10 @@ class Message extends PureComponent {
         <div className={theme['content']}>
           {children}
           {hasAction && (
-            <div className={theme['actions']}>
+            <ButtonGroup className={theme['actions']} marginTop={4}>
               {button && <span className={theme['button']}>{button}</span>}
```
nickwaelkens: Do you think wrapping in a `span` is still necessary? What if we add the `class` to the `button` itself? 🤔

driesd: The button itself is actually injected by a prop `button`. So we kinda need a wrapper to fix it's margins when changing direction to vertical.

Note: This is temporary because I'm refactoring the Box and ButtonGroup component to be responsible for the alignment and correct margins. Afterwards we can drop this wrapping span element.

nickwaelkens: I was not talking about `ButtonGroup`, rather about the `span` that was already there wrapping `{button}`. I'm pretty sure you can achieve the same effect without having the `span` there by doing 
```jsx
{button && React.cloneElement(button, { className: theme['button'] })}
```

driesd: I know you're talking about the span and not the ButtonGroup. And my point is that once the Box is refactored, I can drop this span entirely.

components/message/theme.css
```diff
@@ -7,6 +7,15 @@
   align-items: center;
 }
 
+.content {
+  flex: 2;
+}
+
+.image {
+  flex: 1;
+  text-align: center;
```
InstaK: PP: `text-align: center;` for an svg? Just flex this and `justify-content: center`

components/message/Message.js
InstaK: I think eventual refactoring is no valid argument for writing worse code. Why would you not *always* want to write the most performant code possible when the solution is as easy as this? #notproudtobepartofteamleader with that way of thinking

components/message/theme.css
driesd: You are right @InstaK, we can use flexbox props here.

components/message/Message.js
driesd: so deep, @InstaK 👏

InstaK: Just being passionate and wanting the best :) Thanks for changing it, now I can sleep at night ❤️ 

 :ok: by @InstaK on 14/12/2017 15:02:59

PR #163, opened 2017-12-15
Update/box flex props

driesd:


InstaK:
Aha, apparently you just did. 👍

components/box/Box.js
```diff
@@ -69,10 +114,18 @@ class Box extends PureComponent {
       className,
     );
 
+    const styles = {
+      ...(flex && { flex }),
+      ...(flexBasis && { flexBasis }),
+      ...(flexGrow && { flexGrow }),
+      ...(flexShrink && { flexShrink }),
+      ...(order && { order }),
+    };
```
nickwaelkens: Are we handling prefixes anywhere?
Also, this feels very clunky, PP to use `Object.assign` here, it ignores falsey values 

```js
const y = { a : 1 };
Object.assign({}, y, null, undefined); // { a : 1 }
```

driesd: **Prefixes:** The above changes seem to work in Chrome, Firefox, Safari & IE11 and iOS10. I don't have the feeling we should prefix in this case.

**Object.assign:** This results in an object containing all keys irrespective it's value is undefined or not. And so, the above method is a better solution.

PR #164, opened 2017-12-19
Component/linkbutton

driesd:


components/button/LinkButton.js
```diff
@@ -0,0 +1,85 @@
...
+      Element,
+      props,
+      icon && iconPlacement === 'left' ? icon : null,
+      label || children ? (
+        <span className={theme['children']}>
+          {label}
+          {children}
+        </span>
+      ) : null,
+      icon && iconPlacement === 'right' ? icon : null,
```
nickwaelkens: Get rid of these useless `null` return values and use shorthand `&&`

components/button/LinkButton.js
```diff
@@ -0,0 +1,85 @@
...
+
+    const Element = element || 'button';
+
+    const classNames = cx(
+      theme['link-button'],
+      {
+        [theme['disabled']]: disabled,
+        [theme['icon-only']]: !label && !children,
+        [theme['inverse']]: inverse,
+        [theme[size]]: theme[size],
```
nickwaelkens: `is-disabled`, `is-inverse` or are we not doing that here?

components/button/LinkButton.js
```diff
@@ -0,0 +1,85 @@
...
+      onMouseUp: this.handleMouseUp,
+      onMouseLeave: this.handleMouseLeave,
+      'data-teamleader-ui': 'link-button',
+    };
+
+    return React.createElement(
+      Element,
+      props,
+      icon && iconPlacement === 'left' && icon,
+      label || children ? (
```
InstaK: Q: Could you translate these lines to English so I can just understand what's happening here? I guess I don't know enough of these shorthands and can't really figure out what's happening.

lowiebenoot: It's just a ternary operator, but it's just a bit confusing because of the multine and the OR in the condition.

it's actually just the same as this
```jsx
if (label || children) {
  return <span className={theme['children']}>{label}{children}</span>
}
return null;
```

 :ok: by @lowiebenoot on 04/01/2018 17:02:57

InstaK: Pretty funny you just answered this, I just understood it by looking at it again. My confusion was the comma on line 72, I thought `label || children ?` was connected to the `&&` on line 72, but it's not. Was just about to comment I finally understood :) Thx though

 :ok: by @InstaK on 04/01/2018 17:06:57

nickwaelkens:
```js

PR #166, opened 2017-12-19
Component/pagination

driesd:


stories/pagination.js
```diff
@@ -0,0 +1,59 @@
...
+import { withInfo } from '@storybook/addon-info';
+import { Island, Pagination } from '../components';
+import { baseStyles, centerStyles } from '../.storybook/styles';
+
+const store = new Store({
+  currentPage: 1,
+  numItems: 200,
+});
+
+const handlePageChange = (page) => {
```
InstaK: PP: `const handlePageChange = page => {`

components/pagination/Pagination.js
```diff
@@ -0,0 +1,130 @@
...
+            </li>
+          )}
+          {iterator.map(page => {
+            return (
+              <li key={page} className={theme['list-item']}>
+                {('' + page).indexOf('ellipsis') === -1 ? (
+                  <LinkButton
+                    className={page === currentPage ? theme['current'] : ''}
+                    inverse={inverse}
+                    label={page}
```
InstaK: ``label={$page}`` will remove the warning for the label prop not being a string. Or you could also allow numbers for the label in `LinkButton` of course

components/pagination/Pagination.js
```diff
@@ -0,0 +1,130 @@
...
+    return Math.ceil(this.props.numItems / this.props.numItemsPerPage);
+  }
+
+  handleChange(page) {
+    this.props.onPageChange(page);
+  }
+
+  render() {
+    const { className, currentPage, inverse, maxNumPagesVisible, nextPageText, prevPageText, ...others } = this.props;
+    const numPages = this.getNumPages();
```
InstaK: PP: A bit overkill to have a function doing this? If you destructure `numItems` and `numItemsPerPage` you could just do `const numPages = Math.ceil(numItems / numItemsPerPage);`. I think it's too much because you're not going to use this function more than once in this component. But I get why you would split it up though :)

components/pagination/Pagination.js
```diff
@@ -0,0 +1,130 @@
...
+      return null;
+    }
+
+    const classNames = cx(theme['pagination'], { [theme['inverse']]: inverse }, className);
+    const iterator = [currentPage];
+    const rest = omit(others, ['numItems', 'numItemsPerPage']);
+
+    let counter = 0;
+
+    while (iterator.length < maxNumPagesVisible && iterator.length < numPages) {
```
InstaK: ++ Works really good 😄

components/pagination/theme.css
```diff
@@ -0,0 +1,54 @@
+@import '@teamleader/ui-colors';
+@import '@teamleader/ui-typography';
+@import '@teamleader/ui-utilities';
+
+:root {
+  --ellipsis-min-width: calc(3 * var(--unit));
+}
+
+.pagination {
+  &:not(.inverse) {
```
InstaK: this `:not` is irrelevant since you're just overwriting all the properties in the `.inverse` scope, you can just remove the `:not` and everything will still work as before.

components/pagination/Pagination.js
driesd: Good idea, and so I can drop the `omit` part!

 :ok: by @driesd on 20/12/2017 08:50:52

components/pagination/theme.css
driesd: @InstaK, I feel you, but here's the difference:

**With :not**
schermafdruk 2017-12-20 09 02 56

**Without :not**
schermafdruk 2017-12-20 08 56 14

Why would we prefer to apply css that is overridden anyway?

components/pagination/Pagination.js
driesd: I updated LinkButton and so allowed label to be a number as well.

components/pagination/theme.css
InstaK: Fair enough, thanks for clarifying!

 :ok: by @InstaK on 20/12/2017 10:25:09

components/pagination/Pagination.js
```diff
@@ -0,0 +1,126 @@
...
+          {iterator.map(page => {
+            return (
+              <li key={page} className={theme['list-item']}>
+                {('' + page).includes('ellipsis') ? (
+                  <TextSmall className={theme['ellipsis']} element="span">
+                    ...
+                  </TextSmall>
+                ) : (
+                  <LinkButton
+                    className={page === currentPage ? theme['current'] : ''}
```
nickwaelkens: Why not use `&&` here instead of `?` and dummy `:`?

components/pagination/Pagination.js
```diff
@@ -0,0 +1,126 @@
...
+                inverse={inverse}
+                label={prevPageText}
+                onClick={() => onPageChange(currentPage - 1)}
+              />
+            </li>
+          )}
+          {iterator.map(page => {
+            return (
+              <li key={page} className={theme['list-item']}>
+                {('' + page).includes('ellipsis') ? (
```
nickwaelkens: PP: Cast to string using `String(...)` or `.toString()`, and if you know for sure that it has to start with `ellipsis`, use `.startsWith()` 👍

components/pagination/Pagination.js
```diff
@@ -0,0 +1,126 @@
...
+    iterator[0] = 1;
+    iterator[iterator.length - 1] = numPages;
+
+    // should we replace the 2nd and 2nd last by ellipsis?
+    if (iterator.length > 1 && iterator[1] !== 2) {
+      iterator[1] = 'ellipsis-left';
+    }
+    if (iterator.length > 1 && iterator[iterator.length - 2] !== numPages - 1) {
+      iterator[iterator.length - 2] = 'ellipsis-right';
+    }
```
nickwaelkens: This seems to work quite well, but is very hard to grasp. Do you think it's possible to refactor this to be more readable?

components/pagination/Pagination.js
driesd: :ok: by @driesd on 20/12/2017 12:47:15

components/pagination/Pagination.js
lowiebenoot: This is actually some code I've written in project planning. I kind of broke my head on it back then, and this was actually the easiest solution that I could find. There's probably a better way (I don't even doubt about it), but it works 😄  But feel free to come up with a better solution 💪 💅

components/pagination/Pagination.js
```diff
@@ -0,0 +1,128 @@
...
+  render() {
+    const {
+      className,
+      currentPage,
+      inverse,
+      maxNumPagesVisible,
+      nextPageText,
+      numItems,
+      numItemsPerPage,
+      onPageChange,
```
nickwaelkens: Does it make sense to treat this component more like an `input`, and rename this from `onPageChange` just to `onChange`? The component is called `Pagination`, so `page` is implied. WDYT? 🤔

components/pagination/Pagination.js
```diff
@@ -0,0 +1,128 @@
...
+    currentPage: PropTypes.number.isRequired,
+    onPageChange: PropTypes.func.isRequired,
+    prevPageText: PropTypes.string,
+    className: PropTypes.string,
+  };
+
+  static defaultProps = {
+    currentPage: 1,
+    inverse: false,
+    numItemsPerPage: 10,
```
lowiebenoot: I'm in doubt whether this component should know how many items there are on a page. Maybe we should just pass `numPages` instead?

InstaK: I think he wants to achieve the `show xx items`? If people want to render all items at once or not, then it will need to know this, right?

lowiebenoot: Correct me if I'm wrong, but that's not a part of the pagination? It's not in current spec either.

InstaK: Hmm... Kind of a tough one? It does change the pagination, (more items per page decreases the amount of pages needed to show it) but indeed if you see it as a module (which you should), then pagination should be something split up. Okay I think it would be better to remove it from the Pagination component.

Also Q @lowiebenoot : If it's not in the spec but it will probably be implemented later, do you provide it or not?

lowiebenoot: Another argument is that you might not know the amount of items. An API might just give you the amount of pages. It would be ridiculous to fake the amount of items by doing `numItemsPerPage * numPages`.

lowiebenoot: If it's not in the spec, but it will probably be implemented later, you might take it into account already. But it's also not bad to not take it into account yet, and just refactor it later, when it is implemented. There's no right or wrong here, it just depends on the case.

components/pagination/Pagination.js
```diff
@@ -0,0 +1,128 @@
...
+                    onClick={() => onChange(page)}
+                    size="small"
+                  />
+                )}
+              </li>
+            );
+          })}
+          {currentPage < numPages && (
+            <li className={theme['list-item']}>
+              <LinkButton
```
lowiebenoot: I've been thinking about this component, and with the current implementation, I actually can't do nothing in the invoicing module. 

In my case the pagination buttons have to be React router links, and not just the fixed LinkButton. So we need to be able to pass some custom element or something like that. 

Current implementation uses onClick and onChange, but in my case I don't really want this, but it might be necessary in other implementations. Not sure how we could solve this.

lowiebenoot:
I changed the implementation of the pagination. It now accepts a function als children, which will then be called from within the pagination. This way you can render a custom element as pagination "buttons".

Please review.

PR #168, opened 2017-12-22
ButtonGroup: removed alignItems prop

driesd:
Since Box can accept flexbox props, we can remove this responsibility from ButtonGroup.

### DIFF
--- a/stories/dialogs.js
+++ b/stories/dialogs.js
@@ -43,7 +43,7 @@ storiesOf('Dialogs', module)
           <Box padding={4}>
             <TextBody>Here you can add arbitrary content.</TextBody>
           </Box>
-          <ButtonGroup alignItems="right" padding={4}>
+          <ButtonGroup justifyContent="flex-end" padding={4}>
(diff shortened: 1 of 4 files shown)


stories/dialogs.js
```diff
@@ -43,7 +43,7 @@ storiesOf('Dialogs', module)
           <Box padding={4}>
             <TextBody>Here you can add arbitrary content.</TextBody>
           </Box>
-          <ButtonGroup alignItems="right" padding={4}>
+          <ButtonGroup justifyContent="flex-end" padding={4}>
```
InstaK: Q: Why exactly do you change `align-items: right` to `justify-content: flex-end`?

lowiebenoot: Because `Box` now supports flex props. If you check the component `alignItems="right" was adding a `align-right` class, which contained `justify-content: flex-end`. So the end result is exactly the same. :+1:

 :ok: by @lowiebenoot on 28/12/2017 19:01:23

PR #181, opened 2018-01-10
Excessive margin on Tab Elements fix

InstaK:
### Added
- Class for TabGroup

### Changed
- Removed the margin off each element and added CSS to the TabGroup which adds margin-left to each child and margin-right to the last child. This is how it was intended by design, I just didn't implement it correctly before.
- Changed the story in order for TabGroup to not have Island as child but Tab-elements

### DIFF
--- a/components/tab/theme.css
+++ b/components/tab/theme.css
@@ -1,5 +1,16 @@
 @import '@teamleader/ui-colors';
 @import '@teamleader/ui-animations';
+@import '@teamleader/ui-utilities';
+
+.tab-group {
+  & > * {
+    margin-left: var(--spacer-small);
+  }
+
+  & > *:last-child {
+    margin-right: var(--spacer-small);
+  }
+}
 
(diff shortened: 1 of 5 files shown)


components/tab/theme.css
```diff
@@ -1,5 +1,16 @@
...
+
+.tab-group {
+  & > * {
+    margin-left: var(--spacer-small);
+  }
+
+  & > *:last-child {
+    margin-right: var(--spacer-small);
+  }
+}
```
lowiebenoot: You can nest the last-child into the other selector.

```css
.tab-group {
   & > * {
     margin-left: var(--spacer-small);

     &:last-child {
       margin-right: var(--spacer-small);
     }
   }
 }
```

InstaK: Nice. Should've seen that it was 2 times the same.

 :ok: by @InstaK on 10/01/2018 11:51:21

PR #199, opened 2018-02-20
Input component - enhancements

driesd:
### Description
* implement `new props`:
  * connectedLeft
  * connectedRight
  * autoFocus
  * min
  * max
  * format
  * precision
* use our own `Button` component for the number field spinner controls
* `validation icon` is now placed before of the `validation text`
* hide browser native `spinner controls` in Firefox

schermafdruk 2018-03-01 11 44 50

#### Screenshot after this PR
schermafdruk 2018-02-27 14 49 34

schermafdruk 2018-02-27 14 47 46

schermafdruk 2018-02-27 14 47 59

### Breaking changes
None


components/input/Input.js
```diff
@@ -73,42 +83,91 @@ export default class Input extends Component {
...
+    const { step } = this.props;
+
+    const number = this.toNumber((this.state.value || 0) + step * n);
+
+    if (number !== this.state.value) {
+      this.setState({ value: number });
+      return true;
+    }
+
+    return false;
```
InstaK: I don't think you have to return anything here? I think you should change the function name to something like `updateStep`, get rid of the if statement where you check on `number !== this.state.value` because I assume React is smart enough to not change the state if it notices that the value will be the same. So I think you can just rewrite it as follows:

```js
updateStep(n){
  const { step } = this.props;
  const number = this.toNumber((this.state.value || 0) + step * n);

  this.setState({ value: number });
}
```

InstaK:
* When I'm at the number input label, I can't type once there's 2 decimals, but I can use the numbers 6, 7, 8 and 9 on the numpad to increment the value by 0.01. This also allows me to go to 10.01 and higher even though it looks like 10 is the maximum there.

* NTH: Might be nice to be able to use alt and shift with the up and down arrows in order to increment or decrement in smaller or bigger steps. As a reference: incrementing pixels in the chrome inspector: Arrow up = +1, Shift + Arrow up = +10, Alt + Arrow up = + 0.1

components/input/Input.js
```diff
@@ -73,42 +83,91 @@ export default class Input extends Component {
...
+  toNumber(number) {
+    const { max, min, precision } = this.props;
+
+    let float = parseFloat(number);
+
+    if (isNaN(float) || !isFinite(float)) {
+      float = 0;
+    }
+
+    const q = Math.pow(10, precision === null ? 10 : precision);
```
nickwaelkens: What does `q` stand for here?

components/input/Input.js
nickwaelkens: Correct, I see no reason why to return `true` or `false` here either?
Preferably, this would also go in a functional `setState` where the `prevState` is already known in that scope.

components/input/Input.js
```diff
@@ -73,42 +83,91 @@ export default class Input extends Component {
...
+
+    let formattedNumber = this.toNumber(number);
+
+    if (precision !== null) {
+      formattedNumber = number.toFixed(precision);
+    }
+
+    formattedNumber += '';
+
+    return formattedNumber;
```
nickwaelkens: Maybe renaming this to something like `formatNumber` so it's easier to understand what it's actually doing? Also, why not be explicit when returning a string and actually `return String(formattedNumber)`?

components/input/Input.js
```diff
@@ -41,6 +48,9 @@ export default class Input extends Component {
     inverse: false,
     disabled: false,
     placeholder: '',
+    precision: null,
+    min: -9007199254740991,
+    max: 9007199254740991,
```
nickwaelkens: What the actual fuck. Do we really need these here, and why not use `Number.MAX_SAFE_INTEGER`/`Number.MIN_SAFE_INTEGER` if necessary?

components/input/Input.js
```diff
@@ -59,8 +69,8 @@ export default class Input extends Component {
   }
 
   componentWillUpdate(props, state) {
-    if (props.onChange && state.value) {
-      props.onChange(state.value);
+    if (props.input && props.input.onChange && state.value) {
+      props.input.onChange(state.value);
```
nickwaelkens: Thanks! This means we will need some kind of helper to create a `redux-form`-esque input field for situations where we don't use it so this API works everywhere. Getting there... 👏

components/input/Input.js
```diff
@@ -73,42 +83,91 @@ export default class Input extends Component {
...
+    if (isNaN(float) || !isFinite(float)) {
+      float = 0;
+    }
+
+    const q = Math.pow(10, precision === null ? 10 : precision);
+
+    float = Math.min(Math.max(float, min), max);
+    float = Math.round(float * q) / q;
+
+    return float;
```
nickwaelkens: I feel like most of these helper functions don't really belong in (this) component.

driesd:
I've made two issues to fix the remaining comments:
**Increasing/decreasing with shift key**
#216 

**Increase value by 0.01 beyond it's maximum**
#217

PR #200, opened 2018-02-20
Icon component

driesd:
### Description
We need an Icon component so we can trigger a Tooltip from any Icon in ui-icons.

**Other advantages:**
* We are able to render any Icon in any color (teal, mint, ruby, aqua,...) and any tint (lightest, light, normal, ...).
* Some css could be eliminated from other components that use an icon.

### Breaking changes
None.

### DIFF
--- a/components/icon/index.js
+++ b/components/icon/index.js
@@ -1,0 +1,3 @@
+import Icon from './Icon';
+
+export default Icon;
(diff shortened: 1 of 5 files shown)


components/icon/index.js
```diff
@@ -0,0 +1,3 @@
+import Icon from './Icon';
+
+export default Icon;
```
InstaK: PP: `export default from './Icon';`

PR #201, opened 2018-02-20
Tag: add the children prop so we can show a Tooltip

driesd:
### Description
In order to be able using a Tag to trigger a Tooltip, we need to add the `children` prop in the Tag component.

### Breaking changes
**Label** has been removed. Use `children` instead.

### DIFF
--- a/components/tag/Tag.js
+++ b/components/tag/Tag.js
@@ -11,7 +11,7 @@ class Tag extends PureComponent {
     children: PropTypes.any,
     className: PropTypes.string,
     inverse: PropTypes.bool,
-    label: PropTypes.string.isRequired,
+    label: PropTypes.string,
     onLabelClick: PropTypes.func,
     onRemoveClick: PropTypes.func,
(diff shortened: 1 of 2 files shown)


components/tag/Tag.js
```diff
@@ -42,9 +43,13 @@ class Tag extends PureComponent {
         {onLabelClick ? (
           <Button className={theme['label-button']} onClick={onLabelClick} level="outline" inverse={inverse}>
             {label}
+            {children}
           </Button>
         ) : (
-          <span className={theme['label']}>{label}</span>
+          <span className={theme['label']}>
+            {label}
+            {children}
```
InstaK: Q: What if label or children now isn't provided? I'm not entirely sure what will happen then... Shouldn't there be checks now?

driesd: **A:** I removed the `label` prop and made `children` required. In this case we don't need an extra check. A console warning will be thrown.

 :ok: by @driesd on 20/02/2018 17:05:11

PR #204, opened 2018-02-21
Avatar: component & story updates

driesd:
### Description
* add the `children` prop so we can show a Tooltip
* remove `bullet` prop
* remove `counter` prop
* Storybook: add story for an Avatar with a Tooltip

### Breaking changes
* **bullet** prop has been removed. Use `children` instead.
* **counter** prop has been removed. Use `children` instead.

### DIFF
--- a/components/avatar/theme.css
+++ b/components/avatar/theme.css
@@ -17,7 +17,7 @@
   position: relative;
 }
 
-.connected {
+.children {
   position: absolute;
   z-index: 1;
 }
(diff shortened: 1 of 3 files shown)


components/avatar/theme.css
```diff
@@ -17,8 +17,7 @@
   position: relative;
 }
 
-.bullet,
-.counter {
+.connected {
```
lowiebenoot: Can you explain the name `connected`?

driesd: I called it `connected` because it's visually connected to the left upper corner of the avatar image. I can call it `.children` if you prefer :)

lowiebenoot: I guess that would be better.

PR #220, opened 2018-03-07
Remove named export in component files

driesd:
### Description
This PR removes the named export from component files.

### Breaking changes
None, if you import our components from `@teamleader/ui`.

**Note:** Please update your code if you're importing components straight from  component file (`@teamleader/ui/component/Component.js`).


InstaK:
Also, could you also explain in your descriptions *why* you are doing something? Cause I don't know why this is done now :)

PR #233, opened 2018-03-27
Fix input component not setting the passed value prop

timdegroote:
### Description
The passed value to the input component was always undefined as it was picked directly from `props` instead of `props.input`.

### DIFF
--- a/components/input/Input.js
+++ b/components/input/Input.js
@@ -83,8 +83,17 @@ export default class Input extends Component {
     this.handleDecreaseValue = ::this.handleDecreaseValue;
 
     this.state = {
-      value: props.type === 'number' ? Number(props.value) || '' : props.value || '',
+      value: this.parsePropsValue(),
     };
+  }
+
+  parsePropsValue() {
+    const { input: { value }, type } = this.props;
+    if (type === 'number') {
+      return this.toNumber(value);
+    }
+
+    return value || '';
   }
 
   componentWillUpdate(props, state) {


components/input/Input.js
```diff
@@ -83,10 +83,19 @@ export default class Input extends Component {
     this.handleDecreaseValue = ::this.handleDecreaseValue;
 
     this.state = {
-      value: props.type === 'number' ? Number(props.value) || '' : props.value || '',
+      value: this.parsePropsValue(),
     };
   }
 
+  parsePropsValue() {
+    const { input: { value }, type } = this.props;
```
InstaK: Q: Is this a way to get a deeper key? So you could also do
```js
const { input: { value: { evenDeeperKey } } = this.props;
```
?

timdegroote: @InstaK Yes indeed!

InstaK: Cool, I was looking for a way to do this yesterday or last week I remember. Is there still a way to rename them then? So If I wanted to rename `evenDeeperKey`, could I do:

timdegroote: In that case you'd have to do something like this:
`const { input: { value: { evenDeeperKey: renamedKey } }, anotherProp } = this.props;`

But this very quickly becomes unreadable :D

InstaK: Fair enough, thanks :)

 :ok: by @InstaK on 27/03/2018 16:14:57

PR #236, opened 2018-03-28
Allow patch updates for our own dependency packages

driesd:
### Description
In this PR we've added `^` to the required version of our own dependency packages. This will cause less releases of UI when releasing one of its dependencies.

### Breaking changes
* None


nickwaelkens:
Going to merge this one by myself, seems OK 👍

PR #237, opened 2018-03-28
Check the existence of the input prop before using its value

timdegroote:
### Description
The change from yesterday broke the input component when used outside of a `redux-form` context. Now we only use `props.input.value` when the `input` prop actually exists.

### DIFF
--- a/components/input/FieldInputPropTypes.js
+++ b/components/input/FieldInputPropTypes.js
@@ -1,0 +1,14 @@
+import PropTypes from 'prop-types';
+
+const FieldInputPropTypes = PropTypes.shape({
+  checked: PropTypes.bool,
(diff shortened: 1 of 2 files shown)


timdegroote:
Ideally these components shouldn't be aware of which form library we're using though.

nickwaelkens:
I know; any idea how we could tackle this issue? A HoC that creates the correct props?

timdegroote:
I agree a HOC is worth looking into, perhaps something we could do during the guild week? :)

nickwaelkens:
If it's annoying us right now, feel free to do it during a "regular" week, really. Or add it to https://github.com/teamleadercrm/guild-frontend/issues/28 👍

PR #238, opened 2018-04-10
[Discussion/suggestions] A possible solution for the portal context issue

lowiebenoot:
In `core` we render components in iframes via `React.createPortal`. Due to this, the components are actually running in a wrong JS context. Their context is the "main" frame, instead of the iframe.

Due to this, the `Tooltip` component for example (probably all components that also use `createPortal`), will render in the wrong place, because it's rendering in the main frame, instead of in the iframe.

I created a quick fix, that solves this problem. I could make a cleaner solution (I guess) with a `PortalWrapper` component or HOC or something like that, but wanted to get some feedback first. Maybe there's a better solution? Maybe the UI package shouldn't care and core should create their own solution (if even possible)?

Also, [Node.ownerDocument](https://developer.mozilla.org/en-US/docs/Web/API/Node/ownerDocument) doesn't work in IE11, but it seems to work, probably because it's polyfilled. 

An example of the tooltip rendering in the wrong position:

screen shot 2018-04-10 at 10 56 42

### DIFF
--- a/components/tooltip/Tooltip.js
+++ b/components/tooltip/Tooltip.js
@@ -62,9 +62,13 @@ const tooltipFactory = () => {
           active: false,
           position: this.props.tooltipPosition,
         };
+      }
 
-        this.tooltipRoot = document.createElement('div');
+      componentDidMount() {
+        this.elementDocument = this.node.ownerDocument;
+        this.tooltipRoot = this.elementDocument.createElement('div');
         this.tooltipRoot.id = 'tooltip-root';
+        this.forceUpdate();
       }
 
       getPosition(element) {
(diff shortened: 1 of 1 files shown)


duivvv:
sounds good, just not 100% confident on `but it seems to work, probably because it's polyfilled` 😃

lowiebenoot:
Yes, core is polyfilled. Wondering if that is good enough, because this means that "consumers" of our UI package, should make sure they polyfill it themselves.

duivvv:
screen shot 2018-04-10 at 14 47 44

ah.

Edge > 12
IE > 6

lowiebenoot:
Ah lol, what a fail 😂 

@nickwaelkens the first render is actually a "fake render" just to create a ref, so I can find out what the actual (owner)document is. Once I have found the document, I force a re-render. 
I could also just put the document (called `elementDocument` here) in the state (which will trigger a re-render), but I didn't really think this belonged in the state.

nickwaelkens:
Fair enough, thanks for clarifying! 👍 
Then this approach seems good, but like you proposed - it belongs in a HoC :)

PR #240, opened 2018-04-11
🔥 Remove inheriting of color for button on hover

InstaK:
### Description
I don't really remember why I added `color: inherit` in [this PR](https://github.com/teamleadercrm/ui/pull/234), I definitely did it for a reason, so it might make the buttons on `/integrations.php` look bad on core, but then it needs to be fixed in a different way. 

screen shot 2018-04-11 at 13 47 23

#### Screenshot after this PR
screen shot 2018-04-11 at 13 47 35

### DIFF
--- a/components/button/theme.css
+++ b/components/button/theme.css
@@ -65,7 +65,6 @@
     &:hover {
       z-index: 1;
       text-decoration: none;
-      color: inherit;
     }
 
     &:active {


InstaK:
Hm, if I remove `color: inherit` the buttons do indeed break on `/integrations.php`.
This happens when hovering over a button without the inheriting:
screen shot 2018-04-11 at 13 56 23


and with inheriting:
screen shot 2018-04-11 at 13 55 47

I think this is because of the color appended to all a tags in core css...

lowiebenoot:
Yes, probably because of core's general `a` styling...

duivvv:
off-core-se

PR #241, opened 2018-04-11
Render the Tooltip in the correct document

lowiebenoot:
## Best to review per commit, because I refactored a useless factory

In `core` we render components in iframes via `React.createPortal`. Due to this, the components are actually running in a wrong JS context. Their context is the "main" frame, instead of the iframe.

Due to this, the `Tooltip` component for example (probably all components that also use `createPortal`), will render in the wrong place, because it's rendering in the main frame, instead of in the iframe.

To fix this I created a `DocumentObjectProvider` which does a "fake and invisible" render first, so we can get the correct `document` from the node. Afterwards we can render again and provide the document through context.

A component wrapped with this HOC can consume this "correct" document from the context and then use it.

### DIFF
--- a/components/hoc/DocumentObjectProvider/DocumentObjectProvider.js
+++ b/components/hoc/DocumentObjectProvider/DocumentObjectProvider.js
@@ -1,0 +1,36 @@
+import React, { PureComponent, createContext } from 'react';
+
+export const Context = createContext();
+
+const DocumentObjectProvider = WrappedComponent => {
+  return class extends PureComponent {
+    componentDidMount() {
+      this.forceUpdate(); // force a re-render because we have the document ref now
+    }
+
+    render() {
+      if (!this.documentRef) {
+        return (
+          <span
+            style={{ display: 'none' }}
+            ref={node => {
+              if (!node) {
+                return;
+              }
+
+              this.documentRef = node.ownerDocument;
+            }}
+          />
+        );
+      }
+
+      return (
+        <Context.Provider value={this.documentRef}>
+          <WrappedComponent {...this.props} />
+        </Context.Provider>
+      );
+    }
+  };
+};
+
+export default DocumentObjectProvider;
--- a/components/tooltip/Tooltip.js
+++ b/components/tooltip/Tooltip.js
@@ -28,203 +28,201 @@ const SIZES = {
   },
 };
 
-const tooltipFactory = () => {
-  return ComposedComponent => {
-    class TooltippedComponent extends Component {
-      static propTypes = {
-        children: PropTypes.node,
-        className: PropTypes.string,
-        onClick: PropTypes.func,
-        onMouseEnter: PropTypes.func,
-        onMouseLeave: PropTypes.func,
-        tooltip: PropTypes.oneOfType([PropTypes.string, PropTypes.node]),
-        tooltipColor: PropTypes.oneOf(['white', 'neutral', 'mint', 'violet', 'ruby', 'gold', 'aqua', 'inverse']),
-        tooltipHideOnClick: PropTypes.bool,
-        tooltipIcon: PropTypes.element,
-        tooltipPosition: PropTypes.oneOf(Object.keys(POSITIONS).map(key => POSITIONS[key])),
-        tooltipShowOnClick: PropTypes.bool,
-        tooltipSize: PropTypes.oneOf(Object.keys(SIZES)),
-        documentObject: PropTypes.object.isRequired,
+const Tooltip = ComposedComponent => {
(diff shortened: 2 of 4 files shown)


components/tooltip/Tooltip.js
```diff
@@ -55,15 +57,15 @@ const tooltipFactory = () => {
...
+      constructor(props) {
         super(...arguments);
 
         this.state = {
           active: false,
           position: this.props.tooltipPosition,
         };
 
-        this.tooltipRoot = document.createElement('div');
+        this.tooltipRoot = props.documentObject.createElement('div');
```
nickwaelkens: PP: Get rid of `props` as argument and just access it through `this.props` (like is being done when setting `this.state = {}`)

components/hoc/DocumentObjectProvider/DocumentObjectProvider.js
```diff
@@ -0,0 +1,36 @@
+import React, { PureComponent, createContext } from 'react';
+
+export const Context = createContext();
+
+const DocumentObjectProvider = WrappedComponent => {
+  return class extends PureComponent {
+    componentDidMount() {
+      this.forceUpdate(); // force a re-render because we have the document ref now
```
nickwaelkens: I think this is like the first time I've ever seen a legit use for `forceUpdate` 👍

components/tooltip/Tooltip.js
```diff
@@ -28,203 +28,201 @@ const SIZES = {
...
-        tooltip: PropTypes.oneOfType([PropTypes.string, PropTypes.node]),
-        tooltipColor: PropTypes.oneOf(['white', 'neutral', 'mint', 'violet', 'ruby', 'gold', 'aqua', 'inverse']),
-        tooltipHideOnClick: PropTypes.bool,
-        tooltipIcon: PropTypes.element,
-        tooltipPosition: PropTypes.oneOf(Object.keys(POSITIONS).map(key => POSITIONS[key])),
-        tooltipShowOnClick: PropTypes.bool,
-        tooltipSize: PropTypes.oneOf(Object.keys(SIZES)),
-        documentObject: PropTypes.object.isRequired,
+const Tooltip = ComposedComponent => {
+  class TooltippedComponent extends Component {
```
InstaK: I guess this can also be a PureComponent since shouldComponentUpdate is not being used? Or is this still needed because you were using forceUpdate?

components/tooltip/Tooltip.js
```diff
@@ -28,203 +28,201 @@ const SIZES = {
...
+    static propTypes = {
```
InstaK: Better to have these outside of the Class already

InstaK:
@lowiebenoot I did 🙈 https://puu.sh/A1t4t/c74ca10611.png

components/tooltip/Tooltip.js
```diff
@@ -28,203 +28,201 @@ const SIZES = {
...
+      children: PropTypes.node,
+      className: PropTypes.string,
+      onClick: PropTypes.func,
+      onMouseEnter: PropTypes.func,
+      onMouseLeave: PropTypes.func,
+      tooltip: PropTypes.oneOfType([PropTypes.string, PropTypes.node]),
+      tooltipColor: PropTypes.oneOf(['white', 'neutral', 'mint', 'violet', 'ruby', 'gold', 'aqua', 'inverse']),
+      tooltipHideOnClick: PropTypes.bool,
+      tooltipIcon: PropTypes.element,
+      tooltipPosition: PropTypes.oneOf(Object.keys(POSITIONS).map(key => POSITIONS[key])),
```
timdegroote: PP: I think a quicker way of doing the same thing would be doing `Object.values(POSITIONS)`

components/tooltip/Tooltip.js
```diff
@@ -28,203 +28,201 @@ const SIZES = {
...
-          active: false,
-          position: this.props.tooltipPosition,
-        };
-
-        this.tooltipRoot = props.documentObject.createElement('div');
-        this.tooltipRoot.id = 'tooltip-root';
-      }
+      if (tooltipPosition === POSITIONS.HORIZONTAL) {
+        const origin = element.getBoundingClientRect();
+        const { width: windowWidth } = getViewport();
```
timdegroote: PP: What do you think of moving `const origin` and `const { ... } = getViewPort()` outside of the if block? It looks like the result of those is used in both the `if` and `else` blocks.

lowiebenoot:
@timdegroote we could do that in a separate PR. I actually didn't really touch that code, I just changed the indenting.

PR #242, opened 2018-04-12
Fix input triggering onChange for its old value when receiving props

timdegroote:
### Description
The Input component triggered the onChange with its previous value when receiving props. Thus not taking a new value passes from outside. This PR fixes that by not relying solely on `componentWillUpdate` to update its state.

### DIFF
--- a/components/input/Input.js
+++ b/components/input/Input.js
@@ -1,4 +1,4 @@
-import React, { Component, createElement } from 'react';
+import React, { PureComponent, createElement } from 'react';
 import PropTypes from 'prop-types';
 import cx from 'classnames';
 import omit from 'lodash.omit';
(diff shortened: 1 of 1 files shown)


duivvv:
If we can we should stay away from componentWill[...] functions, they will (pun unintended) be removed in later versions of React

components/input/Input.js
```diff
@@ -141,12 +152,8 @@ export default class Input extends Component {
...
   updateStep(n) {
     const { step } = this.props;
-
-    this.setState(prevState => {
-      const number = this.toNumber((prevState.value || 0) + step * n);
-
-      return { value: number };
-    });
+    const { value = 0 } = this.state;
+    this.updateValue(value + step * n);
```
nickwaelkens: And made this more readable while he's at it. 👏

timdegroote:
@nickwaelkens Another option was to pull them out of the class and call them as normal functions. There is probably a better way of doing what these methods are currently doing though. But would have to spend some time looking into this component properly.

PR #251, opened 2018-05-29
Label: add boxProps to the Box container of the Label component

driesd:
### Description
I pick the boxProps from `...others` and pass them to the Box container of our Label component.

### Breaking changes
None.

### DIFF
--- a/components/label/Label.js
+++ b/components/label/Label.js
@@ -1,7 +1,7 @@
 import React, { Fragment, PureComponent } from 'react';
 import PropTypes from 'prop-types';
 import Input from '../input';
-import Box from '../box';
+import Box, { pickBoxProps } from '../box';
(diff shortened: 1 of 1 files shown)


components/label/Label.js
```diff
@@ -28,23 +28,38 @@ export default class Label extends PureComponent {
...
+      connectedRight,
+      className,
+      inverse,
+      helpText,
+      required,
+      size,
+      ...others
+    } = this.props;
+
+    const boxProps = pickBoxProps({ ...others });
```
lowiebenoot: you can just pass `others` instead of creating a new object and destructuring `others` into it.

lowiebenoot: Also, I don't think you should pick the boxProps here. Now there might be props left out, not being passed to the Box anymore.

PR #253, opened 2018-05-29
Lockscroll props passing

InstaK:
### Description
The Overlay component can have its lockScroll set to false or true, but it's not possible to pass that choice down to the Popover component which uses Overlay. Now it is.

### DIFF
--- a/components/popover/Popover.js
+++ b/components/popover/Popover.js
@@ -48,7 +48,7 @@ const factory = (axis, calculatePositions, Overlay) => {
       document.body.removeChild(this.popoverRoot);
     }
 
-    componentDidUpdate(prevProps, prevState) {
+    componentDidUpdate(prevProps) {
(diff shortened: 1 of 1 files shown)


components/popover/Popover.js
```diff
@@ -34,8 +35,9 @@ const factory = (axis, calculatePositions, Overlay) => {
       active: true,
       backdrop: 'dark',
       color: 'neutral',
+      lockScroll: true,
       offsetCorrection: 0,
-      tint: 'lightest',
+      tint: 'normal',
```
lowiebenoot: Why did this change?

InstaK: Oh wow, I did this to try out some errors in ui when it was linked, probably shouldn't have been in here.

 :ok: by @InstaK on 29/05/2018 15:50:13

PR #256, opened 2018-05-29
DatePicker component

driesd:
### Description
A `DatePicker` component as wrapper for `react-day-picker`.

#### Screenshot after this PR
schermafdruk 2018-03-19 11 33 05

### Breaking changes
None

### DIFF
--- a/components/datepicker/DatePicker.js
+++ b/components/datepicker/DatePicker.js
@@ -1,0 +1,65 @@
+import React, { PureComponent } from 'react';
+import PropTypes from 'prop-types';
+import DayPicker from 'react-day-picker';
+import NavigationBar from './NavigationBar';
+import WeekDay from './WeekDay';
+import { convertModifiersToClassnames } from './utils';
+import cx from 'classnames';
+import theme from './theme.css';
+
+class DatePicker extends PureComponent {
+  state = {
+    selectedDate: null,
+  };
+
+  componentWillMount() {
+    const { selectedDate } = this.props;
+
+    if (selectedDate) {
+      this.setState({ selectedDate });
+    }
+  }
+
+  handleDayClick = day => {
+    this.setState(
+      {
+        selectedDate: day,
+      },
+      this.props.onChange({ selectedDate: day }),
+    );
+  };
+
+  render() {
+    const { className, modifiers, size, ...others } = this.props;
+    const { selectedDate } = this.state;
+
+    const classNames = cx(theme['date-picker'], theme[`is-${size}`], className);
+
+    return (
+      <DayPicker
+        className={classNames}
+        classNames={theme}
+        modifiers={convertModifiersToClassnames(modifiers, theme)}
+        navbarElement={<NavigationBar size={size} />}
+        onDayClick={this.handleDayClick}
+        selectedDays={selectedDate}
+        weekdayElement={<WeekDay size={size} />}
+        {...others}
+      />
+    );
+  }
+}
+
+DatePicker.propTypes = {
+  className: PropTypes.string,
+  modifiers: PropTypes.object,
+  onChange: PropTypes.func,
+  selectedDate: PropTypes.instanceOf(Date),
+  size: PropTypes.oneOf(['small', 'medium', 'large']),
+};
+
+DatePicker.defaultProps = {
+  size: 'medium',
+};
+
+export default DatePicker;
--- a/components/datepicker/WeekDay.js
+++ b/components/datepicker/WeekDay.js
@@ -1,0 +1,22 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+const WeekDay = ({ weekday, className, localeUtils, locale, size }) => {
+	const weekdayName = localeUtils.formatWeekdayLong(weekday, locale);
+
+	return (
+			<div className={className} title={weekdayName}>
+				{size === 'large' ? weekdayName.slice(0, 2) : weekdayName.slice(0, 1)}
+			</div>
+	);
+};
+
+WeekDay.propTypes = {
+	className: PropTypes.string,
+	locale: PropTypes.string,
+	localeUtils: PropTypes.object,
+	size: PropTypes.oneOf(['small', 'medium', 'large']),
+	weekday: PropTypes.number,
+};
+
+export default WeekDay;
--- a/components/datepicker/WeekDay.js
+++ b/components/datepicker/WeekDay.js
@@ -2,21 +2,21 @@ import React from 'react';
 import PropTypes from 'prop-types';
 
 const WeekDay = ({ weekday, className, localeUtils, locale, size }) => {
-	const weekdayName = localeUtils.formatWeekdayLong(weekday, locale);
+  const weekdayName = localeUtils.formatWeekdayLong(weekday, locale);
 
-	return (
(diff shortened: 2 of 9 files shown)


components/datepicker/utils.js
```diff
@@ -0,0 +1,13 @@
+export const convertModifiersToClassnames = (modifiers, theme) => {
+	if (!modifiers) {
+		return;
+	}
+
+	const convertedModifiers = {};
+
+	Object.keys(modifiers).forEach(key => {
```
lowiebenoot: You can use `array.reduce` here

```
  return Object.keys(modifiers).reduce((convertedModifiers, key) => {
    return {
      ...convertedModifiers,
      [theme[key]]: modifiers[key],
    };
  }, {});
```

lowiebenoot: You can use `array.reduce` here

```
  return Object.keys(modifiers).reduce((convertedModifiers, key) => {
    return {
      ...convertedModifiers,
      [theme[key]]: modifiers[key],
    };
  }, {});
```

stories/datePicker.js
```diff
@@ -0,0 +1,38 @@
...
+import MomentLocaleUtils from 'react-day-picker/moment';
+
+const languages = ['da', 'de', 'fr', 'en', 'es', 'fi', 'it', 'nl', 'pt', 'pl', 'sv'];
+const sizes = ['small', 'medium', 'large'];
+
+const preSelectedDate = DateTime.local().plus({days: 3}).toJSDate();
+
+
+storiesOf('DatePicker', module)
+		.addDecorator((story, context) => withInfo({TableComponent: PropTable})(story)(context))
```
lowiebenoot: This is indented with tabs instead of spaces (and our linter isn't linting the stories, maybe we should lint those as well).

components/datepicker/DatePicker.js
```diff
@@ -0,0 +1,65 @@
...
+      this.setState({ selectedDate });
+    }
+  }
+
+  handleDayClick = day => {
+    this.setState(
+      {
+        selectedDate: day,
+      },
+      this.props.onChange({ selectedDate: day }),
```
timdegroote: this is defined as a callback to setState but won't it be executed immediately as it's not wrapped in a function?

components/datepicker/DatePicker.js
```diff
@@ -0,0 +1,65 @@
...
+    return (
+      <DayPicker
+        className={classNames}
+        classNames={theme}
+        modifiers={convertModifiersToClassnames(modifiers, theme)}
+        navbarElement={<NavigationBar size={size} />}
+        onDayClick={this.handleDayClick}
+        selectedDays={selectedDate}
+        weekdayElement={<WeekDay size={size} />}
+        {...others}
```
timdegroote: PP: It might be better to put `{...others}` before defining your props here so that the props which you define (like `classNames`, `onDayClick`, etc) can't be overwritten from outside. Unless of course if that's something we would allow.

components/datepicker/WeekDay.js
```diff
@@ -0,0 +1,22 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+const WeekDay = ({ weekday, className, localeUtils, locale, size }) => {
+  const weekdayName = localeUtils.formatWeekdayLong(weekday, locale);
+
+  return (
+    <div className={className} title={weekdayName}>
+      {size === 'large' ? weekdayName.slice(0, 2) : weekdayName.slice(0, 1)}
```
timdegroote: PP could write this a bit shorter (and IMO I think it's easier to read) by putting the if around the part that's different:
`{weekdayName.slice(0, size === 'large' ? 2 : 1)}`

components/datepicker/DatePicker.js
```diff
@@ -0,0 +1,65 @@
...
+import { convertModifiersToClassnames } from './utils';
+import cx from 'classnames';
+import theme from './theme.css';
+
+class DatePicker extends PureComponent {
+  state = {
+    selectedDate: null,
+  };
+
+  componentWillMount() {
```
nickwaelkens: `componentWillMount` is deprecated (16.3); checking out [this article](https://reactjs.org/blog/2018/03/27/update-on-async-rendering.html#initializing-state), initializing state can be done right in the component without lifecycle method
```jsx
class DatePicker extends PureComponent {
	state = {
		selectedDate: this.props.selectedDate || null,
	};
}
```

components/datepicker/utils.js
```diff
@@ -0,0 +1,12 @@
+export const convertModifiersToClassnames = (modifiers, theme) => {
+  if (!modifiers) {
+    return;
+  }
```
timdegroote: PP: It might be more predictable to initialise modifiers as an empty object if it's passed as `undefined`. That way this function will always return an object making it more consistent instead of sometimes an object, and sometimes `undefined` in the case where `modifiers` is `undefined`.

components/datepicker/WeekDay.js
```diff
@@ -0,0 +1,22 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+const WeekDay = ({ weekday, className, localeUtils, locale, size }) => {
```
nickwaelkens: Functional components do not have a built-in comparison (`shouldComponentUpdate`). Do we want that?

components/datepicker/DatePicker.js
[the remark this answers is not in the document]
driesd: I eliminated `getDerivedStateFromProps` and simplified the code by doing:
```
state = {
    selectedDate: this.props.selectedDate,
  };
```

components/datepicker/DatePicker.js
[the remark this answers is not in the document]
driesd: Done!

driesd: I solved this by using the `componentDidUpdate` method as @duivvv mentioned.
```
componentDidUpdate() {
    this.props.onChange({ selectedDate: this.state.selectedDate });
  }
```

components/datepicker/WeekDay.js
driesd: Valid point. I've changed this to a `class` extending `PureComponent`.

 :ok: by @driesd on 30/05/2018 14:31:31

components/datepicker/DatePicker.js
```diff
@@ -9,24 +9,15 @@ import theme from './theme.css';
...
   };
 
-  componentWillMount() {
-    const { selectedDate } = this.props;
-
-    if (selectedDate) {
-      this.setState({ selectedDate });
-    }
+  componentDidUpdate() {
+    this.props.onChange({ selectedDate: this.state.selectedDate });
```
timdegroote: When is the `componentDidUpdate` callback triggered? Say a parent component passes a new `selectedDate` to the datePicker component. Once the datePicker has re-rendered based on that `prop` change won't it trigger `componentDidUpdate` and thus also trigger the changeHandler passed from the parent even though the value didn't change?

duivvv: it's triggered on render

driesd: I `eliminated` componentDidUpdate because we only want `this.props.onChange` to be called when a user actually clicks on a day.

 :ok: by @driesd on 30/05/2018 16:39:04

PR #257, opened 2018-05-31
DatePickerRange component

driesd:
### Description
A `DatePickerRange` component as wrapper for `react-day-picker`.

#### Screenshot after this PR
schermafdruk 2018-05-31 12 25 27

### Breaking changes
None

### DIFF
--- a/components/datepicker/DatePickerRange.js
+++ b/components/datepicker/DatePickerRange.js
@@ -11,7 +11,7 @@ class DatePickerRange extends PureComponent {
   state = {
     selectedStartDate: null,
     selectedEndDate: null,
-    enteredTo: null,
+    mouseEnteredEndDate: null,
   };
 
   static getDerivedStateFromProps(props, state) {
@@ -23,7 +23,7 @@ class DatePickerRange extends PureComponent {
       return {
         selectedStartDate: props.selectedRange.selectedStartDate,
         selectedEndDate: props.selectedRange.selectedEndDate,
-        enteredTo: props.selectedRange.selectedEndDate,
+        mouseEnteredEndDate: props.selectedRange.selectedEndDate,
       };
     }
 
@@ -38,7 +38,7 @@ class DatePickerRange extends PureComponent {
         {
           selectedStartDate: day,
           selectedEndDate: null,
-          enteredTo: null,
+          mouseEnteredEndDate: null,
         },
         () => this.props.onChange({ selectedStartDate: day, selectedEndDate: null }),
       );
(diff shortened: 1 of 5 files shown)


components/datepicker/DatePickerRange.js
```diff
@@ -0,0 +1,107 @@
...
+    const { className, size, ...others } = this.props;
+    const { from, enteredTo } = this.state;
+    const modifiers = { from, to: enteredTo };
+    const selectedDays = [from, { from, to: enteredTo }];
+
+    const classNames = cx(
+      theme['date-picker'],
+      theme[`is-${size}`],
+      {
+        [theme['has-range']]: hasRange(selectedDays),
```
nickwaelkens: `selectedDays` is an array, but in the implementation of `hasRange` you're doing an `Object.keys`, I'm confused 😄

components/datepicker/utils.js
```diff
@@ -10,3 +12,35 @@ export const convertModifiersToClassnames = (modifiers, theme) => {
...
+
+export const hasRange = selectedDays => {
+  let containsFrom = false;
+  let containsTo = false;
+
+  if (!selectedDays || typeof selectedDays !== 'object') {
+    return false;
+  }
+
+  Object.keys(selectedDays).map(key => {
```
nickwaelkens: Might want to use `Object.entries` here if you need both keys _and_ value?

components/datepicker/DatePickerRange.js
```diff
@@ -0,0 +1,107 @@
...
+import WeekDay from './WeekDay';
+import { convertModifiersToClassnames, hasRange, isSelectingFirstDay } from './utils';
+import cx from 'classnames';
+import theme from './theme.css';
+
+class DatePickerRange extends PureComponent {
+  state = {
+    from: null,
+    to: null,
+    enteredTo: null,
```
lowiebenoot: I don't understand the `to` and `enteredTo` stuff. It's not clear what it actually means.

nickwaelkens: +1

driesd: I renamed these props to something more explicit.

components/datepicker/DatePickerRange.js
driesd: I realized that I could completely eliminate this `hasRange` function since we already know we're going to select a range. The components name therefore is `DatePickerRange`.

 :ok: by @driesd on 04/06/2018 12:20:01

components/datepicker/utils.js
driesd: I removed this util function since we're nog using it anymore.

 :ok: by @driesd on 04/06/2018 12:20:33

PR #258, opened 2018-05-31
DatePickerInput component

driesd:
### Description
A `DatePickerInput` component as wrapper for `react-day-picker`.

#### Screenshot after this PR
schermafdruk 2018-05-31 16 42 16

### Breaking changes
None

### DIFF
--- a/components/datepicker/DatePickerInput.js
+++ b/components/datepicker/DatePickerInput.js
@@ -1,0 +1,172 @@
+import React, { PureComponent } from 'react';
+import PropTypes from 'prop-types';
+import Box from '../box';
+import DayPickerInput from 'react-day-picker/DayPickerInput';
+import Icon from '../icon';
+import InputMetaPropTypes from '../input/InputMetaPropTypes';
+import NavigationBar from './NavigationBar';
+import WeekDay from './WeekDay';
+import { convertModifiersToClassnames } from './utils';
+import { IconCalendarSmallOutline, IconWarningBadgedSmallFilled } from '@teamleader/ui-icons';
+import { TextSmall } from '../typography';
+import cx from 'classnames';
+import omit from 'lodash.omit';
+import theme from './theme.css';
+
+class DatePickerInput extends PureComponent {
+  state = {
+    inputHasFocus: false,
+    selectedDate: null,
+  };
+
+  static getDerivedStateFromProps(props, state) {
+    if (props.selectedDate !== undefined && props.selectedDate !== state.selectedDate) {
+      return {
+        selectedDate: props.selectedDate,
+      };
+    }
+
+    return null;
+  }
+
+  handleBlur = () => {
+    this.setState({ inputHasFocus: false });
+  };
+
+  handleFocus = () => {
+    this.setState({ inputHasFocus: true });
+  };
+
+  handleDayClick = day => {
+    this.setState(
+      {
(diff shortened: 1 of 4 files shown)


components/datepicker/DatePickerInput.js
```diff
@@ -0,0 +1,173 @@
...
+    return null;
+  }
+
+  handleBlur = () => {
+    this.setState({ inputHasFocus: false });
+  };
+
+  handleFocus = () => {
+    this.setState({ inputHasFocus: true });
+  };
```
InstaK: You can just merge this in one handler again:

```js
handleFocus = () => {
  this.setState(({ inputHasFocus }) => { inputHasFocus: !inputHasFocus });
}
```

should be something like this. This handler can be a switch rather than two different handlers.

components/datepicker/DatePickerInput.js
```diff
@@ -0,0 +1,173 @@
...
+
+    return (
+      <DayPickerInput
+        classNames={theme}
+        dayPickerProps={{
+          className: dayPickerClassNames,
+          classNames: theme,
+          modifiers: convertModifiersToClassnames(modifiers, theme),
+          onDayClick: this.handleInputDateChange,
+          selectedDays: selectedDate,
```
InstaK: PP: Little different naming, `selectedDays` sounds like multiple days, `selectedDate` sounds like 1 day

components/datepicker/DatePickerInput.js
```diff
@@ -0,0 +1,173 @@
...
+  }
+}
+
+DatePickerInput.propTypes = {
+  bold: PropTypes.bool,
+  className: PropTypes.string,
+  dayPickerProps: PropTypes.object,
+  disabled: PropTypes.bool,
+  helpText: PropTypes.string,
+  inverse: PropTypes.bool,
```
InstaK: PP: Maybe rename this to isInverted?

InstaK: or isInverse like the css classname

components/datepicker/DatePickerInput.js
driesd: `selectedDays` is the prop I have to pass to the `DayPickerInput`, which is a [third-party component](http://react-day-picker.js.org/api/DayPickerInput). So I cannot change this :)

components/datepicker/DatePickerInput.js
driesd: Currently we don't use prefixes `is` or `has` for prop names. For now it's consistent to the other components.

 :ok: by @driesd on 06/06/2018 10:39:17

lowiebenoot: We use `inverse` throughout the whole UI, so I would leave it as it is.
we also don't use the `is`-naming for these kind props. Otherwise it would also need to be `isDisabled`, `isReadOnly`, ... I can't really explain why, but it feels better to me to have them without the `is`.

nickwaelkens: Because we're used of that from HTML 👍 

 :ok: by @nickwaelkens on 06/06/2018 11:31:49

components/datepicker/DatePickerInput.js
InstaK: You can say that for every switch then 😉

PR #259, opened 2018-06-06
DatePickerInputRange component

driesd:
### Description
A `DatePickerInputRange` component as wrapper for `react-day-picker`.

#### Screenshot after this PR
schermafdruk 2018-06-06 09 22 06

### Breaking changes
None

### DIFF
--- a/components/datepicker/DatePickerInputRange.js
+++ b/components/datepicker/DatePickerInputRange.js
@@ -108,6 +108,8 @@ class DatePickerInputRange extends PureComponent {
       ...others
     } = this.props;
 
+    const dayPickerInputProps = omit(others, ['helpText', 'meta', 'onBlur', 'onChange', 'onFocus']);
+
     const { selectedStartDate, selectedEndDate, mouseEnteredEndDate } = this.state;
     const modifiers = { from: selectedStartDate, to: mouseEnteredEndDate };
     const selectedDays = [selectedStartDate, { from: selectedStartDate, to: mouseEnteredEndDate }];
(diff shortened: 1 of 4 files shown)


components/datepicker/DatePickerInputRange.js
```diff
@@ -0,0 +1,298 @@
...
+  componentWillUnmount() {
+    clearTimeout(this.timeout);
+  }
+
+  focusFrom = () => {
+    this.timeout = setTimeout(() => this.startDate.getInput().focus(), 100);
+  };
+
+  focusTo = () => {
+    this.timeout = setTimeout(() => this.endDate.getInput().focus(), 0);
```
timdegroote: Why do we need the `setTimeout` here before focussing the input fields?

driesd: This is because of the third-party `react-day-picker` component that we are wrapping.
Check the comment above the `focusTo` method in their `Range with two inputs` example: http://react-day-picker.js.org/examples/input-from-to
This is due to some cross-browser issues they're explaining above the `handleInputBlur` and `handleOverlayBlur` methods: https://github.com/gpbl/react-day-picker/blob/master/src/DayPickerInput.js

lowiebenoot:
Said it in the `DatepickerInput` as well, but it has some weird border/shadow thingy going on at the bottom.

screen shot 2018-06-06 at 11 18 41

nickwaelkens:
If the issue with the border/shadow is recurring, can we look into it? 🤔

PR #261, opened 2018-06-06
Lint storybook stories

driesd:
### Description
I modified `package.json` so our Storybook stories now also get linted. That's why it is a big PR. But, all the changes were automatically made by Prettier™.

### Breaking changes
None


nickwaelkens:
Maybe now is also the perfect time to get it working with a precommit hook etc, check out invoicing/calendar repos!

driesd:
@nickwaelkens in package.json we have the following line:
"precommit": "npm run lint:fix"

It actually blocks comitting when there are linting errors.

duivvv:
@driesd 

I think we're talking about removing Prettier from the ESLint config & moving this to a precommit hook with.

https://prettier.io/docs/en/precommit.html#option-2-pretty-quick-https-githubcom-azz-pretty-quick

duivvv:
@kemosabert 

if Prettier doesn’t dictate we shouldn’t, that’s the point of Prettier right.

No need to add custom code formatting rules, that would be against the point of using Prettier

Eslint = making sure you catch code mistakes before running your code
Prettier = code formatting without discussion and personal preferences

Kemosabert:
@duivvv that makes complete sense, let's follow the prettier then

PR #265, opened 2018-06-08
Remove proptypes tables from other components from the proptypes table

Kemosabert:
### Description
In order to keep the proptypes table focused on the proptypes beloning to the selected component, we removed the proptypes from other imported components used in an example from the list.

### Breaking changes
none


stories/checkbox.js
```diff
@@ -4,12 +4,17 @@ import { storiesOf } from '@storybook/react';
 import { checkA11y } from 'storybook-addon-a11y';
 import { withInfo } from '@storybook/addon-info';
 import { withKnobs, boolean, select, text } from '@storybook/addon-knobs/react';
-import { Checkbox, Link, TextBody } from '../components';
+import { Checkbox, Link, TextBody, TextDisplay, TextSmall } from '../components';
+import { Text } from 'glamorous';
+import { textFactory } from '../components/typography/Text';
```
InstaK: Q: Why exactly are you importing these last 2?

stories/datagrid.js
```diff
@@ -23,96 +35,115 @@ const handleButtonClick = () => {
...
-                  <MenuItem onClick={() => console.log('onClick: delete row')}>Remove row</MenuItem>
-                </IconMenu>
-              </DataGrid.Cell>
-            </DataGrid.BodyRow>
-          );
-        })
-      }
+      {rows1.map((row, index) => {
+        return (
+          <DataGrid.BodyRow key={index}>
```
InstaK: Please refactor this index as key to something else, `key={row}` should suffice

stories/checkbox.js
Kemosabert: I think this must have been added automatically by a plugin I used in my editor. Do I just remove this in a new commit?

Kemosabert: @InstaK I removed the unnecessary imports from the checkbox story

PR #266, opened 2018-06-11
Update dependencies to latest version

Kemosabert:
### Description
Update all the project's dependencies to the latest version. 

### Breaking changes
none


duivvv:
Why are we all updating them explicitly?
https://docs.npmjs.com/misc/semver#caret-ranges-123-025-004

We only need to update the dependencies if the major changed or < 1.0.0

package.json
```diff
@@ -63,79 +63,79 @@
...
-    "postcss-loader": "^2.0.7",
-    "postcss-mixins": "^6.1.0",
+    "postcss-import": "^11.1.0",
+    "postcss-loader": "^2.1.5",
+    "postcss-mixins": "^6.2.0",
     "postcss-nested": "^3.0.0",
     "postcss-pseudoelements": "^5.0.0",
     "postcss-reporter": "^5.0.0",
-    "prettier": "^1.4.4",
+    "prettier": "^1.13.5",
```
nickwaelkens: This will result in a follow-up "fix linting" PR 💃

nickwaelkens:
Agreed with @duivvv - feels kinda useless, semver tackles this for us.

InstaK:
I guess this PR can be reduced to only the hardcoding of prettier? So without the caret then?

duivvv:
@InstaK 

+ having a look at what dependencies moved to a major release (and testing them)

Kemosabert:
Addressed this issue in a new PR: #292

PR #269, opened 2018-06-11
Box: move static propTypes & defaultProps outside of the component class

driesd:
### Description
Move static propTypes & defaultProps outside of the component class

### Breaking changes
None


driesd:
👍

PR #276, opened 2018-06-11
Eliminate constructors

driesd:
### Description
Eliminate all component constructors.

### Breaking changes
None

### DIFF
--- a/components/banner/Banner.js
+++ b/components/banner/Banner.js
@@ -7,15 +7,6 @@ import theme from './theme.css';
 import { IconCloseMediumOutline } from '@teamleader/ui-icons';
 
 class Banner extends PureComponent {
-  constructor() {
-    super(...arguments);
-    this.handleClick = ::this.handleClick;
-  }
-
-  handleClick() {
-    this.props.onClose();
-  }
-
   getCloseButtonColor() {
     const { color } = this.props;
     return color === 'white' ? 'neutral' : color;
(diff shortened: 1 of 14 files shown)


components/banner/Banner.js
```diff
@@ -35,7 +26,7 @@ class Banner extends PureComponent {
               className={theme['close-button']}
               icon={<IconCloseMediumOutline />}
               color={this.getCloseButtonColor()}
-              onClick={this.handleClick}
+              onClick={onClose}
```
duivvv: shoudn't this one be in the destructuring?

driesd: I think this is already the case?

const { children, className, icon, `onClose`, fullWidth, ...others } = this.props;

InstaK: @duivvv it is, you probably looked at the props of getCloseButtonColor? 😄

duivvv: Ah, the diff got me confused :)

components/tooltip/Tooltip.js
```diff
@@ -30,17 +30,12 @@ const SIZES = {
...
-      super(...arguments);
+    tooltipRoot = this.props.documentObject.createElement('div');
 
-      this.state = {
-        active: false,
-        position: this.props.tooltipPosition,
-      };
-
-      this.tooltipRoot = props.documentObject.createElement('div');
-      this.tooltipRoot.id = 'tooltip-root';
```
lowiebenoot: This one wasn't used and could just be dropped I guess?

PR #278, opened 2018-06-12
Enhance Badge component styles

Kemosabert:
### Description
Add color themes to the Badge component for the Event Details project and update the different states so it aligns with the styleguide. 

### Screenshot
schermafdruk 2018-06-12 14 31 36

### Breaking changes
none

### DIFF
--- a/components/badge/Badge.js
+++ b/components/badge/Badge.js
@@ -21,10 +21,22 @@ class Badge extends PureComponent {
   };
 
   render() {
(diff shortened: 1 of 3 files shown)


Kemosabert:
@InstaK that is something we will definitely start doing in the near future

duivvv:
Not too sure if we should keep using those exotic PostCSS plugins TBH.

I would prefer to have a look how to make it dynamic via inline style + color prop.

duivvv:
@nickwaelkens I agree, my reaction was a response on @InstaK's suggestion

PR #280, opened 2018-06-13
Select component

driesd:
### Description
A `Select` component as a wrapper for `React select v2` (see [docs](https://react-select.com/home))

#### Screenshot
schermafdruk 2018-06-13 10 48 48

### Breaking changes
None


components/select/Select.js
```diff
@@ -0,0 +1,77 @@
...
+    option: (base, { isFocused, isSelected }) => ({
+      ...base,
+      color: '#344b63',
+      backgroundColor: isFocused ? '#e4e4e6' : isSelected ? '#f7f7fa' : '#fff',
+      '&:active': {
+        backgroundColor: '#e4e4e6'
+      },
+    }),
+  });
+
```
nickwaelkens: It looks like we can still use `classNames` (https://deploy-preview-2289--react-select.netlify.com/upgrade-guide#new-styles-api) - not sure what our approach should be here though 😢

duivvv:
We should also add it to `CHANGELOG.md`

components/select/index.js
```diff
@@ -0,0 +1,4 @@
+import Select from './Select';
+
+export default Select;
+export { Select };
```
lowiebenoot: You can remove this named export, we stopped doing these.

components/select/Select.js
```diff
@@ -0,0 +1,291 @@
...
+    }
+
+    return {
+      ...commonStyles,
+      backgroundColor: isDisabled ? '#e4e4e6' : '#c0c0c4',
+    }
+  };
+
+  getInput = base => {
+    const { size, value } = this.props;
```
lowiebenoot: `value` is missing in the propTypes.

components/select/Select.js
```diff
@@ -0,0 +1,291 @@
...
+
+class Select extends PureComponent {
+  getClearIndicatorStyles = base => {
+    const { inverse } = this.props;
+
+    return {
+      ...base,
+      color: inverse ? '#e1ecfa': '#344b63',
+      '&:hover': {
+        color: inverse ? '#fff' : '#2a3b4d',
```
lowiebenoot: Maybe we should put all the colors (that we use here) in a constant.
```js
const colors = {
  white: '#fff',
  neutral: '...',
  neutralDarkest: ' ...',
...
};
```

PR #284, opened 2018-06-15
Update/counter

Kemosabert:
### Description
Add the possibility to add a small border to the counter for when it is overlapping an avatar

screen shot 2018-06-15 at 10 59 24

#### Screenshot after this PR
screen shot 2018-06-15 at 10 59 08

### Breaking changes
none

### DIFF
--- a/components/counter/Counter.js
+++ b/components/counter/Counter.js
@@ -8,7 +8,18 @@ class Counter extends PureComponent {
   render() {
     const { children, className, color, count, maxCount, size, borderColor, borderTint, ...others } = this.props;
 
-    const classNames = cx(theme['counter'], theme[color], theme[size], className);
+    const classNames = cx(
+      theme['counter'],
+      theme[color],
+      theme[size],
+      {
+        [theme[`border-${borderColor}-${borderTint}`]]: borderTint,
+        [theme[`border-${borderColor}`]]: !borderTint,
+      },
+      className,
+    );
+
+    console.log(classNames);
 
     return (
       <Box className={classNames} element="span" {...others} data-teamleader-ui="counter">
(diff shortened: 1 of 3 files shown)


components/counter/Counter.js
```diff
@@ -6,9 +6,18 @@ import theme from './theme.css';
...
+    const { children, className, color, count, maxCount, size, borderColor, borderTint, ...others } = this.props;
 
-    const classNames = cx(theme['counter'], theme[color], theme[size], className);
+    const classNames = cx(
+      theme['counter'],
+      theme[color],
+      theme[size],
+      {
+        [theme[`border-${borderColor}-${borderTint}`]]: borderTint,
+        [theme[`border-${borderColor}`]]: !borderTint,
```
InstaK: Q: I don't really get this? So if there's a borderTint the classname will be `border-mint-dark` for example, and if there is no borderTint the classname will always be `border-borderColor`? Then you will need a defaultProps for borderColor right? And do we *always* want a borderColor? Or am I just grasping this incorrectly?

InstaK:
I don't think this is the correct solution either? Now you will have a borderColor if borderColor is set and borderTint is not set, but you can have both right? I think the check should just be `borderColor`?

components/counter/Counter.js
```diff
@@ -14,7 +14,7 @@ class Counter extends PureComponent {
       theme[size],
       {
         [theme[`border-${borderColor}-${borderTint}`]]: borderColor && borderTint,
-        [theme[`border-${borderColor}`]]: !borderTint,
+        [theme[`border-${borderColor}`]]: borderColor && !borderTint,
```
InstaK: I don't think this is the correct solution either? Now you will have a borderColor if borderColor is set and borderTint is not set, but you can have both right? I think the check should just be `borderColor`?

InstaK: Ah no, then you have the combination of borderColor-borderTint. Nevermind.

 :ok: by @InstaK on 15/06/2018 11:02:13

PR #286, opened 2018-06-15
Component/progress tracker

Kemosabert:
### Description
Progress tracker component

#### Screenshot after this PR
screen shot 2018-06-18 at 11 48 20

### Breaking changes
none

### DIFF
--- a/components/progressTracker/ProgressStep.js
+++ b/components/progressTracker/ProgressStep.js
@@ -24,7 +24,7 @@ class ProgressStep extends PureComponent {
 
 ProgressStep.propTypes = {
   /** Color theme of the progress tracker. */
-  color: PropTypes.oneOf(['neutral', 'mint', 'aqua', 'violet', 'teal', 'gold', 'ruby']),
+  color: PropTypes.oneOf(['neutral', 'mint', 'aqua', 'violet', 'gold', 'ruby']),
   /** The label for the progress step */
   label: PropTypes.string.isRequired,
   /** Whether or not the step is active */
--- a/components/progressTracker/ProgressStep.js
+++ b/components/progressTracker/ProgressStep.js
@@ -16,7 +16,7 @@ class ProgressStep extends PureComponent {
     return (
       <Box className={classNames}>
         <TextSmall className={theme['progress-step__label']}>{label}</TextSmall>
-        <Box className={theme['status-bullet']} />
+        <span className={theme['status-bullet']} />
       </Box>
     );
   }
--- a/components/progressTracker/ProgressStep.js
+++ b/components/progressTracker/ProgressStep.js
@@ -3,6 +3,8 @@ import PropTypes from 'prop-types';
 import cx from 'classnames';
 import theme from './theme.css';
 
+import Box from '../box';
+
 class ProgressStep extends PureComponent {
   render() {
     const { label, active, completed, color } = this.props;
@@ -13,6 +15,7 @@ class ProgressStep extends PureComponent {
     return (
       <div className={classNames}>
         <label>{label}</label>
+        <Box className={theme['status-bullet']} />
       </div>
     );
   }
--- a/components/progressTracker/ProgressStep.js
+++ b/components/progressTracker/ProgressStep.js
@@ -8,15 +8,17 @@ import { TextSmall } from '../typography';
 
 class ProgressStep extends PureComponent {
   render() {
-    const { label, active, completed, color } = this.props;
+    const { label, active, completed, color, onClickHandler, stepId } = this.props;
     const classNames = cx(theme['progress-step'], theme[color], {
       [theme['is-active']]: active,
       [theme['is-completed']]: completed,
     });
     return (
       <div className={classNames}>
-        <TextSmall className={theme['progress-step__label']}>{label}</TextSmall>
-        <Box className={theme['status-bullet']} />
+        <TextSmall className={theme['progress-step__label']} onClick={() => onClickHandler(stepId)}>
+          {label}
+        </TextSmall>
+        <Box className={theme['status-bullet']} onClick={() => onClickHandler(stepId)} />
       </div>
     );
   }
@@ -31,6 +33,10 @@ ProgressStep.propTypes = {
   active: PropTypes.bool.isRequired,
   /** Whether or not the step has been completed */
   completed: PropTypes.bool.isRequired,
+  /** Function to be called when a step has been clicked */
+  onClickHandler: PropTypes.func,
+  /** Id of the step */
+  stepId: PropTypes.number.isRequired,
 };
 
 ProgressStep.defaultProps = {
--- a/components/progressTracker/ProgressStep.js
+++ b/components/progressTracker/ProgressStep.js
@@ -14,10 +14,10 @@ class ProgressStep extends PureComponent {
       [theme['is-completed']]: completed,
     });
     return (
-      <div className={classNames}>
+      <Box className={classNames}>
         <TextSmall className={theme['progress-step__label']}>{label}</TextSmall>
         <Box className={theme['status-bullet']} />
-      </div>
+      </Box>
     );
   }
 }
--- a/components/progressTracker/ProgressStep.js
+++ b/components/progressTracker/ProgressStep.js
@@ -9,13 +9,13 @@ import { TextSmall } from '../typography';
 class ProgressStep extends PureComponent {
   render() {
     const { label, active, completed, color } = this.props;
-    const classNames = cx(theme['progress-step'], theme[color], {
+    const classNames = cx(theme['step'], theme[color], {
       [theme['is-active']]: active,
       [theme['is-completed']]: completed,
     });
     return (
       <Box className={classNames}>
-        <TextSmall className={theme['progress-step__label']}>{label}</TextSmall>
+        <TextSmall className={theme['step-label']}>{label}</TextSmall>
         <span className={theme['status-bullet']} />
       </Box>
     );
--- a/components/progressTracker/ProgressStep.js
+++ b/components/progressTracker/ProgressStep.js
@@ -1,0 +1,9 @@
+import React, { PureComponent } from 'react';
+
+class ProgressStep extends PureComponent {
+  render() {
+    return null;
+  }
+}
+
+export default ProgressStep;
--- a/components/progressTracker/ProgressStep.js
+++ b/components/progressTracker/ProgressStep.js
@@ -1,9 +1,22 @@
 import React, { PureComponent } from 'react';
+import PropTypes from 'prop-types';
 
 class ProgressStep extends PureComponent {
   render() {
-    return null;
+    const { label, active } = this.props;
+    return (
+      <div>
+        <label>{label}</label>
+      </div>
+    );
   }
 }
+
+ProgressStep.propTypes = {
+  /** The label for the progress step */
+  label: PropTypes.string.isRequired,
+  /** Whether or not the step is active */
+  active: PropTypes.bool.isRequired,
+};
 
 export default ProgressStep;
--- a/components/progressTracker/ProgressStep.js
+++ b/components/progressTracker/ProgressStep.js
@@ -35,6 +35,8 @@ ProgressStep.propTypes = {
 
 ProgressStep.defaultProps = {
   color: 'neutral',
+  active: false,
+  completed: false,
 };
 
 export default ProgressStep;
--- a/components/progressTracker/ProgressStep.js
+++ b/components/progressTracker/ProgressStep.js
@@ -1,11 +1,17 @@
 import React, { PureComponent } from 'react';
 import PropTypes from 'prop-types';
+import cx from 'classnames';
+import theme from './theme.css';
 
 class ProgressStep extends PureComponent {
   render() {
-    const { label, active } = this.props;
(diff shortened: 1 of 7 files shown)


components/progressTracker/ProgressStep.js
```diff
@@ -0,0 +1,42 @@
+import React, { PureComponent } from 'react';
+import PropTypes from 'prop-types';
+import cx from 'classnames';
+import theme from './theme.css';
+
+import Box from '../box';
+import { TextSmall } from '../typography';
+
+class ProgressStep extends PureComponent {
```
InstaK: I guess in order to be conventionally correct this should be in another folder? `Modules/ProgressStep.js`?

stories/progressTracker.js
```diff
@@ -0,0 +1,36 @@
...
+      propTablesExclude: [Island],
+    })(story)(context),
+  )
+  .addDecorator(withKnobs)
+  .addDecorator(checkA11y)
+  .add('Basic', () => (
+    <Island color={select('Color', colors, 'neutral')} size="small">
+      <ProgressTracker color={select('Color', colors, 'neutral')} activeStep={number('Active step', 1, options)}>
+        {steps.map((step, index) => {
+          return <ProgressTracker.ProgressStep label={step} key={index} />;
```
InstaK: You can remove the brackets:
```js
{steps.map((step, index) => <ProgressTracker.ProgressStep label={step} key={index} />}
```

components/progressTracker/ProgressTracker.js
```diff
@@ -0,0 +1,47 @@
...
+class ProgressTracker extends PureComponent {
+  render() {
+    const { color, children, activeStep } = this.props;
+
+    const classNames = cx(theme['progress-tracker']);
+
+    return (
+      <Box className={classNames}>
+        {React.Children.map(children, (child, index) => {
+          const currentActiveStep = activeStep < 0 ? 0 : activeStep;
```
InstaK: Q: How exactly can currentActiveStep ever be `< 0`? The default value is 0 so that shouldn't be a problem right?

components/progressTracker/theme.css
```diff
@@ -0,0 +1,167 @@
+@import '@teamleader/ui-colors';
+@import '@teamleader/ui-typography';
+@import '@teamleader/ui-animations';
```
InstaK: Please prefix with a tilde `@import '~@teamleader/ui-animations'`,...

components/progressTracker/ProgressTracker.js
```diff
@@ -0,0 +1,47 @@
...
+
+          return React.cloneElement(child, {
+            active: index === currentActiveStep,
+            completed: index < currentActiveStep,
+            color: color,
```
InstaK: `color` in ES6 is the same as `color: color`, so you can just get rid of `: color`

components/progressTracker/ProgressTracker.js
```diff
@@ -0,0 +1,47 @@
...
+    const classNames = cx(theme['progress-tracker']);
+
+    return (
+      <Box className={classNames}>
+        {React.Children.map(children, (child, index) => {
+          const currentActiveStep = activeStep < 0 ? 0 : activeStep;
+
+          return React.cloneElement(child, {
+            active: index === currentActiveStep,
+            completed: index < currentActiveStep,
```
InstaK: Also thinking about this completed: What if we want to have a 4-step procedure and step 5 means everything's finished, then we have shouldn't put step 5 as active but immediately finished, this forces us to set the step to 6 in order to have it as `completed`? Should we provide an option to force completed?

components/progressTracker/ProgressStep.js
Kemosabert: @InstaK I was following the same structure as we have for the Datagrid component where child components are placed in the same directory as the main component.

InstaK: Fair enough, not sure if we should change this (because everywhere else I've seen it it was like I just mentioned or just keep going this way 🤔 )

components/progressTracker/theme.css
Kemosabert: If I do this it throws an error from the css-loader saying that those CSS modules cannot be found. I also did not see this being used in any other component as I think they are already being imported as modules?

InstaK: Oh you're right, there's a different postcss config in tl-ui, woops

 :ok: by @InstaK on 18/06/2018 13:51:46

components/progressTracker/ProgressTracker.js
InstaK: Okay I guess it would be a safety, although it would also be a sign for the props being passed that the activeStep is `< 0` and therefor incorrect. Now if you would pass something that's smaller than 0 it will just put the step to 0 and it won't look as if something's wrong. Which is in a way a less destructive way if anything were to go wrong... I can live with the check

 :ok: by @InstaK on 18/06/2018 13:54:36

components/progressTracker/ProgressTracker.js
Kemosabert: I could add a prop called "allStepsAreCompleted" and then check for that too?

InstaK: I think that wouldn't be a luxury :)

components/progressTracker/ProgressTracker.js
nickwaelkens:
```js
// (if you rename activeStep to current)
const currentStep = Math.max(0, current);
```

components/progressTracker/theme.css
```diff
@@ -0,0 +1,167 @@
...
+
+.progress-tracker {
```
nickwaelkens: No need to prefix your classnames as we are using modules anyway.

static/progressTracker/steps.js
```diff
@@ -0,0 +1,3 @@
+const steps = ['Draft', 'Book', 'Send invoices', 'Get paid'];
+
+export { steps };
```
nickwaelkens: Doesn't really deserve its own file imo, and isn't a static file 🤔 Just put it in the story itself.

nickwaelkens:
I have also tested it locally and works great. Only thing I'm not 100% satisfied about is the animation. The colored bar is still moving and the text is already bold/other color and already has the icon. Can we delay that in some way so it feels more natural? 

<sup>Come talk to me if it's not clear!</sup>

stories/progressTracker.js
```diff
@@ -27,7 +27,7 @@ storiesOf('ProgressTracker', module)
   .addDecorator(checkA11y)
   .add('Basic', () => (
     <Island color={select('Color', colors, 'neutral')} size="small">
-      <ProgressTracker color={select('Color', colors, 'neutral')} activeStep={number('Active step', 1, options)}>
+      <ProgressTracker color={select('Color', colors, 'neutral')} activeStep={number('Active step', 1, options)} allStepsCompleted={boolean('All steps are completed', false)}>
```
nickwaelkens: Keep it simple; `completed`?

components/progressTracker/ProgressStep.js
lowiebenoot: Not sure what you're saying. `Modules` isn't even a directory at the moment?

components/progressTracker/ProgressStep.js
```diff
@@ -0,0 +1,42 @@
...
+  render() {
+    const { label, active, completed, color } = this.props;
+    const classNames = cx(theme['progress-step'], theme[color], {
+      [theme['is-active']]: active,
+      [theme['is-completed']]: completed,
+    });
+    return (
+      <Box className={classNames}>
+        <TextSmall className={theme['progress-step__label']}>{label}</TextSmall>
+        <Box className={theme['status-bullet']} />
```
lowiebenoot: Not really useful to make this a box. Just use a `span`.

stories/progressTracker.js
nickwaelkens: Just needs linting 👍 

 :ok: by @nickwaelkens on 18/06/2018 15:51:45

components/progressTracker/ProgressStep.js
Kemosabert: @lowiebenoot if we are talking about specific child components for parent components, they are not placed in a separate folder.

components/progressTracker/ProgressStep.js
Kemosabert: we also use a Box for the Bullet component, so I reused the same here methodology as we applied there to remain consistent.

lowiebenoot: But you can't pass props to this Box component, so it's useless.

components/progressTracker/theme.css
```diff
@@ -64,9 +64,39 @@
...
+    border-radius: 50%;
+    display: block;
+    height: 12px;
+    margin: 6px 0;
+    width: 12px;
+    z-index: 2;
+    position: relative;
+    &:before {
+      background-color: transparent;
+      transition: background-color 6s var(--animation-curve-default) calc(var(--animation-duration) - .1s);
```
lowiebenoot: 6 seconds? 🤔

components/progressTracker/theme.css
```diff
@@ -64,9 +64,39 @@
...
+      position: absolute;
+      width: 6px;
+      height: 6px;
+      top: 50%;
+      left: 50%;
+      transform: translate(-50%, -50%);
+    }
+  }
+
+  .progress-step__label {
```
lowiebenoot: We don't use BEM.

components/progressTracker/theme.css
```diff
@@ -64,9 +64,39 @@
...
+    font-family: var(--font-family-regular);
+    opacity: .5;
+    text-align: center;
+    transition: font-family .1s var(--animation-curve-default) calc(var(--animation-duration) - .1s), opacity .1s var(--animation-curve-default) calc(var(--animation-duration) - .1s);
+    &:hover {
+      cursor: normal;
```
lowiebenoot: What's the reason you're adding this? Because it's "normal" by default.

components/progressTracker/theme.css
Kemosabert: forgot to switch this one out.

 :ok: by @Kemosabert on 19/06/2018 09:53:33

lowiebenoot:
@Kemosabert please don't make commit messages like "fix css". It's not hard to come up with something better and more explaining.

components/progressTracker/theme.css
Kemosabert: Changed the classname

 :ok: by @Kemosabert on 19/06/2018 09:55:06

components/progressTracker/theme.css
Kemosabert: In the beginning I had to make the bullets themselves clickable and interactive, which is why I was adjusting the hover state, but forgot to erase it after putting it back to normal.

 :ok: by @Kemosabert on 19/06/2018 09:56:02

components/progressTracker/theme.css
Kemosabert: Changed in the latest commit

 :ok: by @Kemosabert on 19/06/2018 11:40:42

components/progressTracker/theme.css
```diff
@@ -0,0 +1,180 @@
...
+    position: relative;
+    &:before {
+      background-color: transparent;
+      transition: background-color .1s var(--animation-curve-default) calc(var(--animation-duration) - .1s);
+      border-radius: 50%;
+      content: '';
+      display: block;
+      position: absolute;
+      width: 6px;
+      height: 6px;
```
nickwaelkens: Nitpicking, but it feels like 12px/6px stuff are just wholes/halves. Could become a variable 🙈

components/progressTracker/ProgressTracker.js
```diff
@@ -0,0 +1,52 @@
...
+    const classNames = cx(theme['tracker']);
+
+    return (
+      <Box className={classNames}>
+        {React.Children.map(children, (child, index) => {
+          const activeStep = Math.max(0, currentStep);
+
+          return (
+            <ProgressStep
+              { ...child.props}
```
nickwaelkens: I think you forgot to lint, or is this space allowed?

components/progressTracker/ProgressTracker.js
```diff
@@ -0,0 +1,52 @@
...
+import PropTypes from 'prop-types';
+import cx from 'classnames';
+import theme from './theme.css';
+
+import Box from '../box';
+import ProgressStep from './ProgressStep';
+
+class ProgressTracker extends PureComponent {
+  render() {
+    const { color, children, currentStep, allStepsCompleted } = this.props;
```
nickwaelkens: PP: `done` > `allStepsCompleted`

components/progressTracker/theme.css
duivvv: that's weird, the precommit hook should catch this

components/progressTracker/ProgressTracker.js
duivvv: that's weird, the precommit hook should catch this

Kemosabert: What an eagle eye, but yes. It should have been caught by the pre-commit hook, will check that out now.

 :ok: by @Kemosabert on 19/06/2018 17:40:34

Issue #229, opened 2018-03-22
Unexpected behaviour input


InstaK:
When using the Input component and adding a function to it, e.g.
```js
  <Input input={{onChange: this.handleChange}}/>
```
### Expected behavior
The parameter `event` in the `handleChange` function to be your event
### Actual behavior
The parameter `event` in the `handleChange` function is your event.target.value

Upon fixing this needs to be changed in `client/src/application/SupportPermissionRequest/SupportPermissionRequest.js`

Kemosabert:
Fixed in PR #289

PR #289, opened 2018-06-19
Send full event object down to the provided onchange handler 

Kemosabert:
### Description
Fix for issue #229 

The first parameter in the onChange handler you provide via the props no longer holds only the value, but the entire event object. I still pas down the value as a second argument, as with the increment buttons this value might not always be +1 or -1 as it depends on the defined steps.

### Breaking changes
The first parameter in the onChange handler is not just the value but the entire event object.

### DIFF
--- a/components/input/Input.js
+++ b/components/input/Input.js
@@ -59,20 +59,20 @@ export default class Input extends PureComponent {
   };
 
   handleChange = event => {
-    this.updateValue(event.target.value);
+    this.updateValue(event);
   };
 
-  handleIncreaseValue = () => {
-    this.updateStep(1);
+  handleIncreaseValue = event => {
+    this.updateStep(event, 1);
   };
 
-  handleDecreaseValue = () => {
-    this.updateStep(-1);
+  handleDecreaseValue = event => {
+    this.updateStep(event, -1);
   };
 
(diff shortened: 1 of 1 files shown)


components/input/Input.js
```diff
@@ -59,30 +59,30 @@ export default class Input extends PureComponent {
...
       value,
     });
 
     if (triggerOnChange) {
       if (input && input.onChange) {
-        input.onChange(value);
+        input.onChange(event, value);
       } else if (onChange) {
-        onChange(value);
+        onChange(event, value);
```
InstaK: Q: Not really seeing why both parameters need to be added passed here, can you explain this please?

Kemosabert: If you have an input field for numbers, you can in- and decrement the number by a given step by ao clicking on a button. But as that increment value is not saved on the button element itself, it will not appear in the event object and event.target.value will be null or ' ' for that event.

InstaK: Fair enough, thanks!

 :ok: by @InstaK on 19/06/2018 12:48:21

nickwaelkens:
Thanks for this. As this is a breaking change, please don't forget to update the description of the next draft version https://github.com/teamleadercrm/ui/releases
(or did no one already tell you you should do that? Sorry!)

Issue #230, opened 2018-03-22
Dialog under new navigation


InstaK:
### Expected behavior
Dialog to be over new navigation

### Actual behavior
Dialog is under new navigation (z-index)

PR #290, opened 2018-06-19
Adjust Dialog z-index to make sure it appears over the new navigation

Kemosabert:
### Description
Fix for issue #230

Adjusted the z-index of the Dialog to what the current z-index of the dialogs in production are

### Breaking changes
none

### DIFF
--- a/components/dialog/theme.css
+++ b/components/dialog/theme.css
@@ -16,7 +16,7 @@
   position: fixed;
   top: 0;
   width: 100vw;
-  z-index: var(--z-index-higher);
+  z-index: 6000;
 }
 
 .dialog {


components/dialog/theme.css
```diff
@@ -16,7 +16,7 @@
   position: fixed;
   top: 0;
   width: 100vw;
-  z-index: var(--z-index-higher);
+  z-index: 6000;
```
lowiebenoot: We really need to find a way to handle z-indexes better...

lowiebenoot: I don't know, but we should at least think about it 😄 /cc @nickwaelkens

nickwaelkens: Sigh. Very true. 😢 
https://teamleader.atlassian.net/browse/GFE-6

Kemosabert: Maybe we can work with more specific names for the different tiers of z-index and specify their use more. Explained here: https://css-tricks.com/handling-z-index/

nickwaelkens: Thing is, how are we going to scale this over multiple apps? I think there needs to be some sort of communication across said apps... Food for thought 👍

PR #292, opened 2018-06-19
Update/update packages

Kemosabert:
### Description
Update major version changes of packages and updates in packages below 1.0.0.

Re-do of #266 

### Breaking changes
none


duivvv:
I would also lock Prettier to make sure there are no styling changes between installs.

Issue #277, opened 2018-06-11
Datagrid doesn't call onSelectionChange when the comparableId changed


lowiebenoot:
### Expected behavior
When the `comparableId` changes, the selection resets, but the `onSelectionChange` is not called.

### Actual behavior
It should call the `onSelectionChange`.

Kemosabert:
Fixed in PR #293

PR #293, opened 2018-06-19
Fix/call on selection change when comparable id changes in datagrid

Kemosabert:
### Description
Fix for issue #277 and check if the onSelectionChange prop has actually been provided before calling it.

### Breaking changes
none


components/datagrid/DataGrid.js
```diff
@@ -39,6 +39,8 @@ class DataGrid extends PureComponent {
 
   componentWillReceiveProps(nextProps) {
     if (nextProps.comparableId !== this.props.comparableId) {
+      this.handleSelectionChange();
```
lowiebenoot: You probably should have sent empty array as param.

Issue #223, opened 2018-03-08
Segmented islands


InstaK:
It would be nice to have segmented islands, this might be coming back more often in the designs. https://puu.sh/zDaTh/b08cf333c4.png

Could be easily achieved, css is already being used in `Integrations.js`

```cs
.integration-container {
  &:not(:first-child), &:not(:last-child) {
    border-radius: 0;
  }

  &:not(:first-child) {
    border-right: 0;
  }

  &:not(:last-child) {
    border-right: 0;
  }

  &:first-child {
    border-radius: 4px 0 0 4px;
  }
  &:last-child {
    border-radius: 0 4px 4px 0;
  }
}
```

Kemosabert:
Fixed in PR #294

PR #294, opened 2018-06-20
Provide an easy way to create segmented islands

Kemosabert:
### Description
For enhancement request #223 

#### Screenshot after this PR
screen shot 2018-06-20 at 10 57 06

### Breaking changes
none

### DIFF
--- a/components/island/IslandGroup.js
+++ b/components/island/IslandGroup.js
@@ -3,6 +3,7 @@ import PropTypes from 'prop-types';
 import cx from 'classnames';
 import Box from '../box';
 import theme from './theme.css';
+import Island from './Island';
 
 class IslandGroup extends PureComponent {
   isDark(color) {
@@ -13,7 +14,7 @@ class IslandGroup extends PureComponent {
     return this.props.dark;
   }
   render() {
-    const { children, className, color } = this.props;
+    const { children, className, color, size } = this.props;
 
     const isDark = this.isDark(color);
 
@@ -21,7 +22,13 @@ class IslandGroup extends PureComponent {
 
     return (
       <Box className={classNames} padding={0}>
-        {children}
+        {React.Children.map(children, child => {
+          return (
+            <Island {...child.props} color={color} dark={isDark} size={size}>
+              {child.props.children}
+            </Island>
+          );
+        })}
       </Box>
     );
   }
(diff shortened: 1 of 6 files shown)


components/island/theme.css
```diff
@@ -55,3 +55,34 @@
...
+  &:not(:first-child) {
+    border-right: 0;
+  }
+
+  &:not(:last-child) {
+    border-left: 0;
+  }
+
+  &:first-child {
+    border-radius: 4px 0 0 4px;
```
nickwaelkens: Can we move these instances of `4px` to a separate variable (and the one on L6 as well?) Thanks!

lowiebenoot:
I find it a bit weird that the "child" islands are segmented. I would expect that you have to put the `segment` prop on the parent component, which is than more of an `IslandGroup`. Just like we have `ButtonGroup`.

Kemosabert:
@lowiebenoot That is what I initially tried, but a class on the parent did not have an effect on the children with class island, I thought because of how styled components worked. I double checked however and now this does seem to work, so I will make a new commit for this.

nickwaelkens:
I feel like you're almost there, but what @lowiebenoot is trying to say is that we probably need a wrapper component called `IslandGroup` (or `Archipelago` :trollface:) just like we have `Button`s and `ButtonGroup`

lowiebenoot:
Yes, `IslandGroup` can automatically pass the segmented prop down to the child `Island` components.

components/island/IslandGroup.js
```diff
@@ -0,0 +1,41 @@
+import React, { PureComponent } from 'react';
+import PropTypes from 'prop-types';
+import cx from 'classnames';
+import Box from '../box';
+import theme from './theme.css';
+
```
nickwaelkens: Why pass `color`, but not `dark`? Pass both or don't pass any?

```jsx
const lightColors = ['white', 'neutral'];

isDark () => {
	const { dark, color } = this.props;
	if (lightColors.includes(color)) {
		return false;
	}

	return dark;
}
```

Also, I see this come back again and again. Care to pick this up in a follow-up PR and put this in a util/helper? Thanks!

lowiebenoot:
It would be nice if IslandGroup would pass down more props, just like we do in `ButtonGroup`.

For example, we could set the color prop on the `IslandGroup`.

Now you have to do it like this:
```js
<IslandGroup>
   <Island color="mint">...</Island>
   <Island color="mint">...</Island>
</Island>
```

While it could be like this:
```js
<IslandGroup color="mint">
   <Island>...</Island>
   <Island>...</Island>
</Island>
```

Kemosabert:
@lowiebenoot Completely right, I updated this in the latest commit.

components/island/IslandGroup.js
Kemosabert: Yes, I will make a follow-up commit for this.

 :ok: by @Kemosabert on 28/06/2018 10:46:13

PR #295, opened 2018-06-20
Add custom link and action option to Toasts

Kemosabert:
### Description
Add the option to add a custom action and a custom link to the toast to extend its flexibility. Also updated the styles so it aligns better with the styleguide.

### Breaking changes
The action prop no longer accepts a string, but now accepts a function. The textual label for this action can be provided via actionLabel.

### DIFF
--- a/components/toast/Toast.js
+++ b/components/toast/Toast.js
@@ -89,7 +89,7 @@ class Toast extends PureComponent {
               {action ? (
                 <LinkButton className={theme['action-link']} inverse label={actionLabel} onClick={action} />
               ) : link ? (
-                <Link href={link} target="_blank">
+                <Link href={link} target="_blank" className={theme['action-link']}>
                   {linkLabel}
                 </Link>
               ) : onClose ? (
--- a/components/toast/Toast.js
+++ b/components/toast/Toast.js
@@ -48,37 +48,27 @@ class Toast extends PureComponent {
     }, timeout);
   };
 
-  renderCustomAction = (action, actionLabel) => {
-    if (action) {
-      return <LinkButton className={theme['action-link']} inverse label={actionLabel} onClick={action} />;
-    }
-    return false;
-  };
-  renderCustomLink = (link, linkLabel, linkTarget) => {
-    if (link) {
-      return (
-        <TextBody className={theme['label']} color="white" soft>
-          <Link href={link} target={linkTarget} className={theme['action-link']} inherit>
-            {linkLabel}
-          </Link>
-        </TextBody>
-      );
-    }
-    return false;
-  };
-  renderCloseButton = onClose => {
-    if (onClose) {
-      return (
-        <IconButton
-          className={theme['action-button']}
-          icon={<IconCloseMediumOutline />}
-          color="white"
-          onClick={onClose}
-        />
-      );
-    }
-    return null;
-  };
+  renderCustomAction = (action, actionLabel) =>
+    action && <LinkButton className={theme['action-link']} inverse label={actionLabel} onClick={action} />;
+
+  renderCustomLink = (link, linkLabel, linkTarget) =>
+    link && (
+      <TextBody className={theme['label']} color="white" soft>
+        <Link href={link} target={linkTarget} className={theme['action-link']} inherit>
+          {linkLabel}
+        </Link>
+      </TextBody>
+    );
+
+  renderCloseButton = onClose =>
+    onClose && (
+      <IconButton
+        className={theme['action-button']}
+        icon={<IconCloseMediumOutline />}
+        color="white"
+        onClick={onClose}
+      />
+    );
 
   render() {
     const {
--- a/components/toast/Toast.js
+++ b/components/toast/Toast.js
@@ -5,6 +5,7 @@ import cx from 'classnames';
 import { IconButton, LinkButton } from '../button';
 import { TextBody } from '../typography';
 import LoadingSpinner from '../loadingSpinner';
+import Link from '../link';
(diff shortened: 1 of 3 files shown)


components/toast/Toast.js
```diff
@@ -74,17 +87,19 @@ class Toast extends PureComponent {
...
-                  <IconButton
-                    className={theme['action-button']}
-                    icon={<IconCloseMediumOutline />}
-                    color="white"
-                    onClick={onClose}
-                  />
-                )
+              {action ? (
+                <LinkButton className={theme['action-link']} inverse label={actionLabel} onClick={action} />
+              ) : link ? (
```
duivvv: Not a big fan of nesting ternaries, can we refactor this?

components/toast/theme.css
```diff
@@ -20,29 +21,45 @@
...
+    color: var(--color-teal-light);
+    font-family: var(--font-family-regular);
+    font-size: 14px;
+    letter-spacing: normal;
+    line-height: 19px;
+    text-align: right;
+
+    &:hover {
+      color: var(--color-teal-light);
+    }
```
InstaK: Maybe the `Link` component also needs to be able to receive tints? Then you wouldn't have to do this, right?

Kemosabert: Thanks for the tip. I put the Link in a TextBody component and could thus delete some custom CSS as the Link can inherit styles. 

 :ok: by @Kemosabert on 21/06/2018 13:13:19

components/toast/Toast.js
```diff
@@ -48,6 +48,36 @@ class Toast extends PureComponent {
     }, timeout);
   };
 
+  renderCustomAction = (action, actionLabel) => {
+    if (action) {
+      return <LinkButton className={theme['action-link']} inverse label={actionLabel} onClick={action} />;
+    }
+    return false;
```
InstaK: you can write this:
```js
renderCustomAction = (action, actionLabel) => action && <LinkButton className={theme['action-link']} inverse label={actionLabel} onClick={action} />;
```
If action is falsy the double ampersand will return false, if action is truthy it will return the component.

InstaK: You can actually do that for every single one, returning false will render nothing.

components/toast/Toast.js
```diff
@@ -74,18 +109,9 @@ class Toast extends PureComponent {
...
-                    className={theme['action-button']}
-                    icon={<IconCloseMediumOutline />}
-                    color="white"
-                    onClick={onClose}
-                  />
-                )
-              ) : null}
+              {this.renderCustomAction(action, actionLabel) ||
+                this.renderCustomLink(link, linkLabel, linkTarget) ||
+                this.renderCloseButton(onClose)}
```
nickwaelkens: PP: As you're not doing anything else with these values besides passing them through, I would get the values out of props in the respective render methods:

```jsx
renderCustomAction = () => {
	const { action, actionLabel } = this.props;
    return action && <LinkButton className={theme['action-link']} inverse label={actionLabel} onClick={action} />;
}

render () {
	return (
		<div>
			{
				this.renderCustomAction() || 
				this.renderCustomLink() || 
				this.renderCloseButton()
			}
		</div>
	);
}
```

lowiebenoot:
I'm not sure if the `link` as a string is very useful for us, because we probably want to use a `Link` component from our router. I'm not sure how the router handles other "normal" links (with `<a href="..."`). Do you know @duivvv @timdegroote?

duivvv:
Maybe we can we make this a render prop? Pass in an <a> or <Link>?

Kemosabert:
@lowiebenoot That makes more sense indeed. I changed the component so you can pass in any link component you want basically

components/toast/Toast.js
```diff
@@ -54,24 +53,16 @@ class Toast extends PureComponent {
...
-      link && (
-        <TextBody className={theme['label']} color="white" soft>
-          <Link href={link} target={linkTarget} className={theme['action-link']} inherit>
-            {linkLabel}
-          </Link>
-        </TextBody>
-      )
-    );
+    const { link } = this.props;
+    return link && <TextBody color="white">{React.cloneElement(link, { className: theme['toast-link'] })}</TextBody>;
```
lowiebenoot: * I guess you can remove the `color="white"` here, since you change the color with css as well.
* Not 100% sure, but I think the original `className` will be overwritten from the link element (if it has one). You can probably change it to `cx(link.props.className, theme['toast-link'])`

Kemosabert: I tested the second thing and that was correct. Updated the code in the latest commit.

 :ok: by @Kemosabert on 28/06/2018 17:12:53

components/toast/Toast.js
```diff
@@ -54,7 +54,11 @@ class Toast extends PureComponent {
 
   renderCustomLink = () => {
     const { link } = this.props;
-    return link && <TextBody>{React.cloneElement(link, { className: theme['action-link'] })}</TextBody>;
+    return (
+      link && (
+        <TextBody>{React.cloneElement(link, { className: cx(link.props.className, theme['action-link']) })}</TextBody>
+      )
+    );
```
InstaK: Really small and not blocking, but redundant parentheses:
```js
    return link && (
        <TextBody>{React.cloneElement(link, { className: cx(link.props.className, theme['action-link']) })}</TextBody>
      );
```

Kemosabert: I checked this out and this was added by Prettier when I committed this. I guess it's mainly because the line would have gotten too long.

 :ok: by @Kemosabert on 29/06/2018 09:21:25

Issue #287, opened 2018-06-18
Stacking toasts is not possible


lowiebenoot:
### Expected behavior
When adding multiple toasts they should should be shown above each other.

### Actual behavior
They are shown on top of each other.

Kemosabert:
Fixed in PR #298. Toasts should now be rendered in a toast container

PR #296, opened 2018-06-25
Toast container component

Kemosabert:
### Description
Make a container for the Toasts so they will automatically be stacked. related to issue: #287

### Breaking changes
Toasts should now always be rendered in this container component.

### DIFF
--- a/components/toast/ToastContainer.js
+++ b/components/toast/ToastContainer.js
@@ -1,0 +1,21 @@
+import React, { PureComponent } from 'react';
+import PropTypes from 'prop-types';
+import theme from './theme.css';
+import cx from 'classnames';
+
+class ToastContainer extends PureComponent {
+  render() {
+    const { children } = this.props;
+
+    const classNames = cx(theme['container']);
+
+    return <div className={classNames}>{children}</div>;
+  }
+}
+
+ToastContainer.propTypes = {
+  /** List of toasts to be displayed */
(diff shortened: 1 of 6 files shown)


components/toast/ToastContainer.js
```diff
@@ -0,0 +1,21 @@
...
+import theme from './theme.css';
+import cx from 'classnames';
+
+class ToastContainer extends PureComponent {
+  render() {
+    const { children } = this.props;
+
+    const classNames = cx(theme['container']);
+
+    return <div className={classNames}>{children}</div>;
```
nickwaelkens: Semantically probably makes more sense to use `<ul>`

lowiebenoot:
* [ ] The closing (both manually and automatically) of the toasts doesn't work in the story
* [ ] The closing should "pause" when we hover the toast container
* [ ] We need to get rid of the `active` prop on the `Toast` component. It doesn't make sense that it exists. If we render a Toast we'll always want it to be visible, otherwise we wouldn't render it. Now it actually just exists for animation reasons I think.
* [ ] The other stories are still without the `ToastContainer` and they are a bit bugged, you probably need to add the `ToastContainer` there as well
* [ ] The stacked toasts story is a bit weird, because it only renders something when you activate a toast via the "knobs". Can we avoid this and use a button in in the story instead and make as many toasts as we want? That will show the stacking much better.

PR #298, opened 2018-06-28
Render Toasts in a container and update animation

Kemosabert:
### Description
Toasts no longer need the active prop to be rendered and can now be stacked as they are rendered in a containing element.

### Breaking changes
Toasts should be rendered inside the ToastContainer element which will provide flawless animations.

**NOTE:**
To display this in the stories we have two ways with both their ad- and disadvantages. I provided now two ways from which we can pick. If anyone has a solution to include all the ad- and no disadvantages, shoot! 
1. We can have the toast appearing and disappearing by controlling a numbered input field at the bottom. This has the advantage that it shows the full code with all the props in the preview. (With Close button example)

2. We can have the toasts interactive as in that they will appear on the click of a button and close when clicked or timed out. But this means we can not see the props on the Toasts themselves and see the State component in the code preview. (other examples)

I would love to hear what you guys think is the best way.

### DIFF
--- a/components/toast/Toast.js
+++ b/components/toast/Toast.js
@@ -5,16 +5,11 @@ import cx from 'classnames';
 import { IconButton, LinkButton } from '../button';
 import { TextBody } from '../typography';
 import LoadingSpinner from '../loadingSpinner';
-import { createPortal } from 'react-dom';
 import { IconCloseMediumOutline } from '@teamleader/ui-icons';
 import theme from './theme.css';
 
 class Toast extends PureComponent {
-  toastRoot = document.createElement('div');
-
   componentDidMount() {
-    document.body.appendChild(this.toastRoot);
-
     if (this.props.active && this.props.timeout) {
       this.scheduleTimeout(this.props);
     }
@@ -28,7 +23,6 @@ class Toast extends PureComponent {
 
   componentWillUnmount() {
     clearTimeout(this.currentTimeout);
-    document.body.removeChild(this.toastRoot);
   }
 
   scheduleTimeout = props => {
@@ -74,7 +68,7 @@ class Toast extends PureComponent {
   render() {
     const { active, children, className, label, processing } = this.props;
 
-    const toast = (
+    return (
       <Transition in={active} timeout={{ enter: 0, exit: 1000 }}>
         {state => {
           if (state === 'exited') {
@@ -104,8 +98,6 @@ class Toast extends PureComponent {
         }}
       </Transition>
     );
-
-    return createPortal(toast, this.toastRoot);
   }
 }
 
--- a/components/toast/Toast.js
+++ b/components/toast/Toast.js
@@ -74,7 +74,7 @@ class Toast extends PureComponent {
     const classNames = cx(theme['toast'], className);
 
     return (
-      <div className={theme['toast-wrapper']}>
+      <div>
         <div data-teamleader-ui="toast" className={classNames}>
           {processing && <LoadingSpinner className={theme['spinner']} color="white" />}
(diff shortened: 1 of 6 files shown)


components/toast/Toast.js
```diff
@@ -1,34 +1,27 @@
...
   componentDidMount() {
-    document.body.appendChild(this.toastRoot);
-
-    if (this.props.active && this.props.timeout) {
+    if (this.props.timeout) {
       this.scheduleTimeout(this.props);
     }
   }
 
   componentWillReceiveProps(nextProps) {
```
JorenSaey: PP: since code was already there, but componentWillRecieveProps is deprecated. Perhaps use componentDidUpdate. Or is this part of a later PR?

components/toast/theme.css
```diff
@@ -57,13 +71,36 @@
...
+  transition: opacity 0.3s var(--toast-animation-curve), transform 0.3s var(--toast-animation-curve),
+    max-height 0.3s var(--toast-animation-curve);
+}
+.enter-done {
+  opacity: 1;
+}
+.exit {
+  opacity: 1;
+  max-height: 180px;
+  transform: translateX(0);
```
InstaK: Could you use translate3D? https://jsperf.com/translate3d-vs-xy

Kemosabert: Thanks for the tip, did not really know about the performance difference

 :ok: by @Kemosabert on 29/06/2018 13:17:04

components/toast/Toast.js
```diff
@@ -76,40 +69,22 @@ class Toast extends PureComponent {
...
-                {children}
-              </TextBody>
-              {this.renderCustomAction() || this.renderCustomLink() || this.renderCloseButton()}
-            </div>
-          );
-        }}
-      </Transition>
+    return (
+      <div className={theme['toast-wrapper']}>
+        <div data-teamleader-ui="toast" className={classNames}>
```
lowiebenoot: `toast-wrapper` is not used.
And FYI: we like to put the `data-teamleader-ui` attribute on the root element.

components/toast/Toast.js
```diff
@@ -76,40 +69,22 @@ class Toast extends PureComponent {
...
-                {children}
-              </TextBody>
-              {this.renderCustomAction() || this.renderCustomLink() || this.renderCloseButton()}
-            </div>
-          );
-        }}
-      </Transition>
+    return (
+      <div data-teamleader-ui="toast">
+        <div className={classNames}>
```
InstaK: Q: Just wondering now, is this second div still necessary? Isn't it possible to put `className={classNames}` on the div above it and get rid of this one?

Kemosabert: Good question. I came to doing this as follows:

How the transition works now when a toast disappears is as follows: in the first phase it goes to the right and fades out at the same time. With a small delay, the 2nd phase starts in which I set the toasts max-height to 0 so we can drop the remaining toast stack nice and gently. However, margins and paddings were still left, so when the toast leaves the DOM, it still had a height of +- 30px due to padding and bottom-margin still being applied, resulting in a very snappy animation and a jump. I tried to fix this by animating the top and bottom padding, but this was very laggy and jumpy as was the margin animation as well if applied at all. To overcome this I added another component around it without any margin or padding so it would not be affected by the laggy animations from the padding and margin.  

Thus: as we don't want any spacing on that outer div to keep control over animations, and any other styles would be redundant there, I don't think it makes sense to apply classNames on the outer div.

InstaK: Fair enough 👍 

 :ok: by @InstaK on 03/07/2018 13:42:18

PR #300, opened 2018-06-29
Put isDark function in utils

Kemosabert:
### Description
We had some duplicate code across multiple components to decide whether a component should have a dark border or not. This PR puts that function in the utils folder so it can be shared cross-component

### Breaking changes
none


lowiebenoot:
I think the function name is wrong actually. It's called `elementIsDark` but you pass a color as param 🤔

PR #301, opened 2018-06-29
Fix global bullet styles

Kemosabert:
### Description
In re-doing the bullets I added some global styles to headers and paragraphs so the inline bullets would always be perfectly aligned with the text it is inline with. This is however not really what I wanted so I removed that CSS and also had to change how inline bullets should be build.

### Breaking changes
None

### DIFF
--- a/components/bullet/theme.css
+++ b/components/bullet/theme.css
@@ -9,14 +9,6 @@
   border-radius: 50%;
   display: inline-block;
   vertical-align: middle;
-
-  & + h1,
-  & + h2,
-  & + h3,
-  & + h4,
-  & + p {
-    margin-left: 6px;
-  }
 }
 
 .medium {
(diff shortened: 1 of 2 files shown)


components/bullet/theme.css
```diff
@@ -9,6 +9,14 @@
...
   display: inline-block;
   vertical-align: middle;
+
+  & + h1,
+  & + h2,
+  & + h3,
+  & + h4,
+  & + p {
+    margin-left: 6px;
+  }
```
nickwaelkens: Is it me or does this feel iffy?

Kemosabert: The spacing between Bullets and inline text was defined in the styleguide so I wanted to include it in the component. As I think more about it now, it might not make that much sense anymore as we force people to combine it in a certain way.

nickwaelkens: Exactly! Components should be standalone, not sure how we can fix this properly though.

Kemosabert: <Heading1WithBullet />

Kemosabert: I reverted the commits where I change the structure of the example and the weird spacing CSS so this PR only focuses on the global styles it added.

Kemosabert:
I have reduced this commit to removing the CSS which caused all text elements to have the display: flex styles.

Issue #304, opened 2018-07-02
Popover doesn't reposition after scrollbar is added


InstaK:
### Expected behavior
- The Popover component to recalculate its position when a scrollbar is added while the Popover is open

### Actual behavior
- The Popover stays on the same position while the rest of the page moves slightly to the left or right if a scrollbar gets added or removed

### Steps to reproduce actual behavior
1. Find a place where a Popover is being used
2. Open the Popover
3. Resize your screen in order to get or remove the scrollbar

### Screenshot of when the issue occures
![screen shot 2018-07-02 at 14 06 13](https://user-images.githubusercontent.com/9056632/42163107-8ce0fc04-7e01-11e8-99c6-6461bf800bc0.jpg)

PR #305, opened 2018-07-03
Re-calculate Popover position on screen resize

Kemosabert:
### Description
Fix for issue #304. The issue here was that the throttle function did not work as expected due to how event pooling works in React, hence the callback function was rarely called. Fixed this 

### Breaking changes
none

### DIFF
--- a/components/popover/Popover.js
+++ b/components/popover/Popover.js
@@ -26,8 +26,6 @@ const factory = (axis, calculatePositions, Overlay) => {
     componentDidMount() {
       document.body.appendChild(this.popoverRoot);
 
-      this.setPlacementThrottled = throttle(this.setPlacement, 250);
-
       events.addEventsToWindow({
         resize: this.setPlacementThrottled,
         scroll: this.setPlacementThrottled,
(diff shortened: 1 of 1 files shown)


components/popover/Popover.js
```diff
@@ -23,11 +23,11 @@ const factory = (axis, calculatePositions, Overlay) => {
       },
     };
 
-    setPlacementThrottled = () => throttle(this.setPlacement, 16);
-
     componentDidMount() {
       document.body.appendChild(this.popoverRoot);
 
+      this.setPlacementThrottled = throttle(this.setPlacement, 250);
```
lowiebenoot: Why did it move to `componentDidMount` and why did you change it to 250ms? (16ms is to get approximately 60 fps)

Kemosabert: I put it in componentDidMount because when I put it outside, it threw an error, saying throttle expected a function. I just checked again, and if I put this line after the setPlacement() function, it does work. What would be the reasoning for not putting it in componentDidMount()?

I don't believe the repositioning calculation needs to happen at 60fps and I believe it might have an impact on performance if you have this happening on pages that also have a lot of things happening on resizing of the window.

lowiebenoot: In my opinion it's a bit weird to put it in a lifecycle function, as it has nothing to do with the lifecycle. I would just put it after the `setPlacement` function (or in a constructor, but I prefer the first one).

Kemosabert: That makes sense, I moved the function.

 :ok: by @Kemosabert on 03/07/2018 14:18:32

PR #307, opened 2018-07-04
Enhancement/pause toast timeout on hover

Kemosabert:
### Description
Hovering over a toast will now cancel it's timeout if one has been set, until the mouse leaves the toast after which the timeout will start over. 

### Breaking changes
none

### DIFF
--- a/components/toast/Toast.js
+++ b/components/toast/Toast.js
@@ -68,13 +68,26 @@ class Toast extends PureComponent {
     );
   };
 
+  handleMouseEnter = () => {
+    if (this.props.timeout) {
+      clearTimeout(this.currentTimeout);
+      this.currentTimeout = null;
+    }
+  };
+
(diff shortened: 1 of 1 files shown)


nickwaelkens:
Thinking out loud here, but maybe we need to have an extra check/control that makes sure that, if a user accidentally hovers a toast, that the timeout does not reset. Something like `enterDelay` and `leaveDelay`.

Inspired by: https://github.com/jaredLunde/render-props/tree/master/packages/hover

Kemosabert:
I do not really agree with this. Users don't know how long a certain toast should be visible, so in case they accidentally hover, the impact would not be negative I think. Also, a very very edge case, what if a user intentionally hovers during the last 500ms, what is the behaviour then of the delay? Does it take this into account?

nickwaelkens:
Fair enough, feel free to discard my comment and ship it

PR #309, opened 2018-07-06
Indicate that a user has selected some datagrid items

Kemosabert:
### Description
If a Datagrid's items are selectable, it will now also show when some items have been selected by showing a "-" icon in the checkbox as indicator to the user.

screen shot 2018-07-06 at 11 34 14

#### Screenshot after this PR
screen shot 2018-07-06 at 11 35 41

### Breaking changes
none

### DIFF
--- a/components/checkbox/Checkbox.js
+++ b/components/checkbox/Checkbox.js
@@ -98,11 +98,7 @@ class Checkbox extends PureComponent {
           {...inputProps}
         />
         <span className={theme['shape']}>
-          {partiallySelected ? (
-            <IconMinusSmallOutline className={theme['icon']} />
-          ) : (
-            <IconCheckmark className={theme['icon']} />
-          )}
+          {checked ? <IconCheckmark className={theme['icon']} /> : <IconMinusSmallOutline className={theme['icon']} />}
         </span>
         {(label || children) && (
           <span className={theme['label']}>
--- a/components/checkbox/Checkbox.js
+++ b/components/checkbox/Checkbox.js
@@ -67,7 +67,7 @@ class Checkbox extends PureComponent {
   }
 
   render() {
-    const { checked, disabled, className, size, label, children, partiallySelected, ...others } = this.props;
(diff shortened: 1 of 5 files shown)


lowiebenoot:
Maybe add a story for it under checkbox as well.

components/checkbox/Checkbox.js
```diff
@@ -98,7 +98,11 @@ class Checkbox extends PureComponent {
           {...inputProps}
         />
         <span className={theme['shape']}>
-          <IconCheckmark className={theme['checkmark']} />
+          {partiallySelected ? (
```
lowiebenoot: I would turn them around.
If you set both `partiallySelected` and `checked` to true, I would expect `checked` to get the upper hand, and show a checked checkbox.

driesd: Damn good catch!

components/checkbox/Checkbox.js
```diff
@@ -67,7 +67,7 @@ class Checkbox extends PureComponent {
   }
 
   render() {
-    const { checked, disabled, className, size, label, children, ...others } = this.props;
+    const { checked, disabled, className, size, label, children, partiallySelected, ...others } = this.props;
```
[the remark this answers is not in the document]
Kemosabert: That is actually very true, the variable here is poorly named as we describe a state of the checkbox itself, not what is happening around it. I think "indeterminate" is a perfect word/name for it, but also not sure how clear this is for others. What do you think about this?

components/checkbox/Checkbox.js
Kemosabert: Good catch, although I already made sure in the logic in the datagrid component that both checked  and partiallyChecked can not be true at the same time.

 :ok: by @Kemosabert on 06/07/2018 15:45:49

components/checkbox/Checkbox.js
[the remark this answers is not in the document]
lowiebenoot: `indeterminate` seems good to me. It's not the clearest name, but as it is documented, people will know what it means (and people will probably don't need it very often).

Issue #315, opened 2018-07-24
Indeterminate of DataGrid does not reset after comparableId changes


InstaK:
### Expected behavior

The checkboxes of a datagrid get reset when the comparableId changes, however:

### Actual behavior

the indeterminate checkbox does not.

### Steps to reproduce actual behavior

1.  Have a DataGrid
2.  Select 1 or more items smaller than the total amount in order to get the indeterminate icon
3.  Get the comparableId to change
4.  Notice that the indeterminate icon stays

### Screenshot of when the issue occures

![screen shot 2018-07-24 at 13 55 37](https://user-images.githubusercontent.com/9056632/43136531-4e404cee-8f49-11e8-82c6-83cb9ffbcb8c.jpg)

after comparableId changes:

![screen shot 2018-07-24 at 13 55 48](https://user-images.githubusercontent.com/9056632/43136535-51d088ce-8f49-11e8-82b8-db3dd06bf3a3.jpg)

PR #319, opened 2018-07-24
Reset indeterminate to false of parent checkbox when comparableID changes

Kemosabert:
### Description
Fix for issue #315 

### Breaking changes
none

### DIFF
--- a/components/datagrid/DataGrid.js
+++ b/components/datagrid/DataGrid.js
@@ -119,7 +119,7 @@ class DataGrid extends PureComponent {
       [theme['has-border-right']]: selectable || stickyFromLeft > 0,
     });
 
-    const numberOfBodyRows = children.flat().filter(child => isComponentOfType(BodyRow, child)).length;
+    const numberOfBodyRows = React.Children.toArray(children).filter(child => isComponentOfType(BodyRow, child)).length;
     const isPartiallySelected = selectedRows.length > 0 && selectedRows.length !== numberOfBodyRows;
 
     return (


duivvv:
That makes a lot of sense, try to derive as much as you can from state / props.

Kemosabert:
Sounds better indeed actually.

components/datagrid/DataGrid.js
```diff
@@ -123,6 +119,8 @@ class DataGrid extends PureComponent {
       [theme['has-border-right']]: selectable || stickyFromLeft > 0,
     });
 
+    const isPartiallySelected = selectedRows.length > 0 && selectedRows.length !== children[1].length;
```
lowiebenoot: What's `children[1]`? Something tells me there's something fishy about this (although the code was actually already in here).

Kemosabert: It's indeed a bit weird. It's because the children consists of 2 or more nested arrays. The first one being the HeaderRow elements, the 2nd one being the BodyRow elements and the 3rd the FooterRow elements. In the render method where they map over the children, they check the type with isComponentOfType to render the correct element. Should I flat it out and then filter to take the length of that array?

duivvv: what about

```js

const [headerRow, bodyRow] = children;

```

for readability?

lowiebenoot: I actually think you're a bit mistaken.
Datagrid can contain multiple `BodyRow`'s as children, for example:


```js
<DataGrid>
	<HeaderRow />

	<BodyRow />
	<BodyRow />
	<BodyRow />

	<FooterRow />
	<FooterRow />
</DataGrid>
```

So (if I'm correct), `children[1]` is actually just a single `BodyRow`. It's also not (technically) required to have a `HeaderRow` I think.

I think the best solution here is to filter the children with `isComponentOfType`.

Kemosabert: Hm, okay, I will make sure to filter the children, so we are prepared for the edge cases.

 :ok: by @Kemosabert on 25/07/2018 09:20:13

components/datagrid/DataGrid.js
```diff
@@ -119,7 +119,8 @@ class DataGrid extends PureComponent {
       [theme['has-border-right']]: selectable || stickyFromLeft > 0,
     });
 
-    const isPartiallySelected = selectedRows.length > 0 && selectedRows.length !== children[1].length;
+    const numberOfBodyRows = children.flat().filter(child => isComponentOfType(BodyRow, child)).length;
```
duivvv: Are we polyfilling flat() ?
It's a VERY new array function.

Kemosabert: Upon checking, its indeed very new. Changed it so I use React.Children.toArray() to flatten it out.

 :ok: by @Kemosabert on 25/07/2018 09:52:32

PR #321, opened 2018-07-25
Don't vertically center or resize the icon for when error is multiline

timdegroote:
This fixes issues with the error icon when the error spans multiple lines:
screen shot 2018-07-25 at 12 43 06


InstaK:
🚢

PR #325, opened 2018-07-30
♻️ Center-align cells when DataGrid is selectable

InstaK:
### Description

The Checkboxes of a selectable DataGrid used to take up the whole container they were in. Because they now allow different sizes, when the checkboxSize is small, it's not aligned. Now it is.

screenshot 2018-07-30 17 07 12

#### Screenshot after this PR

screenshot 2018-07-30 17 07 37

### Breaking changes

- none

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,7 +5,7 @@
 ### Changed
 
 - `IconMenu` now returns a `Box` component instead of a `div` ([@InstaK](https://github.com/InstaK) in [#324](https://github.com/teamleadercrm/ui/pull/324))
-- `HeaderCell` and `BodyCell` of `DataGrid` now return a center-aligned `Cell` ([@InstaK](https://github.com/InstaK) in [#325](https://github.com/teamleadercrm/ui/pull/325))
+- The `Checkbox` in `DataGrid` is now centered ([@InstaK](https://github.com/InstaK) in [#325](https://github.com/teamleadercrm/ui/pull/325))
 
 ### Deprecated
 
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -5,6 +5,7 @@
 ### Changed
 
 - `IconMenu` now returns a `Box` component instead of a `div` ([@InstaK](https://github.com/InstaK) in [#324](https://github.com/teamleadercrm/ui/pull/324))
+- `HeaderCell` and `BodyCell` of `DataGrid` now return a center-aligned `Cell` ([@InstaK](https://github.com/InstaK) in [#325](https://github.com/teamleadercrm/ui/pull/325))
```
lowiebenoot: This message is very vague, it's telling how you technically did it, but it's not telling what it actualy changes.

A better message here would be about centering the checkbox.

InstaK: So what's the difference with the message above which is in the format @nickwaelkens asked me to put it in?

lowiebenoot: Not a very big difference, but the intention of the one above was to replace the div with a box, so you can add box props to the IconMenu. So it's still vague, but it's actually saying what you did in the PR.

With your current change, the intention was to center the checkbox in the datagrid. It's quite hard to find the intention of the change in your message. In fact, it's not even correct. Now it looks like every cell in the datagrid should be center-aligned, while it's only the one with the checkbox.

nickwaelkens: The previous thing you changed was actually a technical thing (returning a `Box` instead of `div`) and that was your intention, now your intention is indeed something else!

InstaK: Okay I'm convinced 😄 I'ma change it

PR #327, opened 2018-08-01
Update start script

LennertBlommaert:
### Description

This PR removes the need to pass the port when running `npm start`.
It defaults now directly to port `3000`, which can be overridden.
Starting up the project locally is now easier and more in line with the other teamleadercrm repo's.

Not applicable.

#### Screenshot after this PR

Not applicable.

### Breaking changes

None.


nickwaelkens:
It's the small things like this. 👏

Issue #299, opened 2018-06-29
Stop using the deprecated componentWillReceiveProps lifecycle hook


Kemosabert:
### Expected behavior
In some components, DataGrid, Menu & Toast we are using componentWillReceiveProps which has been deprecated since 16.3.0. We should switch those out for a different lifecycle hook.

PR #330, opened 2018-08-02
Refactor menu component

LennertBlommaert:
### Description

This PR refactors the menu component.
Mainly, it removes the deprecated `compenentWillUpdate` and `componentWillRecieveProps`.
As this component had been copied from another library, I also took the opportunity to write most code more clearly and structured.

This PR resolves issue #299, together with PR #331 

Nothing has visually changed.

screen shot 2018-08-02 at 12 30 33

#### Screenshot after this PR

screen shot 2018-08-02 at 12 29 20

### Breaking changes
None.

### DIFF
--- a/components/menu/Menu.js
+++ b/components/menu/Menu.js
@@ -191,19 +191,21 @@ class Menu extends PureComponent {
   }
 
   render() {
-    const outlineStyle = { width: this.state.width, height: this.state.height };
-    const className = cx(
+    const { width, height, active, position } = this.state;
+    const { className, outline } = this.props;
+
+    const clasNames = cx(
       theme['menu'],
-      theme[this.state.position],
+      theme[position],
       {
-        [theme['active']]: this.state.active,
+        [theme['active']]: active,
       },
-      this.props.className,
+      className,
     );
 
     return (
-      <div data-teamleader-ui="menu" className={className} style={this.getRootStyle()}>
-        {this.props.outline ? <div className={theme['outline']} style={outlineStyle} /> : null}
+      <div data-teamleader-ui="menu" className={clasNames} style={this.getRootStyle()}>
+        {outline ? <div className={theme['outline']} style={{ width, height }} /> : null}
         <ul
           ref={node => {
             this.menuNode = node;
--- a/components/menu/Menu.js
+++ b/components/menu/Menu.js
@@ -68,10 +68,7 @@ class Menu extends PureComponent {
 
   componentWillUpdate(nextProps, nextState) {
     if (!this.state.active && nextState.active) {
-      events.addEventsToDocument({
-        click: this.handleDocumentClick,
-        touchstart: this.handleDocumentClick,
-      });
+      this.addEvents();
     }
   }
 
@@ -80,10 +77,7 @@ class Menu extends PureComponent {
       if (this.props.onHide) {
         this.props.onHide();
       }
-      events.removeEventsFromDocument({
-        click: this.handleDocumentClick,
-        touchstart: this.handleDocumentClick,
-      });
+      this.removeEvents();
     } else if (!prevState.active && this.state.active && this.props.onShow) {
       this.props.onShow();
     }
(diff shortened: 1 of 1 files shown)


components/menu/Menu.js
```diff
@@ -17,162 +17,191 @@ const POSITION = {
 };
 
 class Menu extends PureComponent {
-  state = {
-    active: this.props.active,
-  };
+  state = {};
```
InstaK: Not sure if
```js
state = {};
```
has any functionality, isn't this default?

Then again, I think we're sure the initial state will always be `active: false`, right? Might as well be that then?

components/menu/Menu.js
```diff
@@ -17,162 +17,191 @@ const POSITION = {
...
-      const { width, height } = this.menuNode.getBoundingClientRect();
-      const position = this.props.position === POSITION.AUTO ? this.calculatePosition() : this.props.position;
-      this.setState({ position, width, height });
-    });
+    const { width, height } = this.menuNode.getBoundingClientRect();
+    this.setState({ width, height });
+
+    const { position } = this.props;
+    if (position === POSITION.AUTO) {
+      this.setState({ position: this.calculatePosition });
```
InstaK: Q: Is this correct? Now you're assigning a function to `position`, right? While you want to have the calculated number? So I'd say this needs to be 
```js
this.setState({ position: this.calculatePosition() });
```

components/menu/Menu.js
```diff
@@ -17,162 +17,191 @@ const POSITION = {
...
-    /**
-     * If the menu is being deactivated via props and the current state is
-     * active, it should be hid.
-     */
-    if (this.props.active && !nextProps.active && this.state.active) {
+    if (newStateSlice.active || newStateSlice.position) {
+      return newStateSlice;
+    }
+
+    return null;
```
InstaK: Q: Why do you want to specifically return null?

components/menu/Menu.js
```diff
@@ -17,162 +17,191 @@ const POSITION = {
...
+  }
+
+  getMenuStyle() {
+    const { active } = this.state;
+
+    if (active) {
+      return this.getActiveMenuStyle();
+    }
+
+    return this.getMenyStyleByPosition();
```
InstaK: ✏️  `getMenuStyleByPosition`

components/menu/Menu.js
nickwaelkens: Correct and correct!

components/menu/Menu.js
LennertBlommaert: To indicate that no state should be updated, it is specified in the (React docs)[https://reactjs.org/docs/react-component.html#static-getderivedstatefromprops].

InstaK: Aha. I was wondering why they wouldn't allow `undefined` because that's what a function returns, but I guess it's because you might say `return object` while `object` is undefined... But then again, that would also be fine because then it would be the same result as we have now by returning `null` 🤔 ... Not sure why they chose this... But idd it's in the docs 🤷‍♂️

components/menu/Menu.js
LennertBlommaert: `active` in the state, takes the value of the `active` prop, so the menu can be opened from its parent container, `getDerivedStateFromProps` takes care of this. `getDerivedStateFromProps` is also called before the initial mount, so it makes sure that the value of `active` in the state before the mount, is set correctly (aka set to the value of the (default)prop `active`).

Removing the `state = {};` line, throws following error: `Cannot read property 'x' of null`, when calling the `state`. So I think it still needs to be declared.

components/menu/Menu.js
```diff
@@ -181,37 +207,30 @@ class Menu extends PureComponent {
...
-    this.setState({ active: false });
-  }
-
   render() {
-    const outlineStyle = { width: this.state.width, height: this.state.height };
-    const className = cx(
+    const { width, height, active, position } = this.state;
+    const { className, outline } = this.props;
+
+    const clasNames = cx(
```
nickwaelkens: Typo?

nickwaelkens:
We tend to use the same logic over and over to calculate where something should be rendered (popover, tooltip, menu, ...) - Time for a HoC (with render props)?

components/menu/Menu.js
LennertBlommaert: @nickwaelkens @InstaK what do you guys think?

components/menu/Menu.js
nickwaelkens: You actually forgot to change it here?

PR #332, opened 2018-08-02
[TU-95] Split up Text and Headings into separate components

driesd:
### Description

This PR splits up our `Text` component into separate components:
* Heading1
* Heading2
* Heading3
* Heading4
* TextDisplay
* TextBody
* TextSmall

**Main reason for this:** the documentation in Storybook just wasn't right. It was displaying the wrong component names. Also there were way too much prop tables shown, which was far from clear to other developers. 

schermafdruk 2018-08-02 16 38 09

schermafdruk 2018-08-02 16 38 31

#### Screenshot after this PR

schermafdruk 2018-08-02 16 36 16

schermafdruk 2018-08-02 16 36 30

### Breaking changes

None.


lowiebenoot:
It's a bit sad though, that so many code is repeated now, instead of just having the factory.

driesd:
I know, It is a bit sad idd... But at least the documentation is now correct! :)

nickwaelkens:
I have to agree with @lowiebenoot - this will add quite some duplicate code to our (already quite big) bundles, right? All that, just so we can get something working that is probably considered a bug in Storybook? Have you tried opening an issue on their repo, ask for pointers?

We should get to the root of the problem instead 🤔

thomasnoe07:
@driesd  Can this PR get moving again or is it blocked?

driesd:
-

PR #336, opened 2018-08-09
[TU-88] Button: custom element override

driesd:
### Description

This PR implements the `element` prop for `Button` and `IconButton` which takes an element or string to be visually rendered as a styled button.

### Breaking changes

`Button` & `IconButton` won't render an anchor tag by only passing the `href` prop anymore. From now on, you also need to set the `element` prop to `'a'`.

### DIFF
--- a/components/button/Button.js
+++ b/components/button/Button.js
@@ -101,7 +101,7 @@ Button.propTypes = {
   /** A class name for the button to give custom styles. */
   className: PropTypes.string,
   /** A custom element to be rendered */
-  element: PropTypes.element,
+  element: PropTypes.oneOfType([PropTypes.element, PropTypes.string]),
   /** Determines which kind of button to be rendered. */
   level: PropTypes.oneOf(['outline', 'primary', 'secondary', 'destructive']),
   /** If true, component will be disabled. */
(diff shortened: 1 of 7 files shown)


components/button/Button.js
```diff
@@ -62,15 +60,14 @@ class Button extends PureComponent {
...
-      href,
       ref: node => {
         this.buttonNode = node;
       },
       className: classNames,
       disabled,
       onMouseUp: this.handleMouseUp,
       onMouseLeave: this.handleMouseLeave,
-      type: !href ? type : null,
+      type: element === 'button' ? type : null,
```
lowiebenoot: I think we can actually remove the `type` prop, and just leave it in the `others`.

driesd: Yeah but, if you include a button in a form element without specifying it's just a regular button, it defaults to a submit button.

stories/button.js
```diff
@@ -7,6 +7,7 @@ import { withKnobs, boolean, select } from '@storybook/addon-knobs/react';
 import { IconAddMediumOutline, IconAddSmallOutline } from '@teamleader/ui-icons';
 import { Button } from '../components';
 
+const elements = ['a', 'button', 'div', 'span'];
```
timdegroote: I'd prefer to remove the `div` and `span` examples from the elements array as this isn't something we should be doing anyway for accessibility reasons.

duivvv:
I guess that's a no.

PR #341, opened 2018-08-22
[TU-105] LinkButton display as square

LennertBlommaert:
### Description

This PR updates the styling of the `LinkButton` component, as it didn't display as a square when only containing an icon, due to wrong padding.

screen shot 2018-08-22 at 10 41 08

#### Screenshot after this PR

screen shot 2018-08-22 at 10 41 42

### Breaking changes

None.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,8 @@
 
 ### Fixed
 
+- `LinkButton`: set the appropriate padding, when only containing an icon, so it displays as a square ([@LennertBlommaert](https://github.com/LennertBlommaert) in [#341](https://github.com/teamleadercrm/ui/pull/341)).
+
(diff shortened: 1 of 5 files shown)


InstaK:
@nickwaelkens He uses a small or large chevron? Those have different CSS I suppose

LennertBlommaert:
@nickwaelkens: in the shared CSS of the `.button`s, a class selector `.icon-only` made sure the horizontal padding (and therefore the visual width) got set properly, when there is only an icon. The `LinkButton` applied the `has-icon-only` class, changing it to also use `icon-only` fixed a part of the problem.
For a small `Button` a different size of icon is needed, as a bigger one would also change the `Button`'s dimension. 
(Note: I will rename all `icon-only` class declarations to `has-icon-only` as the latter declaration is preferred)

Issue #303, opened 2018-07-02
Popover renders offscreen


InstaK:
### Expected behavior
- Popover to be onscreen scrollable

### Actual behavior
- A popover gets rendered the top side when a screen is not big enough to fit it. This makes it impossible to see any content anymore, this is not user-friendly at all.

### Steps to reproduce actual behavior
1. Find a place where a Popover is being used
2. Make sure your screen is smaller than the Popover is big
3. See that the Popover content is not visible.

### Screenshot of when the issue occures
Popover fits:
![screen shot 2018-07-02 at 14 05 37](https://user-images.githubusercontent.com/9056632/42163003-2e6d4fc4-7e01-11e8-8a39-8894f67e6dd8.jpg)

Popover doesn't fit:
![screen shot 2018-07-02 at 14 05 54](https://user-images.githubusercontent.com/9056632/42163012-3437a97c-7e01-11e8-86c0-4cbba6ca8da8.jpg)

Kemosabert:
This is due to the programmed behaviour of the element.

But what should the expected behaviour be? Should we add a max-height/width and add a scrollbar and keep it in the desired direction?

InstaK:
Yeah I know it's programmed to do so :)

Yes, imo we should have a scrollbar. But the tricky part is that we only want a scrollbar when it's not possible to render it in another direction.

So let's say that a popover is meant to be downwards, and there is no room for it in the lower part of the screen, then it should be rendered upwards if there's room for it. However, if the screen doesn't even have enough space to properly render the popover in the other direction, it should be in the original direction with a scrollbar.

PR #342, opened 2018-08-22
Fix popover renders offscreen

LennertBlommaert:
### Description

This PR solves issue #303: popover renders offscreen.

### Breaking changes

None.


LennertBlommaert:
Closing the PR, as I will refactor the component first due to its complexity, which makes fixing this bug harder.

PR #344, opened 2018-08-23
Update Input

driesd:
### Description

This PR:
* makes sure we only render `number type attributes` to `number type input tags`;
* only applies `has-spinner` css class when input type is number;
* allows the `name` and any other input attributes to be rendered on input tags;
* removes props that are not used in the component, they will be rendered by the restProps.

### Breaking changes

None.

### DIFF
--- a/components/input/Input.js
+++ b/components/input/Input.js
@@ -111,15 +111,15 @@ export default class Input extends PureComponent {
   }
 
   renderInput() {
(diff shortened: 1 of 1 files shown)


driesd:
@lowiebenoot you're right! So, I omitted the Box props and all the `Input` component props that are not allowed on `input` elements.

driesd:
@nickwaelkens idd! Coming up in next PR: `parse/format`

PR #345, opened 2018-08-29
Add an Async select component

timdegroote:
### Description

A new select component that will load its options asynchronously and optionally paginate them.

It will load its options by calling a function `loadOptions` which is passed through props. This function is called with the current query that the user has typed and optionally the pageSize and pageNumber if options are supposed to be paginated (also configurable through props).

### DIFF
--- a/components/select/AsyncSelect.js
+++ b/components/select/AsyncSelect.js
@@ -1,0 +1,168 @@
+import React, { PureComponent } from 'react';
+import omit from 'lodash.omit';
+import Select from './Select';
+import PropTypes from 'prop-types';
+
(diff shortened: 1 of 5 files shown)


components/select/AsyncSelect.js
```diff
@@ -0,0 +1,168 @@
...
+  };
+
+  invalidateCache() {
+    this.setState({
+      cache: {},
+    });
+  }
+
+  getPageSize(props = this.props) {
+    return props.pageSize || 10;
```
lowiebenoot: Can't we just use `defaultProps` instead?

PR #346, opened 2018-09-03
Input: remove precision prop

driesd:
### Description

Due to buggy behaviour with number type input fields, we removed the `precision` prop.

This PR actually solves two issues:
* Can't type a dot in an Input with type=number ([TU-99](https://teamleader.atlassian.net/browse/TU-99))
* Difficult to change number input value with certain min/max combinations ([TU-107](https://teamleader.atlassian.net/browse/TU-107))

### Breaking changes

- We removed the `precision` prop

### DIFF
--- a/components/input/Input.js
+++ b/components/input/Input.js
@@ -35,21 +35,18 @@ export default class Input extends PureComponent {
     return finalValue || '';
   }
 
-  static parseValue(value, { type, min, max, precision }) {
-    return type === 'number' ? Input.toNumber(value, min, max, precision) : value;
+  static parseValue(value, { type, min, max }) {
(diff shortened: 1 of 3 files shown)


components/input/Input.js
```diff
@@ -93,13 +90,9 @@ export default class Input extends PureComponent {
...
   formatNumber(number) {
-    const { min, max, precision } = this.props;
-
-    let formattedNumber = Input.toNumber(number, min, max, precision);
+    const { min, max } = this.props;
 
-    if (precision !== null) {
-      formattedNumber = number.toFixed(precision);
-    }
+    let formattedNumber = Input.toNumber(number, min, max);
```
nickwaelkens: Can become a `const` now, or just get rid of the variable all together.

Issue #247, opened 2018-05-07
Input should not use input and meta props, it has to be completely decoupled from redux-form


lowiebenoot:
The `Input` component now implemented `input` and `meta` props, to support redux-forms. But it's very weird to use them when you're not using redux-forms.

We need to change this.

PR #349, opened 2018-09-04
Input: decouple from redux forms

driesd:
### Description

This PR decouples our `Input` component from `redux forms`. Therefore we removed the `input` and `meta` props. Also we added an `error` prop which can accept either a string or an element. The value of the `error` prop will be rendered below the input field just like `meta.error` was before this PR.

Solves #247 

### Breaking changes

We removed the `input` and `meta` props to decouple from redux-forms.


components/input/Input.js
```diff
@@ -17,9 +15,9 @@ import theme from './theme.css';
 
 export default class Input extends PureComponent {
   static getDerivedStateFromProps(nextProps, prevState) {
-    const value = nextProps.input ? nextProps.input.value : nextProps.value;
-    if (value !== undefined) {
-      const newValue = Input.parseValue(Input.getPropsValue(nextProps), nextProps);
+    if (nextProps.value !== undefined) {
```
JorenSaey: Is it on purpose that null is a valid value and undefined is not?

driesd: Yes it's on purpose. It could be that the parent container wants to set the value to null.

timdegroote: what this check is doing is checking wether a value has been supplied by the parent component (i.e. if this is not the case the value of `value` will be `undefined`).

If a parent decides to for whatever reason pass `null` then it means the parent has passed a value and this component shouldn't care about it anymore.

If a value is indeed `undefined` though, the Input component will AFAIK keep track of it in its state etc.

JorenSaey: Thanks for the explanation :)

PR #351, opened 2018-09-05
Merge LinkButton and Button

LennertBlommaert:
### Description

This PR merges the `LinkButton` and `Button` components.

The `LinkButton` component can now be displayed by rendering a `Button` which `level` prop is set to `"link"` (The other levels are: "primary", "secondary, "outline" and "destructive").

The design for a link button also got slightly updated, so it is more in line with the other button types, this has also been implemented. Along with a clearer and more performant way to style the different button levels and their states: the logic of when to apply a different styling is now centralised and is easily overridable thanks to CSS variables.

I left the `LinkButton` component and its styling intact, as I am not sure how to handle the deprecation of a component: how would we do this?

The only thing that should've changed visually is the styling of the `Button` of level `"link"`

screen shot 2018-09-05 at 15 15 24

#### Screenshot after this PR

screen shot 2018-09-05 at 15 18 24


LennertBlommaert:
Closing this as intermediate commits broke the styling in IE11.

PR #352, opened 2018-09-06
[TU93] Datagrid multi-select behaviour

LennertBlommaert:
### Description

The design for multi-selections/bulk actions got updated.
This PR implements these changes.

I roughly copied the implementation already made by team invoicing and made the rendering of the number of selected rows and bulk actions a part of the DataGrid.

screen shot 2018-09-06 at 18 13 43

#### Screenshot after this PR

screen shot 2018-09-06 at 18 14 20

### Breaking changes
None.

### DIFF
--- a/components/datagrid/HeaderRow.js
+++ b/components/datagrid/HeaderRow.js
@@ -12,7 +12,6 @@ class HeaderRow extends PureComponent {
       className,
       checkboxSize,
       children,
-      numSelectedRows,
       sliceFrom,
       sliceTo,
       onSelectionChange,
@@ -24,8 +23,6 @@ class HeaderRow extends PureComponent {
     const childrenArray = Array.isArray(children) ? children : [children];
     const childrenSliced = childrenArray.slice(sliceFrom, sliceTo);
     const classNames = cx(theme['header-row'], className);
-
-    console.log(`${numSelectedRows} selected`);
 
     return (
       <Row backgroundColor="neutral" className={classNames} data-teamleader-ui="datagrid-header-row" {...others}>
@@ -44,16 +41,11 @@ HeaderRow.propTypes = {
   checkboxSize: PropTypes.oneOf(['small', 'medium', 'large']),
   className: PropTypes.string,
   children: PropTypes.any,
-  numSelectedRows: PropTypes.number,
   onSelectionChange: PropTypes.func,
   selectable: PropTypes.bool,
   selected: PropTypes.bool,
   sliceFrom: PropTypes.number,
   sliceTo: PropTypes.number,
-};
-
-HeaderRow.defaultProps = {
-  numSelectedRows: 0,
 };
 
 export default HeaderRow;
--- a/components/datagrid/HeaderRow.js
+++ b/components/datagrid/HeaderRow.js
@@ -29,12 +29,7 @@ class HeaderRow extends PureComponent {
       <Row backgroundColor="neutral" className={classNames} data-teamleader-ui="datagrid-header-row" {...others}>
         {selectable && (
           <HeaderCell align="center" flex="min-width">
-            <Checkbox
-              checked={selected}
-              onChange={onSelectionChange}
-              size={checkboxSize}
-              indeterminate={partiallySelected}
-            />
+            <Checkbox checked={selected} onChange={onSelectionChange} size={checkboxSize} />
           </HeaderCell>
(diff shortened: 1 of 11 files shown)


stories/datagrid.js
```diff
@@ -136,6 +167,17 @@ storiesOf('DataGrids', module)
       comparableId={1}
       onSelectionChange={handleRowSelectionChange}
     >
+      <DataGrid.HeaderRowOverlay>
+        <React.Fragment>
```
duivvv: Is this Fragment needed?

stories/datagrid.js
```diff
@@ -78,6 +98,17 @@ storiesOf('DataGrids', module)
   ))
   .add('with footer', () => (
     <DataGrid selectable={boolean('Selectable', true)} comparableId={1} onSelectionChange={handleRowSelectionChange}>
+      <DataGrid.HeaderRowOverlay>
+        <React.Fragment>
```
duivvv: See below

stories/datagrid.js
LennertBlommaert: No, left behind WIP.

components/datagrid/HeaderRow.js
```diff
@@ -25,6 +25,8 @@ class HeaderRow extends PureComponent {
     const childrenSliced = childrenArray.slice(sliceFrom, sliceTo);
     const classNames = cx(theme['header-row'], className);
 
+    console.log(`${numSelectedRows} selected`);
```
JorenSaey: console.log

components/datagrid/HeaderRowOverlay/NumSelectedRows.js
```diff
@@ -0,0 +1,42 @@
...
+import theme from './theme.css';
+
+class NumSelectedRows extends PureComponent {
+  render() {
+    const { numSelectedRows, numSelectedRowsLabel } = this.props;
+
+    return (
+      <Box
+        display="flex"
+        align-items="center"
```
InstaK: I don't think this works with a dash in TL-UI. Should be camelCased, alignItems, right?

components/datagrid/HeaderRowOverlay/theme.css
```diff
@@ -0,0 +1,16 @@
+@import '@teamleader/ui-colors';
+@import '@teamleader/ui-typography';
+
+.header-row-overlap {
+  background: var(--color-neutral-light);
+  position: absolute;
+  left: 48px;
+  right: 0;
```
InstaK: I think `right: 0;` is redundant if you specify `left`, not 100% sure, but should be the same result.

InstaK: Ok, TIL then 👍

components/datagrid/HeaderRowOverlay/theme.css
```diff
@@ -0,0 +1,12 @@
...
+@import '@teamleader/ui-typography';
+
+.header-row-overlay {
+  background: var(--color-neutral-light);
+  position: absolute;
+  left: 48px;
+  right: 0;
+  height: 48px;
+  overflow-x: auto;
+  z-index: 3;
```
nickwaelkens: Any reason why `3`?

components/datagrid/HeaderRowOverlay/HeaderRowOverlay.js
```diff
@@ -0,0 +1,47 @@
...
+import theme from './theme.css';
+
+class HeaderRowOverlay extends PureComponent {
+  render() {
+    const { children, className, numSelectedRows, numSelectedRowsLabel, ...others } = this.props;
+
+    const classNames = cx(theme['header-row-overlay'], className);
+
+    return (
+      numSelectedRows > 0 && (
```
nickwaelkens: Can we not prevent the `HeaderRowOverlay` component from rendering at a higher level instead of doing the check in here?

components/datagrid/HeaderRowOverlay/theme.css
driesd: This is because, for the DataGrid, z-index 1 and 2 are already in use:

```
.has-blend-left .cell:first-child::before,
.has-blend-right .cell:last-child::before {
  z-index: 2;
}
```

```
.has-border-around {
  z-index: 1;
}
```

PR #357, opened 2018-09-13
IslandGroup: apply box props to the Box wrapper

driesd:
### Description

This PR applies `boxProps` to the `Box` wrapper of our `IslandGroup`.

schermafdruk 2018-09-13 13 14 35

#### Screenshot after this PR

schermafdruk 2018-09-13 13 14 13

### Breaking changes

None.

### DIFF
--- a/components/island/IslandGroup.js
+++ b/components/island/IslandGroup.js
@@ -1,21 +1,22 @@
 import React, { PureComponent } from 'react';
 import PropTypes from 'prop-types';
 import cx from 'classnames';
-import Box from '../box';
+import Box, { pickBoxProps } from '../box';
 import theme from './theme.css';
 import Island from './Island';
 import { elementIsDark } from '../utils/utils';
 
 class IslandGroup extends PureComponent {
   render() {
-    const { children, className, color, dark, size } = this.props;
+    const { children, className, color, dark, size, ...otherProps } = this.props;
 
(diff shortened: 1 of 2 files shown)


components/island/IslandGroup.js
```diff
@@ -1,21 +1,22 @@
...
+    const { children, className, color, dark, size, ...otherProps } = this.props;
 
+    const boxProps = pickBoxProps(otherProps);
     const isDark = elementIsDark(color, dark);
 
     const classNames = cx(theme['segmented'], theme['island'], theme[color], { [theme['dark']]: isDark }, className);
 
     return (
-      <Box className={classNames} padding={0}>
+      <Box {...boxProps} className={classNames} padding={0}>
```
nickwaelkens: Maybe put `{...boxProps}` at the end, you never know if someone wants to override the `padding={0}` default?

driesd: I did this on purpose, because the padding should not be overridable in this component. Second reason for this, is that the flexDirection could also not be overridden.

Is that a valid explanation, or do you still want me to put `{...boxProps}` at the end?

nickwaelkens: Fair enough, can still change it later if necessary.

 :ok: by @nickwaelkens on 13/09/2018 16:00:47

PR #359, opened 2018-09-13
Select: error & help text

driesd:
### Description

This PR implements the `error` & `helpText` prop for the `Select` component, which renders the correct message below the input field.

schermafdruk 2018-09-14 11 51 03

#### Screenshot after this PR

schermafdruk 2018-09-14 11 50 01

schermafdruk 2018-09-14 11 49 32

### Breaking changes

None.

### DIFF
--- a/components/select/Select.js
+++ b/components/select/Select.js
@@ -41,10 +41,16 @@ class Select extends PureComponent {
         ...commonStyles,
         backgroundColor: isDisabled ? colors.TEAL_DARK : colors.TEAL,
         '&:hover': {
-          borderColor: colors.TEAL_LIGHT,
+          borderColor: error ? colors.RUBY_LIGHT : colors.TEAL_LIGHT,
         },
-        borderColor: isFocused ? colors.TEAL_LIGHT : isDisabled ? colors.TEAL_DARK : colors.TEAL,
-        boxShadow: isFocused ? `0 0 0 1px ${colors.TEAL_LIGHT}` : 'none',
+        borderColor: error
+          ? colors.RUBY_LIGHT
+          : isFocused
+            ? colors.TEAL_LIGHT
+            : isDisabled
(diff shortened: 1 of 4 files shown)


components/select/theme.css
```diff
@@ -15,3 +15,17 @@
   border-radius: 0 4px 4px 0;
   min-height: 100%;
 }
+
+.validation-icon {
+  margin-top: -1px;
+}
+
+.has-error {
+  &:not(.is-inverse) .validation-text {
```
InstaK: I guess you can remove the not here if you specify what `is-inverse` color should be underneath. As long as it stays in this order, ofc.

driesd: I could do that indeed, but then you end up with a color override instead of applying the right color at once...

components/select/Select.js
```diff
@@ -276,6 +320,7 @@ class Select extends PureComponent {
           styles={this.getStyles()}
           {...restProps}
         />
+        {error ? this.renderValidationMessage() : this.renderHelpText()}
```
lowiebenoot: Why not just call it `renderError`?

components/select/Select.js
```diff
@@ -258,6 +259,18 @@ class Select extends PureComponent {
     );
   };
 
+  renderHelpText() {
+    const { helpText, inverse } = this.props;
+
+    if (helpText) {
```
JorenSaey: PP: don't do the check here bu in the general render function. That way only one function defines what is rendered and what not

components/select/Select.js
driesd: I called it `renderValidationMessage` because it would be expandable with other message types, like you can see in the design guidelines.

schermafdruk 2018-09-17 09 41 25

But, the other message types are not implemented yet. So, I can change the name to `renderError` for now if you want?

PR #361, opened 2018-09-17
IslandGroup: refactor to solve multiple small bugs

driesd:
### Description

This PR fixes:
* unwanted center alignment of children when `direction` is vertical
* weird `border-radius` & `border-width` behaviour when there's less or more then 3 Islands in an IslandGroup
* redundant `isDark` check (which is already done by `Island` component)
* background color overflow bug with rounded corners
* repetition of the Island component while mapping over children

schermafdruk 2018-09-17 15 24 05

schermafdruk 2018-09-17 15 36 39

#### Screenshot after this PR

schermafdruk 2018-09-17 15 23 09

schermafdruk 2018-09-17 15 22 30

### Breaking changes

None.

### DIFF
--- a/stories/island.js
+++ b/stories/island.js
@@ -4,15 +4,16 @@ import { storiesOf } from '@storybook/react';
 import { checkA11y } from 'storybook-addon-a11y';
 import { withInfo } from '@storybook/addon-info';
 import { withKnobs, boolean, select } from '@storybook/addon-knobs/react';
-import { Island, IslandGroup, TextBody, Heading3 } from '../components';
+import { Island, IslandGroup, Link, TextBody, TextSmall, Heading3 } from '../components';
 import {
   IllustrationInvoices120X120Static,
   IllustrationMeetings120X120Static,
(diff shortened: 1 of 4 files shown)


stories/island.js
```diff
@@ -62,4 +62,25 @@ storiesOf('Island', module)
...
+    <IslandGroup
+      dark={boolean('Dark', false)}
+      color={select('Color', colors, 'neutral')}
+      direction="vertical"
+      size={select('Size', sizes, 'medium')}
+    >
+      <Island>
+        <TextBody>kanye.west@goodmusic.com</TextBody>
+      </Island>
+      <Island display="flex" flexDirection="row" alignItems="center">
```
InstaK: Flex is default row so redundant direction

InstaK:
Hm. Island has by default flexDirection set to 'column', so that's why it needed to be set to 'row'. Thought you added it on an item which was just flexed, because a flexed item is by default 'row', so in this story it actually needs to be `flexDirection="row"` 🙈

PR #343, opened 2018-08-22
Refactor Popover

LennertBlommaert:
### Description

This PR refactors the [`Popover` component](https://components.teamleader.design/?knob-Locale=nl&knob-Size=medium&knob-Color=neutral&knob-Tint=lightest&selectedKind=Popover&selectedStory=vertical&full=0&addons=1&stories=1&panelRight=0&addonPanel=storybooks%2Fstorybook-addon-knobs).

And solves the issue #303: it now renders more accurately, by not blindingly choosing the opposite `direction`, if it would be rendered off-screen in the given one. 
In the last case, it chooses for the direction with the most space and if needed, it becomes vertically scrollable.

### Screenshots

In this example, the `direction` is set to `'south'`.

_When `Popover` fits in viewport, following the given `direction`:_
screen shot 2018-09-03 at 18 01 25

_When `Popover` does not fit in the viewport, following both the given and its opposite `direction` (and there is more space in the current direction):_
screen shot 2018-09-03 at 18 01 37

_When `Popover` does not fit in the viewport, following both the given and its opposite `direction` (and there's more space in the opposite direction):_
screen shot 2018-09-03 at 18 01 51

### Breaking changes

None.

PR #331, opened 2018-08-02
Remove deprecated lifecycles

LennertBlommaert:
### Description

This PR removes and replaces all the deprecated lifecycles `componentWillRecieveProps` and `componentWillUpdate` (there are no `componentWillMount' lifecycles used over the repo).

This PR resolves issue #299, together with PR #330

Nothing has visually changed

#### Screenshot before this PR

Not applicable.

#### Screenshot after this PR

Not applicable.

### Breaking changes

None.

PR #366, opened 2018-09-20
Release 0.14.2

driesd:
### Changed

- `Popover`: renders more accurately, by not blindingly choosing the opposite direction, if it would be rendered off-screen in the given one. In the last case, it chooses for the direction (on the same axis) with the most space and if needed, it becomes scrollable. ([@LennertBlommaert](https://github.com/LennertBlommaert) in [#343](https://github.com/teamleadercrm/ui/pull/331))

### Fixed

- `Input`: fixed a bug that you could not adjust the current value in a number type input field combined with min & max attributes ([@driesd](https://github.com/driesd) in [#365](https://github.com/teamleadercrm/ui/pull/365))

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,8 @@
 
 ### Fixed
 
+- `Input`: fixed a bug that you could not adjust the current value in a number type input field combined with min & max attributes ([@driesd](https://github.com/driesd) in [#365](https://github.com/teamleadercrm/ui/pull/365))
+
 ## [0.14.1] - 2018-09-19
 
 ### Added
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -12,6 +24,8 @@
 
 ### Fixed
 
+- `Input`: fixed a bug that you could not adjust the current value in a number type input field combined with min & max attributes ([@driesd](https://github.com/driesd) in [#365](https://github.com/teamleadercrm/ui/pull/365))
```
InstaK: Q: If you're going to release this, why are you adding it here?

PR #374, opened 2018-09-25
Upgrade to webpack 4

driesd:
### Description

This PR fixes the build failure on production. It contains upgrades for the following npm packages:
* `webpack@4.1.1`
* `extract-text-webpack-plugin@4.0.0-beta.0`
* `html-webpack-plugin@4.0.0-alpha.2`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,14 +1,20 @@
-## [Unreleased]
+## [unreleased]
 
 ### Added
 
 ### Changed
+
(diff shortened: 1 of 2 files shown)


InstaK:
Hm, rather than saying which packages got updated here (which we can see in the files changed), it might be better to explain why this was a problem so we understand why these changes are necessary? Cause bumping these packages sounds okay to me, I don't understand why it fixes it though.

PR #376, opened 2018-09-25
Merge LinkButton into Button

LennertBlommaert:
### Description
This PR merges the LinkButton and Button components.

The `LinkButton` component can now be displayed by rendering a `Button` which `level` prop is set to `"link"` (The other levels are: "primary", "secondary, "outline" and "destructive").

The design for a link button also got slightly updated, so it is more in line with the other button types, this has also been implemented.

I left the `LinkButton` component and its styling intact, as I am not sure how to handle the deprecation of a component, how would we do this?

#### Screenshot of the `LinkButton`
image

#### Screenshot of the `Button` with prop `level` set to `"link"`
image

### Breaking changes
The `LinkButton` has been removed.


components/button/Button.js
```diff
@@ -83,13 +83,14 @@ class Button extends PureComponent {
...
       icon && iconPlacement === 'right' && icon,
-      processing && (
-        <LoadingSpinner
-          className={theme['spinner']}
-          color={this.getSpinnerColor()}
-          size={size === 'small' ? 'small' : 'medium'}
-        />
-      ),
+      processing &&
+        level !== 'link' && (
```
lowiebenoot: Any specific reason why a LinkButton should not be able to have a processing state?

InstaK: Comment to his commit is
> as we do not have an appropriate spinner style for it yet
I'm not sure if that's an issue though? I'd say just allow it to do the style we have now? Or is Stephen radically against this?

LennertBlommaert: At this moment the `LoadingSpinner` comes in only two colours,
`"white"` and `"teal"`
screen shot 2018-09-25 at 17 26 52
screen shot 2018-09-25 at 17 26 52

Normally the spinner takes the same color as the text does, for the `"link"` level this would be `aqua`, adding this color to the `LoadingSpinner` felt something for another PR to me. (the inverse `"link"` can actually take the already available white spinner).

Also the original `"LinkButton"` does not have a processing state (but I also  think it should have one).

lowiebenoot: Go for `a` or `c`, and I would even prefer `a`.

LennertBlommaert: `a` it is.

lowiebenoot:
@LennertBlommaert can you solve the conflicts?

PR #377, opened 2018-09-26
Popover refactor: remove factory pattern

LennertBlommaert:
### Description

This PR refactors the [Popover component](https://components.teamleader.design/?knob-Position=middle&knob-Direction=west&knob-Lock%20scroll=true&knob-Color=neutral&knob-Backdrop=transparent&knob-Locale=nl&knob-Tint=lightest&knob-Offset%20correction=0&knob-Size=medium&selectedKind=Popover&selectedStory=horizontal&full=0&addons=1&stories=1&panelRight=0&addonPanel=storybooks%2Fstorybook-addon-knobs).

As [requested](https://github.com/teamleadercrm/ui/pull/343#issuecomment-419834046) in a [previous refactor](https://github.com/teamleadercrm/ui/pull/343): I removed the factory pattern, as it did not have much use anymore.

Nothing should have visually changed.

### Breaking changes

`PopoverHorizontal` and `PopoverVertical` are removed. Use the new `Popover` instead. The `direction` prop (its possible values are: `"north`, `"south"`, `"east"`, "`west`") determines on which axis the component will be rendered.

### DIFF
--- a/components/popover/positionCalculation.js
+++ b/components/popover/positionCalculation.js
@@ -9,8 +9,8 @@ const getElementPositionValues = element => {
     left,
     right,
     bottom,
-    middle: top + (bottom - top) / 2,
-    center: left + (right - left) / 2,
+    centerY: top + (bottom - top) / 2,
+    centerX: left + (right - left) / 2,
   };
 };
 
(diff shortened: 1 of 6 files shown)


components/popover/positionCalculation.js
```diff
@@ -88,11 +88,11 @@ const getHorizontalDirectionPositionTopValue = ({
   inputOffsetCorrection,
 }) => {
   switch (position) {
-    case 'middle':
-      return anchorPosition.middle - popoverDimensions.height / 2;
-    case 'top':
+    case 'center':
```
JorenSaey: PP: I would work with constants for position and define these somewhere so that is clear what options are possible.

LennertBlommaert: @JorenSaey 	would you suggest something like this?

````
const POSITION_START = 'start';
const POSITION_CENTER = 'center';
const POSITION_END = 'end';
````

PR #379, opened 2018-09-27
Bugfix: avoid using Math.max()

driesd:
### Description

Before this PR we were using `Math.max()` in our `Input` component. Apparently this was breaking the production deploy of our design-components CI. This was caused by Storybooks dependency usage of `babel-preset-minify`.

This PR avoids the usage of `Math.max()` by refactoring the code with an `if/else if` notation.

### Breaking changes

None.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,8 @@
 
 ### Fixed
 
+- `Input`: fixed the failing Storybook build by avoiding the usage of Math.max() ([@driesd](https://github.com/driesd) in [#379](https://github.com/teamleadercrm/ui/pull/379))
+
 ## [0.15.2] - 2018-09-25
 
 ### Added
--- a/components/input/Input.js
+++ b/components/input/Input.js
@@ -45,7 +45,13 @@ export default class Input extends PureComponent {
 
   bindToMinMax(value) {
     const { min, max } = this.props;
-    return Math.min(Math.max(value, min), max);
+
+    if (value > max) {
+      return max;
+    } else if (value < min) {
+      return min;
+    }
+    return value;
   }
 
   state = {


lowiebenoot:
Because that is not correct.

Top level: 

```
  "presets": [
    "@babel/preset-env",
    "@babel/preset-react",
    [
      "minify",
      {
        "builtIns": false
      }
    ]
  ],
```

driesd:
As @lowiebenoot and I found out together, the `.babelrc` overrides (in all the formats mentioned above) simply don't work. We decided to close this PR because working around `Math.max()` works, but is no solid solution.

stivni:
Hi all!

It's important that our storybook builds & deploys succeed again.
Can we first merge this PR and think about a better solution when we have a working build again?
Unless someone has a better alternative at the top of their mind?

Thanks!

lowiebenoot:
Why is it important?

stivni:
I assumed that teamleader.design serves as the documentation of how these components should be use and behave? (If not - why do we have it 🤔)

If so, documentation and code will slowly get out of sync, which will cause a lot of headaches.

Also : broken window syndrome. If a build/deploy break, fix it, before it gets worse, because than it'll become more difficult to fix.

lowiebenoot:
We can still just run it locally.

I just think this is a ridiculous solution, for trying to use a broken alpha version. Where does it stop? Which functions can we use and which functions can't we use. 

Can't we deploy it in dev mode, so the minifying doesn't run?

stivni:
@lowiebenoot you can do whatever you want, but the automatic builds need to be fixed

discuss, dismiss, ... but find a solution you can live with in that case

duivvv:
How is this breaking the CI? What a strange situation :)

lowiebenoot:
In the meanwhile, it isn't anymore. There was an issue with the building of storybook 4 alpha in production mode. But they (kind of) fixed it.

lowiebenoot:
Also hard to discuss/dismiss if people don't answer...

PR #385, opened 2018-10-03
[TU-132] LoadingBar component

LennertBlommaert:
### Description

Say hello to our new `LoadingBar` component, which comes in all your favourite colours and different sizes!

As you will see I started of with a SVG implementation, but ended up using our `Box` component (= `div` tags). 

SVGs are the better choice for animating performance-wise.

But so is animating the transform property (instead of e.g. position), sadly [IE11 can't handle](https://caniuse.com/#search=transform) setting that prop via CSS on SVGs. 

Therefore: `div`s.

Animating the SVG via JS is an option, but I don't feel like that's worth the performance gain. Also, there shouldn't be multiple loading bars animating at a time. (If we really notice performance issues, we will get back to this decision)

In a next PR we will integrate this in the `DataGrid`.

Note that animations are something we're starting to prospect more in the design team, likely the animation will be tweaked in the future, but nothing about implementing it should change.

#### Screenshots

screen shot 2018-10-03 at 15 41 51

### Breaking changes

None.


components/loadingBar/theme.css
```diff
@@ -0,0 +1,335 @@
+@import '@teamleader/ui-animations';
+@import '@teamleader/ui-colors';
+@import '@teamleader/ui-utilities';
+
+@keyframes slideOut {
+  to {
+    transform: translateX(650%);
```
InstaK: PP: translate3d()

components/loadingBar/theme.css
```diff
@@ -0,0 +1,335 @@
...
+  }
+}
+
+.loading-bar {
+  width: 100%;
+
+  will-change: transform;
```
InstaK: Nice 👍

components/loadingBar/theme.css
LennertBlommaert: TIL 👌

components/loadingBar/theme.css
```diff
@@ -0,0 +1,335 @@
...
+
+.is-medium {
+  height: calc(var(--spacer-unit) * 2);
+}
+
+.is-large {
+  height: calc(var(--spacer-unit) * 3);
+}
+
+.is-neutral {
```
Kemosabert: PP: in the Bullet component we are using PostCSS loops to set the different color - tint variations for the classes, saving a bunch of duplicate code and making it easier to make adjustments.

LennertBlommaert: Great idea 👍

components/loadingBar/LoadingBar.js
```diff
@@ -0,0 +1,45 @@
...
+      theme['loading-bar'],
+      theme[`is-${color}`],
+      theme[`is-${size}`],
+      theme[`is-${tint}`],
+      className,
+    );
+
+    return (
+      <Box data-teamleader-ui="loading-bar" className={classNames} {...others}>
+        <Box className={theme['loading-bar-indicator']} />
```
lowiebenoot: It's useless to use a Box for this one.

LennertBlommaert:
@lowiebenoot I do agree to make the animation go faster.
But I'm not sure about using `linear` as the animation curve. I sat down with Stephen and we decided to keep it on `ease` (at least for now).

Note that this is just a first version. As we will do more research into animation and still have to 'battle test' this component (e.g. in the `DataGrid`), the animation will probably be updated.

lowiebenoot:
Alright. It's also just hard to know how it feels like when you just see the raw `LoadingBar`. It probably feels different when you see it in a `DataGrid`.

Will this component actually be used in different places as well? Because I actually just expected it to be added directly inside the `DataGrid` component.

LennertBlommaert:
@lowiebenoot, I entirely agree, when put into context we will see what it looks like.

From a design point of view, the `LoadingBar` can be used in other cases than only the `DataGrid`. Like in this design spec for [page loading](https://app.zeplin.io/project/599af68db2f4978a56dec839/screen/5bb627fd9655661459d401b7).

PR #387, opened 2018-10-04
[TU 138] Box alignItems missing values

LennertBlommaert:
### Description

In this PR `baseline` and `stretch` are added as possible values for the `alignItems` prop.

Nothing should have visually changed.

### Breaking changes

None.


LennertBlommaert:
@lowiebenoot, I think `inherit` could be a use case. But I'm not sure about `initial` as this is already default value.
We should then also add these for all other props.
Any thoughts?

lowiebenoot:
@LennertBlommaert Yeah, I'm not sure if `initial` is necessary in component based architecture with scoped css.
For `inherit` there could be a use case, but maybe we can just add it once we need it.

lowiebenoot:
@nickwaelkens yeah, I feel that way as well actually.

PR #391, opened 2018-10-08
[TU-132] DataGrid add loading state

LennertBlommaert:
### Description

Remember the [`LoadingBar` component](https://components.teamleader.design/?selectedKind=LoadingBar&selectedStory=Basic&full=0&addons=1&stories=1&panelRight=1&addonPanel=storybooks%2Fstorybook-addon-knobs&background=%23ffffff) I added [previously](https://github.com/teamleadercrm/ui/pull/385)?

Well this PR renders it inside the notorious [`DataGrid`](https://components.teamleader.design/?selectedKind=DataGrids&selectedStory=Basic&full=0&addons=1&stories=1&panelRight=1&addonPanel=storybooks%2Fstorybook-addon-knobs&background=%23ffffff).
If the new boolean prop `processing` is true, that is.

Note that this is an intermediate version of the loading state of a `DataGrid`, there will still be iterations coming up. But at least we can already indicate that the `DataGrid` is processing something 👍.

screen shot 2018-10-08 at 14 39 44

#### Screenshot after this PR

screen shot 2018-10-08 at 14 39 31

Due to the `LoadingBar`s transparent nature, the borders that are behind it are visible. This is not a desired effect. To fix this a few changes need to be done to the `LoadingBar`, which felt out of the scope of the PR, expect a follow-up pretty soon.

### Breaking changes

None.

### DIFF
--- a/components/datagrid/DataGrid.js
+++ b/components/datagrid/DataGrid.js
@@ -2,6 +2,7 @@ import React, { PureComponent } from 'react';
 import ReactDOM from 'react-dom';
 import PropTypes from 'prop-types';
 import Box from '../box';
+import LoadingBar from '../loadingBar';
 import HeaderRowOverlay from './HeaderRowOverlay';
 import Cell from './Cell';
 import HeaderCell from './HeaderCell';
@@ -110,7 +111,7 @@ class DataGrid extends PureComponent {
   };
 
   render() {
-    const { checkboxSize, children, className, selectable, stickyFromLeft, stickyFromRight, ...others } = this.props;
+    const { checkboxSize, children, className, processing, selectable, stickyFromLeft, stickyFromRight, ...others } = this.props;
     const { selectedRows } = this.state;
 
     const classNames = cx(theme['data-grid'], className);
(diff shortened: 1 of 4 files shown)


components/datagrid/theme.css
```diff
@@ -93,6 +93,13 @@
   color: var(--color-teal-darkest);
 }
 
+.loading-bar {
+  position: absolute;
+  top: 48px;
+  left: 0;
+  z-index: 3;
```
LennertBlommaert: FYI: This is the same value as the HeaderRowOverlay (this is where the bulk actions etc. are displayed)

components/datagrid/DataGrid.js
```diff
@@ -196,10 +198,12 @@ DataGrid.propTypes = {
   stickyFromLeft: PropTypes.number,
   stickyFromRight: PropTypes.number,
   onSelectionChange: PropTypes.func,
+  processing: PropTypes.bool,
```
lowiebenoot: A bit weird to call this prop `processing`, but it's adding a `LoadingBar`. Any reason to not call it `loading`?

LennertBlommaert: The `Toast` and `Button` both also use `processing` to render their `LoadingSpinner`s.

You could think about it as: the "`DataGrid` is processing your request".

In the end more then only rendering a `LoadingBar` should be happen if `processing` is true (e.g.: like diming the opacity of the body cells).
Although this case `loading` would also still be relevant.

lowiebenoot: Alright, fair enough 👍

PR #392, opened 2018-10-08
Icon: add 'element' prop

driesd:
### Description

This PR implements an `element` prop for our `Icon` component. This way,  when needed, you can override the default `span` element to be rendered.

### Breaking changes

None


lowiebenoot:
This was actually already possible, because the "other" props are destructured into the box.

PR #396, opened 2018-10-09
Select: dropdown arrow restyling

driesd:
### Description

[add description here]

[add screenshot here]

#### Screenshot after this PR

[add screenshot here]

### Breaking changes

None


stephen-tl:
@lowiebenoot thanks for the feedback! Yes, Dries and I came to this version together. Let's ship it like it looks now and discuss if we need to improve in a next iteration.

PR #397, opened 2018-10-09
Select: remove the max-width so it becomes fluid

driesd:
### Description

This PR removes the `max-width` limitation, component will now take the full width of its parent.

schermafdruk 2018-10-09 14 13 03

#### Screenshot after this PR

schermafdruk 2018-10-09 14 12 46

### Breaking changes

None.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,8 @@
 
 ### Removed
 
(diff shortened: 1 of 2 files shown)


driesd:
And also, if you would measure the widest item to set a fixed width on the select, it could break your layout. The max width should always be the width of the parent, which you can set manually.

PR #399, opened 2018-10-10
Dialog: add flex behavior to the inner wrapper

driesd:
### Description

A `Dialog` gets a max-height according to your browser viewport. Before this PR, the content inside the Dialog is cut off when it gets too high. This PR adds some styling to the `inner` wrapper to fix this issue.

schermafdruk 2018-10-10 10 44 38

#### Screenshot after this PR

schermafdruk 2018-10-10 10 43 55

### Breaking changes

None.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,8 @@
 
 ### Fixed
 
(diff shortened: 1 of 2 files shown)


InstaK:
Can you make a new release out of this? Would like to implement it in RB

PR #409, opened 2018-10-12
Move static folder inside stories

driesd:
### Description

This PR moves the `static` folder (only being used in our component stories) inside the `stories` folder.

### Breaking changes

None


InstaK:
@driesd ,	@nickwaelkens 	commented on your PR, could you answer that? I'd also like to know.

PR #410, opened 2018-10-22
DatePicker: fix collapsing when container is resizing

ArnaudWeyts:
### Description

The DatePicker was collapsing when the container was getting resized (or when the container width relied on the DatePicker's width)

image

#### Screenshot after this PR

image

### Breaking changes

None

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,8 @@
 
 ### Fixed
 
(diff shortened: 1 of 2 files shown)


lowiebenoot:
You should add your change to the changelog file as well 😉

ArnaudWeyts:
@lowiebenoot Whoops, good point, on it 😅

PR #411, opened 2018-10-23
Input: Refactor

LennertBlommaert:
### Description

Hi folks!

This PR aims to make our dear beloved [`Input` component](https://components.teamleader.design/?selectedKind=Inputs&selectedStory=Input%20only&full=0&addons=1&stories=1&panelRight=1&addonPanel=storybooks%2Fstorybook-addon-knobs&background=%23ffffff) a little less complex.

I tried to achieve this by creating a dumb `InputBase` component - which is only concerned with its looks (what a narcissist) - and build on top of that the `Input` component which keeps track of its value.

The original `Input` component contained a lot of extra lines specifically used for when the input type was a number.
I extracted these lines and moved them to a new component: `NumericInput` 🎉,
which also is build upon the `InputBase`.

So from now on, if you want an input of the type number: use a `NumericInput`.
As a regular `Input` is unable to render a spinner (you know - the two arrows on the side to change the value).

With this PR I also aim to make the implementation of a textarea component easier.

#### Questions
This PR does not really contain a breaking change to me. As you're still able to generate an `input` that only accepts numbers, using the `Input` component (by setting the prop `type` to `"number"`). It just lost its ability to render a spinner, parse the values to an integer and keep min and max into account.

Am I wrong?

Should I make sure that for an `Input`, `type` can not be `number`(by using `PropTypes.oneOf`)?

### DIFF
--- a/components/input/InputBase.js
+++ b/components/input/InputBase.js
@@ -1,0 +1,171 @@
+import React, { PureComponent } from 'react';
+import PropTypes from 'prop-types';
+import cx from 'classnames';
+import omit from 'lodash.omit';
+import Box, { omitBoxProps, pickBoxProps } from '../box';
+import { ErrorText, HelpText } from '..';
+import theme from './theme.css';
+
+export default class InputBase extends PureComponent {
+  state = {
+    inputHasfocus: false,
+  };
+
+  handleBlur = event => {
+    this.setState({ inputHasFocus: false });
+    if (this.props.onBlur) {
+      this.props.onBlur(event);
+    }
+  };
+
+  handleChange = event => {
+    if (this.props.onChange) {
+      this.props.onChange(event);
+    }
+  };
+
+  handleFocus = event => {
+    this.setState({ inputHasFocus: true });
+    if (this.props.onFocus) {
+      this.props.onFocus(event);
+    }
+  };
+
+  renderOneOrMultipleElements(prop) {
+    if (Array.isArray(prop)) {
+      return prop.map((element, index) => React.cloneElement(element, { key: index }));
+    }
+
+    return prop;
+  }
+
+  renderInput = () => {
+    const { bold, ...otherProps } = this.props;
+
+    const classNames = cx(theme['input'], { [theme['is-bold']]: bold });
+
+    const restProps = omit(omitBoxProps(otherProps), [
+      'className',
+      'connectedLeft',
+      'connectedRight',
+      'helpText',
+      'inverse',
+      'onChange',
+      'prefix',
+      'size',
+      'suffix',
+      'value',
+    ]);
+
+    const props = {
+      className: classNames,
+      onBlur: this.handleBlur,
+      onChange: this.handleChange,
+      onFocus: this.handleFocus,
+      ...restProps,
+    };
+
+    return <input {...props} />;
+  };
+
+  render() {
+    const {
+      className,
+      connectedLeft,
+      connectedRight,
+      disabled,
+      error,
+      helpText,
+      inverse,
+      prefix,
+      size,
+      suffix,
+      readOnly,
+      ...others
+    } = this.props;
+
+    const { inputHasFocus } = this.state;
+
+    const classNames = cx(
+      theme['wrapper'],
+      theme[`is-${size}`],
+      {
+        [theme['has-error']]: error,
+        [theme['has-focus']]: inputHasFocus,
+        [theme['has-connected-left']]: connectedLeft,
+        [theme['has-connected-right']]: connectedRight,
+        [theme['is-inverse']]: inverse,
+        [theme['is-disabled']]: disabled,
+        [theme['is-read-only']]: readOnly,
+      },
+      className,
+    );
+
+    const rest = pickBoxProps(others);
+
+    return (
+      <Box className={classNames} {...rest}>
+        <div className={theme['input-wrapper']}>
+          {connectedLeft}
+          <div className={theme['input-inner-wrapper']}>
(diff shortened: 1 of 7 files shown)


LennertBlommaert:
Thanks @nickwaelkens! And it has been done.

InstaK:
I'm not sure, but I think this *will* announce breaking changes? If you now want to render a number input, you have to change `Input` to `NumericInput`, right? That's a breaking change, because you won't have what you want anymore when you update UI without changing code?

components/input/NumericInput.js
```diff
@@ -0,0 +1,132 @@
...
+    const { inverse, handleOnSpinnerUpClick, handleOnSpinnerdownClick } = this.props;
+    const iconProps = {
+      color: inverse ? 'teal' : 'neutral',
+      element: 'button',
+      tint: inverse ? 'lightest' : 'darkest',
+    };
+
+    return (
+      <div className={theme['spinner']}>
+        <Icon className={theme['spinner-up']} onClick={() => handleOnSpinnerUpClick()} {...iconProps}>
```
InstaK: PP:
```suggestion
        <Icon className={theme['spinner-up']} onClick={handleOnSpinnerUpClick} {...iconProps}>
```

components/input/NumericInput.js
```diff
@@ -0,0 +1,132 @@
...
+          <IconChevronUpSmallOutline />
+        </Icon>
+        <Icon className={theme['spinner-down']} onClick={() => handleOnSpinnerdownClick()} {...iconProps}>
```
InstaK: PP
```suggestion
        <Icon className={theme['spinner-down']} onClick={handleOnSpinnerdownClick} {...iconProps}>
```

InstaK: Also PP that the function name here becomes `handleOnSpinnerDownClick` because `Up` is also capitalized for the other function.

components/input/NumericInput.js
```diff
@@ -0,0 +1,132 @@
...
+      this.setState({ value: newValue });
+    }
+  };
+
+  parseValue = value => this.bindToMinMax(this.toNumber(value));
+
+  toNumber = number => {
+    let float = parseFloat(number);
+    if (isNaN(float) || !isFinite(float)) {
+      float = 0;
```
InstaK: (not saying you wrote this, but:) This is actually not a float 🙈

components/input/NumericInput.js
```diff
@@ -0,0 +1,132 @@
...
+    return float;
+  };
+
+  bindToMinMax = value => {
+    const { min, max } = this.props;
+    return Math.min(Math.max(value, min), max);
+  };
+
+  handleIncreaseValue = () => {
+    this.updateStep(1);
```
InstaK: Q: Shouldn't this step actually be variable? If you now want the NumericInput to increment by 5, shouldn't this be this.props.step?

LennertBlommaert: `step` _is_ a prop, but I see where the confusion comes from. The argument passed to the `updateStep` method here is more of a factor, by which `this.props.step` is multiplied before adding it to the value.

Something like this happens, with `n` being the argument passed to `updateStep`:
```
this.state.value + this.props.step * n
```

LennertBlommaert:
@InstaK you can still create a number input using the `Input` component, but its not certainly recommended. When setting the `type` prop of `Input` to `'number'` an error will be thrown, but it'll still work in a way (just no parsing, stepper or min/max) 🤔

I feel like its somewhere in the grey area, any other opinions on this?

components/input/NumericInput.js
InstaK: Okay, but if you do `this.props.step * n` and `n` is always 1, then `* n` is just redundant, no?

LennertBlommaert: `n` is not always 1, in the `handleDecreaseValue` method it is -1.
So it is useful to toggle between increasing and decreasing the value.

InstaK:
@LennertBlommaert 	IMO that's a breaking change. I get the grey area, but it breaks the expected behaviour so yeah...

components/input/NumericInput.js
InstaK: Aha ok, then I didn't look far enough, thanks for elaborating 👍

components/input/InputBase.js
```diff
@@ -0,0 +1,191 @@
+import React, { PureComponent } from 'react';
+import PropTypes from 'prop-types';
+import cx from 'classnames';
+import omit from 'lodash.omit';
+import Box, { omitBoxProps, pickBoxProps } from '../box';
+import { ErrorText, HelpText } from '..';
```
nickwaelkens: PP: Care to have it explicitly imported? Don't want to (potentially) lose hours again figuring out what's going wrong when `npm link`-ing

components/input/InputBase.js
```diff
@@ -0,0 +1,191 @@
...
+import theme from './theme.css';
+
+export default class InputBase extends PureComponent {
```
nickwaelkens: When do you choose to 
```js
export default class X extends Y {}
``` 
and when to do it as
```js
class X extends Y {}
export default X;
```

LennertBlommaert: Didn't really think about it tbh, the second notation is what we normally use, so I'll follow that convention.

InstaK:
If it only throws a warning then it's not breaking ofc 😬

LennertBlommaert:
I discussed it with @nickwaelkens and @driesd and we'll leave it at breaking.
Due to the fact that warnings will be thrown, the spinner won't render anymore and values won't be parsed.

PR #413, opened 2018-10-24
Import components from their own folders

nickwaelkens:
### Description

These imports from the `index.js` file would break when compiling.

### Breaking changes

None.


mikeverf:
I think it was circular dependencies, you're importing from the components/index, which imports HelpText, which imports index, which imports HelpText....

PR #416, opened 2018-10-24
Proposal: HOC InjectTheme

LennertBlommaert:
### Description

Hi y'all,

This PR is a proposal for a HoC that'd remove quite some boilerplate.

`color`, `tint` and `size` are all props that are repeated throughout our library.

They are all used in the same way: their string values are used to construct another string, with the following signature: `'is-[PROP VALUE]'` (or at least they should've).

This string is then used as a class name to apply the appropriate styling.

The valid values for these props should also be the same over the whole library.

Wrapping the components in the HoC created in this PR, takes care of all of the above.

It would centralise all this logic and enhance consistency.

A concern of mine in this PR are the default props:
in the current implementation these'd be the same for all components.

We could say that from now on the default props of `withTheme` will be default for all components or remove the default props declaration out of `withTheme` and keep declaring them in the components themselves.

**What do you guys think?**

### Breaking changes

None, as this is HoC is something that we'll be using internally.

### DIFF
--- a/components/hoc/withTheme/index.js
+++ b/components/hoc/withTheme/index.js
@@ -1,0 +1,2 @@
+import withTheme from './withTheme';
+export default withTheme;
--- a/components/hoc/withTheme/withTheme.js
+++ b/components/hoc/withTheme/withTheme.js
@@ -1,0 +1,34 @@
+import React, { PureComponent } from 'react';
+import PropTypes from 'prop-types';
+import cx from 'classnames';
+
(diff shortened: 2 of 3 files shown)


lowiebenoot:
What will happen with the occasions where we do `isComponentOfType(Button)`? I guess that will stop working, as it will be of type `WithTheme` (not sure though).

LennertBlommaert:
@lowiebenoot I checked it and the `isComponentOfType()` checks are still valid.

PR #418, opened 2018-10-25
Input: Extract ValidationText

LennertBlommaert:
### Description

In this PR, The code behind displaying the help/error text is extracted in all inputs, into a new component `ValidationText`.
DRY +1.

With this PR I try to make composing inputs easier, with the upcoming component for textareas in the back of my mind.

### Breaking changes

None.


components/datepicker/DatePickerInputRange.js
```diff
@@ -232,11 +231,7 @@ class DatePickerInputRange extends PureComponent {
           {this.renderIcon()}
           {this.renderDayPickerInput()}
         </div>
-        {error ? (
-          <ErrorText inverse={inverse}>{error}</ErrorText>
-        ) : (
-          helpText && <HelpText inverse={inverse}>{helpText}</HelpText>
-        )}
+        <ValidationText error={error} help={helpText} inverse={inverse} />
```
LennertBlommaert: In all inputs there is somewhat a mismatch in naming for the `error` and `helpText` props, which are very similar, making their naming consistent didn't feel part of this PR for me (it would also introduce a breaking change.)

PR #419, opened 2018-10-25
NumericInput: Fix spinner color for inverse input

LennertBlommaert:
### Description

This PR fixes the spinner color of the inverse `NumericInput`.
The visual issue was caused by not passing down the correct prop.

screen shot 2018-10-25 at 19 09 19

#### Screenshot after this PR
screen shot 2018-10-25 at 19 08 48

### Breaking changes

None.


LennertBlommaert:
I don't think I need to put this, in the changelog, as the [PR](https://github.com/teamleadercrm/ui/pull/411) where I introduced this issue, is actually still unreleased.

PR #420, opened 2018-10-26
[TU-118] Component z-indexes

driesd:
### Description

This PR fixes the z-index values of `DatePicker`, `Dialog`, `Menu`, `Popover`, `QTip`, `Select` &`Toast` components according to this [design model](https://app.zeplin.io/project/599af68db2f4978a56dec839/screen/5bd03326767a220982b415f8).

### Breaking changes

None.

### DIFF
--- a/components/datepicker/theme.css
+++ b/components/datepicker/theme.css
@@ -344,7 +344,7 @@
   left: 0;
   margin-top: 16px;
   position: absolute;
-  z-index: 1;
+  z-index: 400;
 
   &::after {
     border: solid var(--color-neutral);
@@ -357,7 +357,7 @@
     left: 10px;
     background: var(--color-white);
     transform: rotate(45deg);
-    z-index: 2;
+    z-index: 401;
   }
 }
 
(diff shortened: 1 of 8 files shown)


components/datepicker/theme.css
```diff
@@ -344,7 +344,7 @@
   left: 0;
   margin-top: 16px;
   position: absolute;
-  z-index: 1;
+  z-index: 400;
```
JorenSaey: Just a quick question: do you have a system to define z-indexes? Am interested in that :p

driesd: Yes we have, Stephen made these [visual guidelines](https://app.zeplin.io/project/599af68db2f4978a56dec839/screen/5bd03326767a220982b415f8).

InstaK: Stephen is still working this out, you can already find an idea of what the system should look like on Zeplin I think. I see he started it here: https://guidelines.teamleader.design/primitives/layers but there'll be more available soon

PR #425, opened 2018-10-29
Constants for colors & tints

driesd:
### Description

This PR adds:

1. **constants** for `colors` & `tints`:
- `COLOR`: an object containing colors & tints keys with their hex values
- `COLORS`: an array with all the color names in lowercase
- `TINTS`: an array with all the tint names in lowercase

2. **functions** to get `colors` & `tints` names without the ones passed as an argument
- `colorsWithout`: a function which returns all the color names without the ones passed as an argument
- `tintsWithout`: a function which returns all the tint names without the ones passed as an argument

3. An **implementation** for the `Badge` component & story to demo how `colorsWithout` should be used.

### Breaking changes

None.

### DIFF
--- a/components/badge/Badge.js
+++ b/components/badge/Badge.js
@@ -1,6 +1,8 @@
 import React, { PureComponent } from 'react';
 import PropTypes from 'prop-types';
+import without from 'lodash.without';
 import Box from '../box';
+import { COLORS } from '../../constants';
 import { TextBody } from '../typography';
 import cx from 'classnames';
 import theme from './theme.css';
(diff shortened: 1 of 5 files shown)


duivvv:
What's the reasoning behind the '...Without' functions?

driesd:
@duivvv This way we don't always have to do this `lodash import` in our components:

```
import without from 'lodash.without';
```

It's a bit similar to the `pickBoxProps` & `omitBoxProps` functions exported by our `Box` component.

PR #426, opened 2018-10-30
InputBase: Extract SingleLineInputBase

LennertBlommaert:
### Description

In a [previous PR](https://github.com/teamleadercrm/ui/pull/411) we already split `InputBase` from `Input` in order to reduce complexity form the `Input` component and enhance compose-ability (on top of this `InputBase` both the `NumericInput` and `Input` are build).

In this PR we even take it a step further. 

The `InputBase` still has some code in there, which has no use to the [`TextArea`](https://github.com/teamleadercrm/ui/pull/427). For the later the `prefix`, `suffix`, `connectedLeft` and `connectedRight` are irrelevant.

So in this PR we similarly split the new `SingleLineInputBase` from the `InputBase`. The `SingleLineInputBase` is build on a now simplified `InputBase` and the `NumericInput` and `Input` both are build upon `SingleLineInputBase`. The responsibility of rendering those elements that are irrelevant to the `TextArea` now moves to the `SingleLineInputBase`.

This allows us to fairly easily create a `TextArea` on top of `InputBase`.

### Breaking changes

None.

### DIFF
--- a/components/input/SingleLineInputBase.js
+++ b/components/input/SingleLineInputBase.js
@@ -13,6 +13,20 @@ class SingleLineInputBase extends PureComponent {
     inputHasfocus: false,
   };
 
+  handleBlur = event => {
+    this.setState({ inputHasfocus: false });
+    if (this.props.onBlur) {
+      this.props.onBlur(event);
+    }
+  };
+
+  handleFocus = event => {
+    this.setState({ inputHasfocus: true });
+    if (this.props.onFocus) {
+      this.props.onFocus(event);
+    }
+  };
+
   renderOneOrMultipleElements(prop) {
     if (Array.isArray(prop)) {
       return prop.map((element, index) => React.cloneElement(element, { key: index }));
@@ -57,6 +71,8 @@ class SingleLineInputBase extends PureComponent {
       disabled,
       error,
       inverse,
+      onBlur: this.handleBlur,
+      onFocus: this.handleFocus,
       readOnly,
       ...omitBoxProps(others),
     };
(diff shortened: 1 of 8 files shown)


components/input/SingleLineInputBase.js
```diff
@@ -0,0 +1,102 @@
...
+            {prefix && <div className={theme['prefix-wrapper']}>{this.renderOneOrMultipleElements(prefix)}</div>}
+            <InputBase
+              onFocus={() => {
+                this.setState({ inputHasfocus: true });
+                onFocus && onFocus();
+              }}
+              onBlur={() => {
+                this.setState({ inputHasfocus: false });
+                onBlur && onBlur();
+              }}
```
InstaK: Small PP:
Might be better to just make these handlers?

stories/input.js
```diff
@@ -14,12 +14,39 @@ import {
...
   TextSmall,
   Tooltip,
+  SingleLineInputBase,
 } from '../components';
 import { IconCalendarSmallOutline, IconInfoBadgedSmallFilled } from '@teamleader/ui-icons';
 
 const colors = ['aqua', 'gold', 'mint', 'neutral', 'ruby', 'teal', 'violet'];
-const sizes = ['small', 'medium', 'large'];
+const elements = [null, 'input', 'textarea'];
+const sizes = [null, 'small', 'medium', 'large'];
```
InstaK: Q: Not really sure why null is being added here? 😬

LennertBlommaert: It is for the storybook `select` knob.
Adding `null` tot those arrays, combined with writing the prop on the component like so:
```
size={select('size', sizes) || undefined}
```
Simplifies the 'Story Source' (the code snippet you can copy-paste)

LennertBlommaert: But has no use anymore, as I removed the story, so removing it too 👍.

lowiebenoot:
Didn't you move a little bit too much to SingleInputLineBase? 
Shouldn't the focus state be in the InputBase?

LennertBlommaert:
@lowiebenoot do you mean the `this.state.inputHasFocus`?
That is only needed in `SingleLineInputBase`.

On a plain `input` we can simply use the `:focus` selector to style.
But in the `SingleLineInputBase` the prefix and suffix are wrapped with the input inside a `div`. This `div` has to mimic the focus styling. But as we can't use `:focus` on a `div`, we toggle a class based on the `inputHasFocus` value, on which we can apply the correct styling on the wrapping div.

lowiebenoot:
@LennertBlommaert alright.
FYI: I recently figured out you can actually have focus on a div, if you give it a tabIndex. But that's not what we want here.

LennertBlommaert:
@lowiebenoot oh TIL, thanks!

PR #443, opened 2018-11-08
[TU-204] Dialog: create DialogBase and update Dialog

LennertBlommaert:
### Description

In this PR I take the first steps towards making the `Dialog`s body scrollable.

Before this PR we had the `Dialog` component, which was actually a big container where you pass down all elements of a dialog down as `children`.

So if you wanted to create a dialog as it is [speced out](https://zpl.io/a7ZEgqM).
You had to pass down the `Banner` for the header a `Box` for the body and a `ButtonGroup` for the footer yourself.

In this PR I rename the `Dialog` component to `DialogBase` and built a new `Dialog` on top of that.

So the `DialogBase` is the empty container you had to use before and with our new `Dialog`(which is built on top of the `DialogBase`) you can more easily create the dialog as it is speced.

Once this PR gets trough, I will follow up with a PR to make the body of the `Dialog` scrollable.

Nothing should have visually changed

### Breaking changes

- `Dialog`: moved the responsibility to render a header and a footer to the component itself, based on the received props
- `Dialog`: the `children` are rendered inside a `Box` with the padding prop set to 4


LennertBlommaert:
@nickwaelkens correct 👍!

src/components/dialog/Dialog.js
```diff
@@ -1,102 +1,76 @@
...
-      onEscKeyDown,
-      onOverlayClick,
-      onOverlayMouseDown,
-      onOverlayMouseMove,
-      onOverlayMouseUp,
-      size,
-    } = this.props;
+    return (
+      <ButtonGroup justifyContent="flex-end" padding={4}>
+        {tertiaryAction && <Link inherit={false} {...tertiaryAction} />}
```
lowiebenoot: Q: when are we supposed to use `<Link />` and when are we supposed to use `<Button level="link" />`?

LennertBlommaert: Using the `<Button level="link" />` does not visually match well next to the other buttons. Due to its transparent background and padding you get the feeling the buttons are not equally spaced. This is not the case with the `Link`.

@stephen-tl thinks that the `<Button level="link" />` only is suited for specific cases, such as the pagination for example.

LennertBlommaert:
@lowiebenoot I don't agree.

With this new `Dialog` we're sure that all `Dialog`s are aligned, which is the goal of a design system.

Also creating a standard `Dialog` now is a matter of filling out the correct props. ± 4 props already get you a basic dialog.

While otherwise (also with the subcomponents flow) you should still provide all the right components and their props yourself, everytime you create dialog. Which is also prone to styling and structural errors.

I [Polaris](https://polaris.shopify.com/components/overlays/modal#navigation) and [Carbon](http://react.carbondesignsystem.com/?selectedKind=Modal&selectedStory=Default&full=0&addons=1&stories=1&panelRight=0&addonPanel=storybooks%2Fstorybook-addon-knobs&knob-Tooltip%20direction%20%28direction%29=bottom&knob-Trigger%20text%20%28triggerText%29=Tooltip%20label), were my inspiration.

stories/dialogs.js
```diff
@@ -21,81 +22,58 @@ storiesOf('Dialogs', module)
...
+        <Button onClick={handleActiveToggle} label="Open a dialog" />
+        <State store={store}>
+          <Dialog
+            headerColor={select('headerColor', COLORS, 'neutral')}
+            headerIcon={passHeaderIcon && <IconWarningBadgedMediumOutline />}
+            headingLevel={number('headingLevel', 3, { min: 2, max: 3 })}
+            onCloseClick={handleActiveToggle}
+            primaryAction={{
+              label: text('primaryAction.label', 'Confirm'),
+              onMouseUp: () => console.log('primaryAction.onMouseUp'),
```
lowiebenoot: Q: any reason why you're using `onMouseUp` and not just `onClick`?

LennertBlommaert: None in particular. You could actually pass down any prop or attribute that you can pass to a `Button`.

But it makes more sense to use `onClick` in this case so I'm going to change it.

LennertBlommaert:
> @lowiebenoot I don't agree.
> 
> With this new `Dialog` we're sure that all `Dialog`s are aligned, which is the goal of a design system.
> 
> Also creating a standard `Dialog` now is a matter of filling out the correct props. ± 4 props already get you a basic dialog.
> 
> While otherwise (also with the subcomponents flow) you should still provide all the right components and their props yourself, everytime you create dialog. Which is also prone to styling and structural errors.
> 
> [Polaris](https://polaris.shopify.com/components/overlays/modal#navigation) and [Carbon](http://react.carbondesignsystem.com/?selectedKind=Modal&selectedStory=Default&full=0&addons=1&stories=1&panelRight=0&addonPanel=storybooks%2Fstorybook-addon-knobs&knob-Tooltip%20direction%20%28direction%29=bottom&knob-Trigger%20text%20%28triggerText%29=Tooltip%20label) were my inspiration for this implementation.

@lowiebenoot what's your opinion on this?

I also think this implementation will become even more relevant when extra custom styling/logic is added to the `Dialog` (such as the scrollable body).

lowiebenoot:
@LennertBlommaert it's indeed much easier to create a default dialog now. I got triggered by the props `primaryAction`, `secondaryAction` and `tertiaryAction`. It doesn't feel very flexible. But on the other hand, we'll probably never need to make a custom dialog. And otherwise I'll slap your ass.

PR #441, opened 2018-11-07
Maintenance: new folder structure

driesd:
### Description

[add description here]

#### Screenshot before this PR

[add screenshot here]

#### Screenshot after this PR

[add screenshot here]

### Breaking changes

- breaking change 1
- breaking change 2
- ...

PR #444, opened 2018-11-08
[TU-205] withTheme: maintenance

LennertBlommaert:
### Description

This PR preforms some general maintenance on the `withTheme` HOC.

It moves it into our [new folder structure](https://github.com/teamleadercrm/ui/pull/441) and makes use of the generally declared `COLORS` and `TINTS` [constants](https://github.com/teamleadercrm/ui/pull/425).


LennertBlommaert:
That's how we like 'm.

PR #449, opened 2018-11-09
Add displayName to all components

driesd:
### Description

Due to Storybook's minification, all our component names are reduced to one or two random characters:
* `<n .../>`
* `<t .../>`
* `<Kn .../>`
* ...

This PR adds a displayName to all components, which solves the above issue.

schermafdruk 2018-11-09 12 06 50

schermafdruk 2018-11-09 12 07 00

#### Screenshot after this PR

schermafdruk 2018-11-09 12 06 30

schermafdruk 2018-11-09 12 07 11

### Breaking changes

None.

### DIFF
--- a/src/components/avatar/Avatar.js
+++ b/src/components/avatar/Avatar.js
@@ -38,4 +38,6 @@ Avatar.defaultProps = {
   size: 'medium',
 };
 
+Avatar.displayName = 'Avatar';
+
 export default Avatar;
--- a/src/components/avatar/AvatarStack.js
+++ b/src/components/avatar/AvatarStack.js
@@ -54,4 +54,6 @@ AvatarStack.defaultProps = {
   size: 'medium',
 };
 
+AvatarStack.displayName = 'AvatarStack';
+
 export default AvatarStack;
--- a/src/components/badge/Badge.js
+++ b/src/components/badge/Badge.js
@@ -106,4 +106,6 @@ Badge.defaultProps = {
   color: 'neutral',
 };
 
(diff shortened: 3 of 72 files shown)


driesd:
@nickwaelkens Unfortunately, `displayName` is not stripped out in production. According to `Dan Abramov`, this is not something they can do. The bundle size difference is *3kb*.

This PR is just a proposal to make out documentation readable again, because `there's no I in Teamleader UI`...

*With displayName*
734.255 bytes

*Without displayName*
731.361 bytes

Any more opinions?

driesd:
Q: Is this 3kb considered `unnecessary clutter`, if you know that it helps all the frontenders, providing them with documentation that actually makes sense again?

lowiebenoot:
Maybe you can just disable the minification?

nickwaelkens:
Oh okay; I was expecting more than 3kB, but maybe-like @lowiebenoot suggests-we have to investigate why things are going wrong in Storybook?

driesd:
> Maybe you can just disable the minification?

@lowiebenoot That is a very good solution!

In webpack.config.js I've added `storybookBaseConfig.optimization.minimize = false;`.

Bundle size stays the same as before. And now I only need to add displayName to the Typography components, because they are wrapped in a factory.

This PR will be closed and a new one will be opened. Many thx!

duivvv:
I just assumed that was not an option. 😕

driesd:
I actually had the same assumption and thought it would result in unwanted side effects...

PR #451, opened 2018-11-09
[TU-193] Avatar multiple shapes

LennertBlommaert:
### Description

In this PR we add (next to the circle) an extra possible shape for the [`Avatar`](https://components.teamleader.design/?selectedKind=Avatars&selectedStory=sizes&full=0&addons=1&stories=1&panelRight=1&addonPanel=storybooks%2Fstorybook-addon-knobs&background=%23ffffff): a rounded square.

A contact should be displayed in a circle, while a company in a rounded square ( for more: [see the specs](https://zpl.io/VxnJEMx)).

Therefore the shape is based on the new `type` prop, which values can be `'company'` or `'contact'`.

When no `type` is supplied the `Avatar` will still be displayed as a circle.

screenshot 2018-11-09 at 14 30 09

#### Screenshot after this PR

screenshot 2018-11-09 at 14 30 17

### Breaking changes

None.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,8 +2,8 @@
 
 ### Added
 
-- `Avatar`: added the `type` prop ([@LennertBlommaert](https://github.com/LennertBlommaert) in [#451](https://github.com/teamleadercrm/ui/pull/451))
-- `Avatar`: added a rounded square as a possible shape, to get this shape set the `type` prop to `'company'` ([@LennertBlommaert](https://github.com/LennertBlommaert) in [#451](https://github.com/teamleadercrm/ui/pull/451))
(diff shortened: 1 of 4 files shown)


LennertBlommaert:
@nickwaelkens:

This is how we currently to it [on mobile](https://github.com/teamleadercrm/mobile-core/blob/38b9e31ad26cfe363c428de914626785551f648f/src/shared/Avatar/Avatar.js#L38)(which might not - sorry guys - always be a great reference), but I get your point. 

Is something like `shape`, which can be `'circle'` or `'roundedSquare'` is preferred?

nickwaelkens:
In my personal opinion, yes. Let's wait for more feedback :)

ArnaudWeyts:
Same, I would prefer a `shape` prop.

PR #459, opened 2018-11-13
Storybook: add a playground story to play around

driesd:
### Description

This PR adds a `Playground` story where we can play around with e.g. z-indexes. This becomes very handy for the styleguide team when combining multiple components to see how they influence each other.

schermafdruk 2018-11-13 14 51 56

### Breaking changes

None, because this is `only in Storybook`, `not in the lib` bundle.

### DIFF
--- a/stories/playground.js
+++ b/stories/playground.js
@@ -1,0 +1,259 @@
+import React from 'react';
+import MomentLocaleUtils, { formatDate, parseDate } from 'react-day-picker/moment';
+import { storiesOf } from '@storybook/react';
+import { boolean } from '@storybook/addon-knobs/react';
+import { DateTime } from 'luxon';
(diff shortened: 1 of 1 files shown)


stories/playground.js
```diff
@@ -0,0 +1,259 @@
...
+const preSelectedDate = DateTime.local()
+  .plus({ days: 3 })
+  .toJSDate();
+
+const TooltippedStatusBullet = Tooltip(StatusBullet);
+
+const MyDatagrid = ({ ...props }) => (
+  <DataGrid
+    comparableId={1}
+    // onSelectionChange={handleRowSelectionChange}
```
LennertBlommaert: PP:
```suggestion
    // onSelectionChange={() => console.log('onSelectionChange')}
```

PR #461, opened 2018-11-13
[TU-153] Dialog: separate footer and body 

LennertBlommaert:
### Description

When the body of a [`Dialog`](https://components.teamleader.design/?selectedKind=Dialogs&selectedStory=Dialog&full=0&addons=1&stories=1&panelRight=1&addonPanel=storybooks%2Fstorybook-addon-knobs&background=%23ffffff) is overflowing, there used to be a weird transition between it and the footer.

But fear no more, as in this PR we render a border on top of the footer, when the body is scrollable (as in the [specs](https://zpl.io/a7ZEgqM)).

You might be wondering: "But hey, if the `setIsBodyOverflowing` only is called on resize, then how can the border be correctly rendered when you initially open the `Dialog`? 🤔"

Well, this is because scrolling on the page is locked when you open the `Dialog`.
This triggers the resize, thus also setting the border correctly when initially opening the `Dialog`.

If this was not the case, I'd have to call `setIsBodyOverflowing` in `componentDidUpdate` (if the `active` prop changed).

_**Q: Do you guys think I still have to add this or is it good as is?**_ 

screenshot 2018-11-13 at 20 19 45

#### Screenshot after this PR

screenshot 2018-11-13 at 20 19 13

### Breaking changes

None.

### DIFF
--- a/src/components/dialog/Dialog.js
+++ b/src/components/dialog/Dialog.js
@@ -18,7 +18,8 @@ class Dialog extends PureComponent {
   };
 
   componentDidUpdate(prevProps) {
-    if (this.props.active !== prevProps.active) {
+    if (this.props.active && !prevProps.active) {
+      console.log('Trigger check from CDU');
       this.setIsBodyOverflowing();
     }
   }
--- a/src/components/dialog/Dialog.js
+++ b/src/components/dialog/Dialog.js
@@ -1,14 +1,22 @@
-import React, { PureComponent } from 'react';
+import React, { PureComponent, createRef } from 'react';
 import PropTypes from 'prop-types';
 import omit from 'lodash.omit';
 import cx from 'classnames';
+import ReactResizeDetector from 'react-resize-detector';
 
 import theme from './theme.css';
 
 import { Banner, Box, Button, ButtonGroup, DialogBase, Heading2, Heading3, Link } from '../../index';
 import { COLORS } from '../../constants';
+import { isElementOverflowingY } from '../utils/utils';
 
 class Dialog extends PureComponent {
+  bodyRef = createRef();
+
+  state = {
+    isBodyOverFlowing: false,
+  };
+
   getHeader = () => {
     const { headerColor, headerIcon, headingLevel, onCloseClick, title } = this.props;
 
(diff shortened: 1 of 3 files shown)


lowiebenoot:
It will only trigger a resize, if your page was scrollable before opening the dialog.

LennertBlommaert:
@lowiebenoot I also thought that. I tested it out when the page was not scrollable, yet opening the dialog still triggered the resize detector 🤔.

Maybe I should just add the triggering of `setIsBodyOverflowing` in `componentDidUpdate` to be sure?

lowiebenoot:
@LennertBlommaert yes, I would do it that way.

src/components/dialog/Dialog.js
```diff
@@ -21,16 +29,29 @@ class Dialog extends PureComponent {
...
       </ButtonGroup>
     );
   };
 
+  setIsBodyOverflowing = () => {
+    if (isElementOverflowingY(this.bodyRef.current.node)) {
+      this.setState({ isBodyOverFlowing: true });
+    } else {
+      this.setState({ isBodyOverFlowing: false });
+    }
```
nickwaelkens: PP:
```jsx
setIsBodyOverflowing = () => {
	this.setState({ isBodyOverflowing: isElementOverflowingY(this.bodyRef.current.node) });
};
```

src/components/dialog/Dialog.js
```diff
@@ -47,10 +68,11 @@ class Dialog extends PureComponent {
     return (
       <DialogBase className={classNames} {...restProps}>
         {title && this.getHeader()}
-        <Box className={theme['dialog-body']} padding={4}>
+        <Box className={theme['dialog-body']} padding={4} ref={this.bodyRef}>
```
kevinwaelkens: @nickwaelkens Apparently for Lennert createRef works 🙃

nickwaelkens: The better front-ender 🙈

kevinwaelkens: Teach us

LennertBlommaert: Eat lots of green vegetables and stay hydrated.

src/components/dialog/Dialog.js
```diff
@@ -17,6 +17,12 @@ class Dialog extends PureComponent {
     isBodyOverFlowing: false,
   };
 
+  componentDidUpdate(prevProps) {
+    if (this.props.active !== prevProps.active) {
```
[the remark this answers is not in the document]
LennertBlommaert: Will do, it would spare a redundant `isElementOverflowingY` calculation.

src/components/dialog/Dialog.js
LennertBlommaert: Only if you start bets on shaving it

src/components/dialog/Dialog.js
```diff
@@ -18,7 +18,8 @@ class Dialog extends PureComponent {
   };
 
   componentDidUpdate(prevProps) {
-    if (this.props.active !== prevProps.active) {
+    if (this.props.active && !prevProps.active) {
+      console.log('Trigger check from CDU');
```
lowiebenoot: 💩

LennertBlommaert: 🤦‍♂️

PR #465, opened 2018-11-15
DatePicker: don't change date when a disabled date was selected

ArnaudWeyts:
### Description

When a disabled day is selected (by typing for example), you don't want run the handler.
see: http://react-day-picker.js.org/examples/disabled-interaction

### Breaking changes

- None.

### DIFF
--- a/src/components/datepicker/DatePickerInput.js
+++ b/src/components/datepicker/DatePickerInput.js
@@ -36,8 +36,11 @@ class DatePickerInput extends PureComponent {
     this.setState({ inputHasFocus: true }, () => this.props.onFocus && this.props.onFocus());
   };
 
-  handleInputDateChange = date => {
-    this.setState({ selectedDate: date }, () => this.props.onChange(date));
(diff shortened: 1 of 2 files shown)


src/components/datepicker/DatePickerInput.js
```diff
@@ -36,8 +36,11 @@ class DatePickerInput extends PureComponent {
     this.setState({ inputHasFocus: true }, () => this.props.onFocus && this.props.onFocus());
   };
 
-  handleInputDateChange = date => {
-    this.setState({ selectedDate: date }, () => this.props.onChange(date));
+  handleInputDateChange = (date, modifiers = {}) => {
+    if (modifiers.disabled) {
```
mikeverf: Q: Why not just an inverse check and put it in the if?

ArnaudWeyts: I like to keep my edge cases separate, and I suppose there can be other modifiers than can influence the date being called or not as well.

PR #467, opened 2018-11-16
[TU-224] Popover: work around default stretching

LennertBlommaert:
### Description

Up in this PR: the [`Popover`](https://components.teamleader.design/?selectedKind=Popover&selectedStory=Basic&full=0&addons=1&stories=1&panelRight=1&addonPanel=storybooks%2Fstorybook-addon-knobs&background=%23ffffff)!

The `Popover` stretches itself (in height) according to its contents.
It only stopped stretching until it filled all the available space from its anchor point to the edge of the viewport. This made the `Popover` less clear in some cases.

In this PR we keep this default stretching behaviour, but introduce the `fullHeight` prop. Setting this prop to false, will set the height of the `Popover` to a set - to keep things consistent - value.

The `Popover` also squashes itself when its contents would overflow the edge of the viewport. This behaviour is preserved. When the set height would cause the `Popover` to overflow, the `Popover` skips this set value and also starts squashing.

screenshot 2018-11-19 at 11 08 46

#### Screenshot after this PR

screenshot 2018-11-19 at 11 08 58

### Breaking changes

None.

### DIFF
--- a/src/components/popover/Popover.js
+++ b/src/components/popover/Popover.js
@@ -13,7 +13,7 @@ import theme from './theme.css';
 class Popover extends PureComponent {
   popoverRoot = document.createElement('div');
 
-  state = { positioning: { left: 0, top: 0, arrowLeft: 0, arrowTop: 0, maxPopoverHeight: 'initial' } };
+  state = { positioning: { left: 0, top: 0, arrowLeft: 0, arrowTop: 0, maxHeight: 'initial' } };
 
   componentDidMount() {
     document.body.appendChild(this.popoverRoot);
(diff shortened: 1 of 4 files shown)


src/components/popover/Popover.js
```diff
@@ -169,7 +177,7 @@ Popover.defaultProps = {
   active: true,
   backdrop: 'dark',
   direction: 'south',
-  fullHeight: false,
+  fullHeight: true,
```
LennertBlommaert: `fullHeight` is `true by default, as this has always been its standard behaviour.

PR #471, opened 2018-11-19
Remove blue ouline when clicking on a Link with element='button'

driesd:
### Description

This PR **removes** the `blue outline` when clicking on a `Link` with `element='button'`.

[add screenshot here]

#### Screenshot after this PR

[add screenshot here]

### Breaking changes

None.

### DIFF
--- a/src/components/link/theme.css
+++ b/src/components/link/theme.css
@@ -19,6 +19,10 @@
     margin-left: var(--spacer-smallest);
   }
 
+  &:active {
+    outline: 0;
+  }
+
   &:not(.inherit) {
     color: var(--color-aqua-dark);
     font-family: var(--font-family-regular);


driesd:
Ok @lowiebenoot I'll check with @stephen-tl how we can enhance the focus state. Is it ok to do that in a next PR so we could ship this one already?

nickwaelkens:
Does this mean we have no visual indication when use tab to navigate around the page then? IMO still need "something", not the blue indicator per se 👍

driesd:
@nickwaelkens If you tab (focus state), the blue indicator is still there. I only disabled this blue indicator when the user `clicks` (active state) on the link.

stephen-tl:
> Ok @lowiebenoot I'll check with @stephen-tl how we can enhance the focus state. Is it ok to do that in a next PR so we could ship this one already?

Imo, we should keep de outline focus behaviour as is. The blue outlines is the default browser setting and most clear for a user tabbing through the interface

PR #476, opened 2018-11-28
[TU-238] Select: add possibility to create a new option

driesd:
### Description

This PR adds the possibility to create a new option that is not in the option list.

schermafdruk 2018-11-28 15 17 52

#### Screenshot after this PR

schermafdruk 2018-11-28 15 18 09

### Breaking changes

None.


driesd:
@kevinwaelkens I think, what you mean, is the `onCreateOption` prop. You can always check the [documentation](https://react-select.com/props#creatable-props) for more information about this prop.

PR #477, opened 2018-11-29
Feature: ScrollContainer

LennertBlommaert:
### Description

Howdy!

In this PR we welcome a new component: the `ScrollContainer` 🎉!

It has no styling at all.

You can use it to create components that have a sticky header and footer, whilst the body is scrollable.

We will use it for the `Popover` for example, in: #478.

#### Screenshot

screenshot 2018-11-29 at 14 07 40

### Breaking changes

None.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ### Added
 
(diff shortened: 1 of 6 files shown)


LennertBlommaert:
Never heard of it myself

PR #483, opened 2018-12-25
DataGrid: Indicate a column can be sorted by passing false as the sorted prop

ArnaudWeyts:
### Description

This PR allows to pass `false` to the sorted prop on the `HeaderCell`. This way we can indicate a field is sortable by showing the icon for it.

##### Not sorted on anything

screenshot 2018-12-25 at 11 26 40

##### Sorted on a column

screenshot 2018-12-25 at 11 27 33

##### Not sorted on anything

screenshot 2018-12-25 at 11 28 28

##### Sorted on a single field

screenshot 2018-12-25 at 11 29 08

### Breaking changes

- None.

### DIFF
--- a/src/components/datagrid/theme.css
+++ b/src/components/datagrid/theme.css
@@ -114,15 +114,21 @@
 .is-sortable {
   svg {
     margin-left: var(--spacer-smallest);
+    color: var(--color-teal-light);
   }
 
   &:hover {
     color: var(--color-teal-darkest);
     cursor: pointer;
+
+    & svg {
+      color: var(--color-teal-darkest);
+    }
   }
 }
 
-.is-sorted {
+.is-sorted,
+.is-sorted svg {
   color: var(--color-teal-darkest);
 }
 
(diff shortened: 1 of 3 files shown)


src/components/datagrid/theme.css
```diff
@@ -114,15 +114,21 @@
 .is-sortable {
   svg {
     margin-left: var(--spacer-smallest);
+    color: var(--color-teal-light);
```
LennertBlommaert: Use a neutral color for the icon in an unsorted state (so it matches the text of the header cell), also use a lighter tint (than the text, which is in the 'darkest' tint) to indicate the sorting is inactive.

```suggestion
    color: var(--color-neutral-dark);
```

ArnaudWeyts:
Just discovered the cells are not showing the ellipsis when cells are overflowing anymore because of this PR. 🙈

kevinwaelkens:
🚨 	Also: conflicts

PR #485, opened 2018-12-31
[TU-235] Popover: Use ReactResizeDetector throttling

LennertBlommaert:
### Description

The `ReactResizeDetector` manages throttling itself internally.
This PR moves the throttling responsibility to said component.

### Breaking changes

None.

### DIFF
--- a/src/components/popover/Popover.js
+++ b/src/components/popover/Popover.js
@@ -20,13 +20,11 @@ class Popover extends PureComponent {
 
   componentDidMount() {
     document.body.appendChild(this.popoverRoot);
-
-    events.addEventsToWindow({ resize: this.setPlacementThrottled, scroll: this.setPlacementThrottled });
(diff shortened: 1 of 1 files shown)


src/components/popover/Popover.js
```diff
@@ -20,13 +20,11 @@ class Popover extends PureComponent {
 
   componentDidMount() {
     document.body.appendChild(this.popoverRoot);
-
-    events.addEventsToWindow({ resize: this.setPlacementThrottled, scroll: this.setPlacementThrottled });
+    events.addEventsToWindow({ scroll: this.setPlacementThrottled });
```
lowiebenoot: Do yo still need to add the scroll event? Because it seems to be handled by the `ReactResizeDetector as well`?

LennertBlommaert: Yes we still do need this. I added the onScroll prop to the `ReactResizeDetector` as a quick test and just forgot to remove it again (since it has no effect).

PR #488, opened 2019-01-03
[TU-227] DataGrid: use Array.isArray instead of isArray from node

driesd:
### Description

This PR replaces the deprecated node `isArray` function by the official `Array.isArray` function.

### Breaking changes

None

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,8 @@
 
 ### Changed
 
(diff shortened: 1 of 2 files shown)


driesd:
Yeah it did work, but I don't understand how either 🤔

duivvv:
You probably had the util package installed via a dependency https://www.npmjs.com/package/util

driesd:
@duivvv I already checked package.json because I thought the exact same as are, but no cigar...

PR #490, opened 2019-01-03
[TU-261] Popover: fix cross browser height

LennertBlommaert:
### Description

This PR aims to fix a bug in IE11(/10), where the `Popover`s body collapses.

The solution currently applied is based upon: a [flexbugs workaround](https://github.com/philipwalton/flexbugs#workaround-2) and [this JSFiddle](https://jsfiddle.net/gktpsyv5/).

🚨 **Disclaimer**
The current solution works, but only partially. The `Popover` does not stretch itself fully anymore, once it has been squashed first (by resizing the window, in IE only). This is an edge case, but we can still put effort in fixing it.

This is in relation with the dimensions of the body's content (the less height, the less the `Popover` stretches again.) Also the `max-height` keeps being set correctly.

screenshot 2019-01-02 at 17 53 33

### DIFF
--- a/src/components/popover/Popover.js
+++ b/src/components/popover/Popover.js
@@ -44,15 +44,15 @@ class Popover extends PureComponent {
     }
   };
 
-  getMaxHeight = () => {
+  getHeight = () => {
     const { fullHeight } = this.props;
-    const { maxHeight } = this.state.positioning;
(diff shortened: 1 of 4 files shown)


LennertBlommaert:
Closing this and applying current fix in a new and clean branch :+1:.

PR #494, opened 2019-01-04
[TU-191] Input: directly provide width

driesd:
### Description

This PR makes it possible to override the default full width of an `single line input field` with a custom `value + unit`. I've implemented the `width` prop. Not implemented for our Textarea component for now.

schermafdruk 2019-01-04 14 57 54

schermafdruk 2019-01-04 14 58 19

#### Screenshot after this PR

schermafdruk 2019-01-04 14 44 21

schermafdruk 2019-01-04 14 46 09

### Breaking changes

None.

### DIFF
--- a/src/components/input/SingleLineInputBase.js
+++ b/src/components/input/SingleLineInputBase.js
@@ -82,10 +82,7 @@ class SingleLineInputBase extends PureComponent {
       <Box className={classNames} {...boxProps}>
         <div className={theme['input-wrapper']}>
           {connectedLeft}
-          <div
-            className={theme['input-inner-wrapper']}
-            style={{ width: width && `${width}px`, flex: width && '0 0 auto' }}
-          >
+          <div className={theme['input-inner-wrapper']} style={{ width, flex: width && '0 0 auto' }}>
             {prefix && <div className={theme['prefix-wrapper']}>{this.renderOneOrMultipleElements(prefix)}</div>}
             <InputBase {...inputProps} />
             {suffix && <div className={theme['suffix-wrapper']}>{this.renderOneOrMultipleElements(suffix)}</div>}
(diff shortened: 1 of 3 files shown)


duivvv:
Are there other examples of `width` props and what units they use?

driesd:
Yeah, maybe we should convert `width` from number to `string` and leave the unit for the frontender to decide?

Currently, the only width related props we're already using in other components is the `fullWidth` boolean prop. Used in `Button` & `Banner` components. So no inconsistencies whatsoever.

src/components/input/SingleLineInputBase.js
```diff
@@ -81,7 +82,10 @@ class SingleLineInputBase extends PureComponent {
       <Box className={classNames} {...boxProps}>
         <div className={theme['input-wrapper']}>
           {connectedLeft}
-          <div className={theme['input-inner-wrapper']}>
+          <div
+            className={theme['input-inner-wrapper']}
+            style={{ width: width && `${width}px`, flex: width && '0 0 auto' }}
```
lowiebenoot: What happens if width is false? Won't you get something like `style={{ width: false, flex: false }}` then?

duivvv: Nice one.

duivvv:
This https://repl.it/@duivvv/valid-CSS-units is the value is a string + true if it is numeric

src/components/input/SingleLineInputBase.js
nickwaelkens: Actually, will it not just be evaluated as something that's not supported and "just work" because browsers are awesome?

lowiebenoot: Probably, but it still makes me cringe.

nickwaelkens: Totally, not saying we should ship it like this :)

LennertBlommaert: But `width` will only be `false`, if you assign it that value, right?

Which seems to come close to setting `width` to `false` in CSS yourself, you wouldn't do both things deliberately.

Also in the prop types it will be defined that a boolean is invalid for `width`, so you'd also get an error in the console if you assign such a value.

lowiebenoot: ah, you're right 👍

duivvv: Still, should width be always be passed to flex? I would keep width for width and use a flex prop

PR #500, opened 2019-01-15
[TU-256] QTip: overlay event handler props & closed becomes active

driesd:
### Description

This PR renames the `closed` prop to `active` and inverts the according logic & styles. It also implements the missing `overlay event handler props` to be consistent with Popover & Dialog.

#### Screenshot of the QTip

schermafdruk 2019-01-15 09 19 08

### Breaking changes

- `closed` prop has been renamed to `active`, logic & styling have been inverted
- `onEscKeyDown` prop has to be explicitly passed instead of reusing the `onChange` prop

### DIFF
--- a/src/components/qTip/theme.css
+++ b/src/components/qTip/theme.css
@@ -11,7 +11,7 @@
   overflow: hidden;
   pointer-events: none;
 
-  &:not(.is-closed) {
+  &:not(.is-active) {
     z-index: 600;
   }
 }
@@ -23,11 +23,12 @@
   border-right: 0;
   border-top-left-radius: 4px;
   border-bottom-left-radius: 4px;
+  transform: translateX(100%) translateX(calc(-1 * var(--tip-width) + 1px)); /* split into 2 separete translate, otherwise IE can't handle it */
   transition: all var(--animation-duration) var(--animation-curve-default);
   pointer-events: all;
 
-  .is-closed & {
-    transform: translateX(100%) translateX(calc(-1 * var(--tip-width) + 1px)); /* split into 2 separete translate, otherwise IE can't handle it */
+  .is-active & {
+    transform: translateX(0);
   }
 }
 
--- a/src/components/qTip/theme.css
+++ b/src/components/qTip/theme.css
@@ -23,7 +23,7 @@
   border-right: 0;
   border-top-left-radius: 4px;
   border-bottom-left-radius: 4px;
-  transform: translateX(100%) translateX(calc(-1 * var(--tip-width) + 1px)); /* split into 2 separete translate, otherwise IE can't handle it */
+  transform: translate3d(100%, 0, 0) translate3d(calc(-1 * var(--tip-width) + 1px), 0, 0); /* split into 2 separete translate, otherwise IE can't handle it */
   transition: all var(--animation-duration) var(--animation-curve-default);
   pointer-events: all;
 
--- a/src/components/qTip/theme.css
+++ b/src/components/qTip/theme.css
@@ -28,7 +28,7 @@
   pointer-events: all;
 
   .is-active & {
-    transform: translateX(0);
+    transform: translate3d(0, 0, 0);
   }
 }
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,8 +3,12 @@
 ### Added
 
 - `SingleLineInputBase`: added the `width` prop to override its full width default behavior. ([@driesd](https://github.com/driesd) in [#494](https://github.com/teamleadercrm/ui/pull/494))
+- `QTip`: added the `onEscKeyDown`, `onOverlayClick`, `onOverlayMouseDown`, `onOverlayMouseMove` & `onOverlayMouseUp` props. ([@driesd](https://github.com/driesd) in [#500](https://github.com/teamleadercrm/ui/pull/500))
 
 ### Changed
+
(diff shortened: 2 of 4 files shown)


src/components/qTip/theme.css
```diff
@@ -23,11 +23,12 @@
   border-right: 0;
   border-top-left-radius: 4px;
   border-bottom-left-radius: 4px;
+  transform: translateX(100%) translateX(calc(-1 * var(--tip-width) + 1px)); /* split into 2 separete translate, otherwise IE can't handle it */
```
kevinwaelkens: PP: translate3d is slightly more performant:
```suggestion
  transform: translate3d(100%, 0, 0) translate3d(calc(-1 * var(--tip-width) + 1px), 0, 0); /* split into 2 separete translate, otherwise IE can't handle it */
```

src/components/qTip/theme.css
```diff
@@ -23,11 +23,12 @@
...
   transition: all var(--animation-duration) var(--animation-curve-default);
   pointer-events: all;
 
-  .is-closed & {
-    transform: translateX(100%) translateX(calc(-1 * var(--tip-width) + 1px)); /* split into 2 separete translate, otherwise IE can't handle it */
+  .is-active & {
+    transform: translateX(0);
```
kevinwaelkens: But then ofc you also need to change this
```suggestion
    transform: translate3d(0, 0, 0);
```

src/components/qTip/theme.css
duivvv: If we're worried about performance here (which I'm not convinced we need to here). 

I would prefer the will-change property over the 3d hack. 
It makes the actually translate easier to read and achieves the same result.

https://developer.mozilla.org/en-US/docs/Web/CSS/will-change

But again, as in the MDN docs.

> Important: will-change is intended to be used as a last resort, in order to try to deal with existing performance problems. It should not be used to anticipate performance problems.

src/components/qTip/theme.css
duivvv: see comment above

src/components/qTip/theme.css
kevinwaelkens: I think it's weird that they say `It should not be used to anticipate performance problems.` and don't give the possibility to read up on why it shouldn't be used that way. To me it seems, anything that can anticipate performance problems should be implemented?

If I take an extreme comparison to explain why I feel that way:
> It's not a problem to do 3 for loops in each other here - once we notice there are performance problems, then we should tackle it 👍 	

Right? 😛

duivvv: > To me it seems, anything that can anticipate performance problems should be implemented?

No cause that would cripple the speed at which you develop things, that would make PRs a constant battle of 'write it like this, cause this is more performant' which to me doesn't make sense and wastes a lot of time in a lot of cases.

> It's not a problem to do 3 for loops in each other here - once we notice there are performance problems, then we should tackle it 👍

I can't see anything wrong with 3 loops in each other when you're not anticipating a lot of data.

Everything depends on the context, as I meant with "should we be worried about performance here?". In the context of what we're doing now, it doesn't make sense to think about making a single small open and close animation ops/sec faster (cause that's what the translateX -> translate3d will achieve), while losing a bit of readability in the process).

Or in short, "premature optimization is the root of all evil".

duivvv: But I see we merged it anyways 🤷‍♂️

kevinwaelkens: I can agree with what you're saying, but I think that should not be a rule, more like a guideline... If it would be a big hassle for a small performance optimization then I would also just leave it. Otherwise I honestly can't understand why you would vouch for something less performant 🤔 	.

kevinwaelkens: > But I see we merged it anyways 🤷‍♂️
Yeah idd 🤔 	...

duivvv: So you think using something like `translate3d` and explaining other devs later in the project that it's not actually 3d but a trick to make it faster for no performance gain in this context is worth it?

kevinwaelkens: I do in fact, if I got that explanation I would react with: "Oh okay, good to know 👍 	" 🤷‍♂️ 	And definitely worth it, because hopefully you only have to tell this person once...  Even in the smallest of projects I would use translate3d because you might as well just make a good habit out of using the most performant possibility there is, no?

duivvv: I really strongly disagree with that statement because of the reasons above. 

I really don't get it :)

kevinwaelkens: As always, if there's a better idea about something / a smarter or better approach for something, I'm always open to change my mind, but, just as you aren't, I'm not convinced yet 😬

duivvv: Don't (prematurely) optimize for things that are not applicable or not perceived by the user. That's not an opinion, that's just common sense.

I think I outlined 'the smart or better approach' a couple of times in the previous comments. Optimize for **real cases**, **consider the context** and **keep the readability in mind**.

I don't have anything more to add to this discussion.

duivvv: Also, please don't take my word on this topic. There are at least 1000 blog posts on premature optimization and why it's harmful. 

including this one. 
https://cdb.reacttraining.com/react-inline-functions-and-performance-bdff784f5578

PR #508, opened 2019-01-17
[TU-270] Popover: fix flexbox behavior

driesd:
### Description

This PR fixes the buggy flexbox behavior in our `Popover` component.

atwhi0og

#### Screenshot after this PR

atwhi0og_2

### Breaking changes

None.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,7 @@
 ### Fixed
 
 - `DatePickerInput`: merged the passed inputProps with the internal inputProps. ([@duivvv](https://github.com/duivvv) in [#503](https://github.com/teamleadercrm/ui/pull/503))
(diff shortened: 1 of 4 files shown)


kevinwaelkens:
Just noticed the second screenshot is photoshopped? 😂 	

Could you please verify that there is no overflowing if you add a fixed height to the popover? I used to be able to pass down a width and a height to the popover and then the body would not overflow, currently it is doing that though (or could you tell me another way how to achieve that)

https://www.dropbox.com/s/tu823hn71z2oq9s/Screenshot%202019-01-17%2011.02.09.jpg?dl=0

driesd:
> Could you please verify that there is no overflowing if you add a fixed height to the popover? I used to be able to pass down a width and a height to the popover and then the body would not overflow, currently it is doing that though (or could you tell me another way how to achieve that)

Verified! Behaves as expected!

PR #515, opened 2019-01-25
Use a div as the element for the toast label/children wrapper

lowiebenoot:
Changed the element for the `TextBody` that is wrapping the Toast content, because if you use some custom content in a Toast, React might complain that you shouldn't use (for example) a `<p>` inside a `<p>`.

```
Warning: validateDOMNesting(...): <p> cannot appear as a descendant of <p>.
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,8 @@
 
 ### Changed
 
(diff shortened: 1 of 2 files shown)


kevinwaelkens:
🚢 	?

PR #523, opened 2019-02-01
Box: check if flexGrow, flexShrink & order are a numbers

driesd:
### Description

Before this PR, when passing `flexGrow`, `flexShrink` & `order` with value `0`, these styles wouldn't be applied to the Box component. This is because `0 is falsy`, so I check on `!isNan()` instead.

### Breaking changes

None.


driesd:
If you pass flexGrow={0}, flexShrink={0} or order={0}, you'd expect these styles actually getting applied. Before this PR, these styles didn't get applied because 0 is falsy.

I agree in most cases you could use the `flex` shorthand instead of `flexGrow={0}` or `flexShrink={0}`. But It should be possible to override and thus, solve this bug.

Also for order={0} there's no shorthand. Although `0` is a valid `order` value.

PR #527, opened 2019-02-14
Fix: AsyncSelect refetch options when handler changed

mikeverf:
### Description

This PR fixes an issue where options aren't refetched when loadOptions handler has changed. This results in old options unexpectedly being shown instead of new ones.

### Breaking changes
none

### DIFF
--- a/src/components/select/AsyncSelect.js
+++ b/src/components/select/AsyncSelect.js
@@ -18,7 +18,7 @@ class AsyncSelect extends PureComponent {
   }
 
   componentDidUpdate(prevProps) {
-    if (this.props.pageSize !== prevProps.pageSize) {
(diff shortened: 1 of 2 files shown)


src/components/select/AsyncSelect.js
```diff
@@ -18,7 +18,7 @@ class AsyncSelect extends PureComponent {
   }
 
   componentDidUpdate(prevProps) {
-    if (this.props.pageSize !== prevProps.pageSize) {
+    if (this.props.pageSize !== prevProps.pageSize || this.props.loadOptions !== prevProps.loadOptions) {
```
duivvv: I think this is the right way to do it, we need to reset the pagination as well.

(can you confirm @driesd)

PR #534, opened 2019-02-21
Z indexes reviewed

driesd:
### Description

This PR fixes several `z-index` problems across multiple components.

I also introduced the `depth` prop in our `Popover` component. This is in preparation to refactor our DatePickerInput & DatePickerInputRange components, which will be based on the Popover.

**Note:** most of the changed lines are just in the storybook playground stories.

### Breaking changes

Select: `usePortal` prop has been removed because it will always use a Portal

### DIFF
--- a/src/components/dialog/theme.css
+++ b/src/components/dialog/theme.css
@@ -16,7 +16,7 @@
   position: fixed;
   top: 0;
   width: 100vw;
-  z-index: 600;
+  z-index: 400;
 }
 
 .dialog-base {
@@ -33,6 +33,7 @@
     transform var(--animation-duration) var(--animation-curve-default);
   transition-delay: var(--animation-delay);
   width: 100vw;
+  z-index: 2;
 }
 
 .inner {
--- a/src/components/popover/Popover.js
+++ b/src/components/popover/Popover.js
@@ -55,6 +55,7 @@ class Popover extends PureComponent {
       children,
       className,
       color,
+      depth,
       fullHeight,
       lockScroll,
       onOverlayClick,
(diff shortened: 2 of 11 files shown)


src/components/dialog/theme.css
```diff
@@ -33,6 +33,7 @@
     transform var(--animation-duration) var(--animation-curve-default);
   transition-delay: var(--animation-delay);
   width: 100vw;
+  z-index: 2;
```
kevinwaelkens: Q: Is working with low values like these smart? Maybe keep some more breathing-space? Just thinking out loud (so keep the bigger values?).

driesd: It's a relative value. Once a wrapper has `position: relative, absolute or fixed`, the inner z-index values are being added up to the parent value. In this case `.dialog-base` will have a z-index of `400 + 2`.

You will also notice that the Overlay component always have a `z-index: 1`.

Dialog wrapper: 400
>   Overlay: 1
>   Dialog: 2

src/components/popover/Popover.js
```diff
@@ -164,6 +168,7 @@ Popover.propTypes = {
 Popover.defaultProps = {
   active: true,
   backdrop: 'dark',
+  depth: 300,
```
lowiebenoot: Any reason why you're not just calling it zIndex?

driesd: I was inspired by [teamleader.design](https://teamleader.design/primitives/depths). Although it's a good remark.

Do you guys think I should rename it to `zIndex` instead of `depth`?

ArnaudWeyts: I think it's better to just use zIndex to avoid any confusion

PR #535, opened 2019-02-27
[AHOY-218] Date picker - refactor based on Input & Popover

driesd:
### Description

This PR refactors our `DatePickerInput` component based on our own `Input` & `Popover` components.

Advantages of using the `Input` component:
- all states are already in place: focus, disabled, read-only, error, warning, help text, placeholder, ...
- it comes with 3 sizes, has support for icons, prefix, suffix, etc...

Advantages of using the `Popover` component:
- it uses a portal to avoid issues in an overflow context (Popover & Dialog)
- it becomes scrollable when there's not enough space to render
- it switches positions when there's not enough space to render
- it has its own markup & behavior

**Todo in a follow-up PR:** provide a decent `parseDate` function that can convert a localized string value into a JS date. Currently, it's not possible to type in the input field to change the date in the date picker.

**Thinks we consider doing next:** provide a way to switch years/months without navigating with the arrow buttons.

schermafdruk 2019-02-27 11 35 26

#### Screenshot after this PR

schermafdruk 2019-02-27 11 35 01

### Breaking changes

It's not possible anymore to type in the input field to change the date in the date picker.


src/components/datepicker/utils.js
```diff
@@ -18,3 +19,9 @@ export const isSelectingFirstDay = (from, to, day) => {
   const isRangeSelected = from && to;
   return !from || isBeforeFirstDay || isRangeSelected;
 };
+
+export function JSDateToLocaleString(date, locale, format = DateTime.DATE_SHORT) {
+  return DateTime.fromJSDate(date)
+    .setLocale(locale)
+    .toLocaleString(format);
+}
```
nickwaelkens: Feel free to use ES6 notation

src/components/datepicker/DatePickerInput.js
```diff
@@ -1,20 +1,20 @@
...
-import WeekDay from './WeekDay';
-import { convertModifiersToClassnames } from './utils';
-import { IconCalendarSmallOutline } from '@teamleader/ui-icons';
-import ValidationText from '../validationText';
+import Input from '../input';
+import Popover from '../popover';
 import cx from 'classnames';
-import omit from 'lodash.omit';
 import theme from './theme.css';
+import MomentLocaleUtils from 'react-day-picker/moment';
```
nickwaelkens: This on its turn imports the entire `moment` library. No go 👎 
https://github.com/gpbl/react-day-picker/blob/master/src/addons/MomentLocaleUtils.js#L3

driesd: I know, this `react-day-picker` package relies on the `Moment` library. I previously tried to convert these local utils to work with Luxon.
But, the problem is with the `getFirstDayOfWeek` function, which decides if the locale first day of the week is a monday or sunday. `Luxon` (based on the Intl api) or `Date-fns` doesn't provide such a function, forcing us to keep using Moment under the hood of our DatePickerInput component.

Before this PR we had to pass the `MomentLocaleUtils` as a prop. But since we don't have a solid alternative, I decided to build this inside our `DatePickerInput` component.

Unless anyone has a better solution, I think we should merge this PR :)

**Sources:**
https://github.com/moment/luxon/issues/373
https://stackoverflow.com/questions/50350110/how-to-create-momentlocaledata-firstdayofweek-in-luxon

lowiebenoot: And we can't provide the `getFirstDayOfWeek` function ourselves in any way? Importing the entire `momentjs` package is not really acceptable as it increases our package size a lot.

driesd: @lowiebenoot You mean by hard-coding the locales (string) and their first day (integer) of the week?

Something like this?

```
const firstDayOfWeek = {
  fr: 1,
  nl: 1,
  pt: 0,
...
};

const getFirstDayOfWeek = locale => {
  return firstDayOfWeek[locale];
}
```

driesd: That's also basically how moment does it :)

lowiebenoot: Yeah, I guess. It has to be "hard-coded" somewhere in the end... It's not that you can derive it from the language so that code has to be somewhere.

src/components/datepicker/DatePickerInput.js
```diff
@@ -1,20 +1,19 @@
...
-import WeekDay from './WeekDay';
-import { convertModifiersToClassnames } from './utils';
-import { IconCalendarSmallOutline } from '@teamleader/ui-icons';
-import ValidationText from '../validationText';
+import Input from '../input';
+import Popover from '../popover';
 import cx from 'classnames';
-import omit from 'lodash.omit';
 import theme from './theme.css';
+import LocalUtils, { formatDate } from './localUtils';
```
lowiebenoot: Why `localUtils` instead of `localeUtils`?

driesd: Damn, typo 🙈

PR #537, opened 2019-03-05
[AHOY-286] Input elements: innerRef

driesd:
### Description

This PR implements the `innerRef` prop in our `InputBase` component. The `innerRef` prop is passed down to the actual `html input` element as `ref`.

This makes it possible to directly target the `html input element` of an `Input`, `NumericInput` or `Textarea` component.

#### Screenshot after this PR

schermafdruk 2019-03-05 14 21 19

### Breaking changes

None.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ### Added
 
+- `InputBase`: added `innerRef` prop to directly target the `html input element` of an `Input`, `NumericInput` or `Textarea` component. ([@driesd](https://github.com/driesd) in [#537](https://github.com/teamleadercrm/ui/pull/537))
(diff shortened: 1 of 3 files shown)


ArnaudWeyts:
Wait, so we do have to pass it as `innerRef` now? I thought the plan was to be able to just use `ref` on the component?

lowiebenoot:
I think I actually prefer using `innerRef`. Using `ref` with `forwardRef` only seems to make things more complicated. Or can using `innerRef` give weird issues?

ArnaudWeyts:
@lowiebenoot, I guess it does make the component's code a bit more complicated but it's just a wrapper tbh 🤷‍♂️ I thought we agreed that being able to just use `ref` would be more intuitive, but maybe I'm mistaken

duivvv:
A bonus for using `innerRef` is that this is not a breaking change. But I would want to move to `ref` some day.

PR #553, opened 2019-03-19
Badge: refactor using our own Icon component

driesd:
### Description

This PR refactors our `Badge` component to use our `Icon` component instead of a custom styled span.

### Breaking changes

None.

### DIFF
--- a/src/components/badge/Badge.js
+++ b/src/components/badge/Badge.js
@@ -1,6 +1,7 @@
 import React, { createRef, PureComponent } from 'react';
 import PropTypes from 'prop-types';
 import Box from '../box';
+import Icon from '../icon';
 import { colorsWithout } from '../../constants';
 import { TextBody } from '../typography';
 import cx from 'classnames';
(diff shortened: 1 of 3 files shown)


src/components/badge/Badge.js
```diff
@@ -25,6 +26,20 @@ class Badge extends PureComponent {
...
   };
 
+  renderIcon = () => {
+    const { color, icon, inverse } = this.props;
+
+    return (
+      <Icon
+        className={theme['icon']}
+        color={color === 'neutral' ? 'teal' : color}
+        tint={inverse ? 'lightest' : 'darkest'}
```
kevinwaelkens: Q: Don't we want to be able to also pass down a tint? Or will this *never* be the case?

driesd: According to the design specs it should be neutral darkest :)

kevinwaelkens: But now it's going to be `teal` `darkest` when it's `neutral` right :P?

driesd: That is correct and is expected behavior. We did this for contrast reasons. You can check the design specs on Zeplin if you want: https://app.zeplin.io/project/599af68db2f4978a56dec839/screen/5a537ce8a7baf1d15409c86e

PR #554, opened 2019-03-19
DurationInput component

TaraldRotsaert:
### Description
- Addition of a new DurationInput component where our user can give in hours, minutes and seconds (time tracking, etc.).
- This component isn't based on an input "type=time" field, but consists of three number input fields, since we want to be able to give in more than 24 hours.

#### Screenshot
Screen Shot 2019-03-19 at 17 11 51

### Breaking changes
- None

### Manual check
- Check the several input fields by trying different "scenarios". Side note: right now you can only change the value with the spinners.


ArnaudWeyts:
Played around with it, seems like there's still some weird things happening though:

* Decreasing the minute input from `01` changes the input to `60`, from this point on you can get the input in some pretty weird state, up to the point that it doesn't display anything in the inputs anymore.

Also a suggestion:

One of the first things I tried was holding down on the increase button to keep increasing the input but that doesn't work, something for a future PR?

TaraldRotsaert:
I hoped there wouldn't be any bugs in the increase/decrease, i am going to look into it to fix it. The suggestion for the spinners is definitely something interesting to consider for a future PR. Thanks for the feedback!

src/components/input/DurationInput.js
```diff
@@ -0,0 +1,404 @@
...
+          newSeconds = 0;
+        }
+        this.setState({ hours: newHours, seconds: newSeconds, minutes: newMinutes }, () =>
+          this.props.onChange(durationToSeconds(hours, newMinutes, newSeconds)),
+        );
+        break;
+
+      default:
+        break;
+    }
```
timdegroote: Something that might help simplify these calculations could be to hold a single value in your state, this would be the total amount of seconds which represents the duration. In the `updateStep` function we should then be able to add the `n` to what's in the state. (`n * 60 * 60` for type hours, `n * 60` for minutes and `n` for seconds).

Then we can move the work of splitting this into hours/minutes/seconds to our render function to show them in the separate input fields. An easy trick to do this is to use rest calculations with the modulo operator (`%`).

Something like this could work:
```
const secondsInMinute = 60;
const secondsInHour = secondsInMinute * 60;

const splitSecondsIntoHoursMinutesSeconds = seconds => {
	const hours = Math.floor(seconds / secondsInHour);
	const minutes = Math.floor((seconds % secondsInHour) / secondsInMinute);
	const remainingSeconds = seconds % secondsInMinute;
	return {
		hours,
		minutes,
		seconds: remainingSeconds,
	};
};
```

src/components/input/DurationInput.js
```diff
@@ -0,0 +1,404 @@
...
+    switch (type) {
+      case 'hours':
+        if (hours >= 0 && n >= 0) {
+          newHours = hours + n;
+        } else if (hours >= Math.abs(n) && n < 0) {
+          newHours = hours + n;
+        } else if (hours <= Math.abs(n) && n < 0) {
+          newHours = 0;
+        }
+        this.setState({ hours: newHours }, () => this.props.onChange(durationToSeconds(newHours, minutes, seconds)));
```
lowiebenoot: `onChange` is not required, so you need to add a check.

src/components/input/DurationInput.js
```diff
@@ -0,0 +1,404 @@
...
+  state = {
+    hours: 0,
+    minutes: 0,
+    seconds: 0,
+    hoursInputHasfocus: false,
+    minutesInputHasfocus: false,
+    secondsInputHasfocus: false,
+  };
+
+  updateStep = (type, n) => {
```
lowiebenoot: Can you give a better name to `n`? Why not just use `step`?

src/components/input/DurationInput.js
duivvv: Damn, I was writing this down 

So. I agree 😃

src/components/input/DurationInput.js
```diff
@@ -0,0 +1,404 @@
...
+    const { hours, minutes, seconds } = this.state;
+    let newHours = hours;
+    let newMinutes = minutes;
+    let newSeconds = seconds;
+
+    switch (type) {
+      case 'hours':
```
lowiebenoot: You can make this calculation easier by using `Math.max()`.

```js
case 'hours':
  newHours = Math.max(0, hours + n);
```

src/components/input/DurationInput.js
```diff
@@ -0,0 +1,404 @@
...
+          newHours = hours + n;
+        } else if (hours >= Math.abs(n) && n < 0) {
+          newHours = hours + n;
+        } else if (hours <= Math.abs(n) && n < 0) {
+          newHours = 0;
+        }
+        this.setState({ hours: newHours }, () => this.props.onChange(durationToSeconds(newHours, minutes, seconds)));
+        break;
+
+      case 'minutes':
```
lowiebenoot: I think you can make all these calculations a lot simpler by using some Math functions and operators.

src/components/input/DurationInput.js
```diff
@@ -0,0 +1,404 @@
...
+        <Box display="flex" alignItems="center" justifyContent="space-between">
+          <Box className={classNamesHours} {...boxProps}>
+            <div className={theme['input-wrapper']}>
+              <div className={theme['input-inner-wrapper']} style={{ flex: '1 2 auto' }}>
+                <InputBase
+                  value={hours < 10 ? `0${hours}` : hours}
+                  type="number"
+                  size={size}
+                  {...commonInputProps}
+                  id={'hours'}
```
lowiebenoot: The component really shouldn't rely on id's. Try to use refs for this.

src/components/input/DurationInput.js
```diff
@@ -0,0 +1,404 @@
...
+import cx from 'classnames';
+import InputBase from './InputBase';
+import { TextBody } from '../typography';
+import Box, { pickBoxProps } from '../box';
+import ValidationText from '../validationText';
+import { durationToSeconds } from '../utils';
+
+import theme from './theme.css';
+
+class SpinnerControls extends PureComponent {
```
lowiebenoot: Not really a fan of having 2 classes in 1 file.

src/components/input/DurationInput.js
lowiebenoot: Made some comments on it as well, but this indeed seems to be the best solution 👍

TaraldRotsaert:
Thanks for all the feedback! I am going to look into it, i definitely learned a lot from it.

src/components/input/DurationInput.js
TaraldRotsaert: I am going to make a separate PR for this change so the `NumericInput` component can also benefit from it.

src/components/input/DurationInput.js
```diff
@@ -288,16 +244,34 @@ class DurationInput extends PureComponent {
     );
 
     const boxProps = pickBoxProps(others);
+    const seconds = this.state.seconds;
+
+    const secondsInMinute = 60;
+    const secondsInHour = secondsInMinute * 60;
+
+    const splitSecondsIntoHoursMinutesSeconds = seconds => {
```
duivvv: move this out of the render function and calculate the values once in render
now you're creating this function on every render and calling it 9 times per render.

src/components/input/DurationInput.js
```diff
@@ -302,9 +302,9 @@ class DurationInput extends PureComponent {
                 <div className={theme['input-inner-wrapper']} style={{ flex: '1 2 auto' }}>
                   <InputBase
                     value={
-                      splitSecondsIntoHoursMinutesSeconds(seconds).minutes < 10
-                        ? `0${splitSecondsIntoHoursMinutesSeconds(seconds).minutes}`
-                        : splitSecondsIntoHoursMinutesSeconds(seconds).minutes
+                      this.splitSecondsIntoHoursMinutesSeconds(seconds).minutes < 10
```
duivvv: You moved it out, but you're still calling it 9 times.

- call it once in above the return
- destructure the result into hours, minutes and seconds
- use the minutes, seconds, hours variables at the places where you call the function

src/components/input/DurationInput.js
```diff
@@ -0,0 +1,415 @@
...
+      const minutes = Math.floor((seconds % secondsInHour) / secondsInMinute);
+      const remainingSeconds = seconds % secondsInMinute;
+      return {
+        hours,
+        minutes,
+        seconds: remainingSeconds,
+      };
+    };
+
+    const { hours, minutes, seconds } = splitSecondsIntoHoursMinutesSeconds(this.state.seconds);
```
duivvv: Now you're still calling the function defined within render?

src/components/utils/time.js
```diff
@@ -1,3 +1,3 @@
-export const durationToSeconds = (hours, minutes, seconds) => {
-  return hours * 3600 + minutes * 60 + seconds;
+export const durationToSeconds = seconds => {
+  return seconds;
```
ArnaudWeyts: 🤔 No need for this function anymore

src/components/input/DurationInput.js
TaraldRotsaert: I moved the function out of the component and placed it in the utils folder. So we can use it in other files if necessary.

TaraldRotsaert:
I am going to make a separate PR with some further changes to input.

TaraldRotsaert:
Closed since we don't have any use case at this moment.

duivvv:
@TaraldRotsaert the new calendar (if picked up again) could use this component in the future

PR #561, opened 2019-03-22
Fix: disabled and read-only on ie11

TaraldRotsaert:
### Description

- Separated the grouped css for the disabled and read-only props of input elements, otherwise, this code doesn't get executed on IE11.

Schermafdruk 2019-03-22 09 45 27

#### Screenshot after this PR

Schermafdruk 2019-03-22 09 45 42

### Breaking changes

None

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 ### Changed
 
 - `TimeInput`: added time input validation. ([@mikeverf](https://github.com/mikeverf) in [#558](https://github.com/teamleadercrm/ui/pull/558))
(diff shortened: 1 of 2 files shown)


mikeverf:
Q: Do the `readOnly` and `disabled` attributes actually *disable* the inputs too on IE? It should, but you never know with IE :)

driesd:
> Q: Do the `readOnly` and `disabled` attributes actually _disable_ the inputs too on IE? It should, but you never know with IE :)

A: Yes, the inputs are actually disabled for user input

mikeverf:
@driesd Because we noticed the issue with the DatePicker, which didn't look disabled, but was still clickable as well (datepicker popover still popped up)

driesd:
> @driesd Because we noticed the issue with the DatePicker, which didn't look disabled, but was still clickable as well (datepicker popover still popped up)

Yes I remember, that's why we took some action ;)

PR #566, opened 2019-03-25
New component: TimerPulser

TaraldRotsaert:
### Description
- Add a TimerPulser animation component

#### Screenshot after this PR
Screen Shot 2019-03-25 at 10 19 29

### Breaking changes
- None

### DIFF
--- a/src/components/timerPulser/theme.css
+++ b/src/components/timerPulser/theme.css
@@ -1,0 +1,75 @@
+@import '@teamleader/ui-colors';
+@import '@teamleader/ui-utilities';
+@import '@teamleader/ui-animations';
+
+:root {
+  --small-outer-size: calc(2.4 * var(--unit));
+  --medium-outer-size: calc(3 * var(--unit));
+  --large-outer-size: calc(3.6 * var(--unit));
+
+  --small-inner-size: calc(0.8 * var(--unit));
+  --medium-inner-size: calc(var(--unit));
+  --large-inner-size: calc(1.2 * var(--unit));
+}
+
+.pulser-container {
+  width: 30px;
+  height: 30px;
+  border-radius: 50%;
+  background-color: var(--color-violet-dark);
+  display: flex;
+  align-items: center;
+  justify-content: center;
+
+  &:hover {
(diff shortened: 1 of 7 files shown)


src/components/timerPulser/theme.css
```diff
@@ -0,0 +1,132 @@
...
+  position: relative;
+  z-index: 3;
+  &::after {
+    content: '';
+    width: 100%;
+    height: 100%;
+    background: white;
+    display: block;
+    position: absolute;
+    z-index: -2;
```
kevinwaelkens: Q: Won't this z-index act up weirdly with other regularly z-indexed items now?

TaraldRotsaert:
Stephen taught it would be a good addition to our library since we are going to reuse the animation in the future.

duivvv:
Wouldn't it make more sense to think about re-using the animation only then? (maybe a HOC or a render-prop component)

Now we're adding a very specific stop button with animation in UI?

kevinwaelkens:
I'm also a little surprised by this addition tbf 😬

ArnaudWeyts:
Agree with @lowiebenoot and @duivvv here, I don't think this should be a ui-component, I can understand that you'd want to re-use the animation but then it belongs in [ui-animation](https://github.com/teamleadercrm/ui-animations).

mikeverf:
So what's happening with this PR, @TaraldRotsaert? Are you adding it to `ui-animations` and closing here? Or do you really need this entire component to be in ui?

TaraldRotsaert:
Somewhere today or monday Stephen and i are going to look into it and make a decision.

stephen-tl:
Discussed this with @TaraldRotsaert and we should add this to the styleguide. It is a dedicated part of our time tracking patterns and will be needed for other components that involve time tracking.

TaraldRotsaert:
We are not going to integrate the animation in our `ui-animations` library but i am going to expand this component so it has more options and can be displayed in several ways.

kevinwaelkens:
🚨 	Conflicts need to be fixed first

PR #568, opened 2019-03-25
AvatarName: new component

TaraldRotsaert:
### Description
- Add a AvatarName component that displays the first and last letter of the name of a user, client, ...

#### Screenshot after this PR
Screen Shot 2019-03-25 at 15 04 08

### Breaking changes
- None

### DIFF
--- a/src/components/avatar/AvatarInitials.js
+++ b/src/components/avatar/AvatarInitials.js
@@ -5,8 +5,8 @@ import cx from 'classnames';
 import theme from './theme.css';
 import { Heading4 } from '../..';
 
-class AvatarName extends PureComponent {
-  getfirstLetters = () => {
+class AvatarInitials extends PureComponent {
+  getInitials = () => {
     const name = this.props.value;
     const nameSplitted = name.split(' ');
     const firstLetter = nameSplitted
@@ -38,13 +38,13 @@ class AvatarName extends PureComponent {
 
     return (
       <Box className={avatarClassNames} {...others} data-teamleader-ui="avatar-name">
-        <Heading4 className={theme['content']}>{this.getfirstLetters()}</Heading4>
+        <Heading4 className={theme['content']}>{this.getInitials()}</Heading4>
       </Box>
     );
   }
 }
 
-AvatarName.propTypes = {
+AvatarInitials.propTypes = {
   /** A class name for the wrapper to give custom styles. */
   className: PropTypes.string,
   /** The color of the avatar. */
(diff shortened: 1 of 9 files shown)


src/components/avatar/AvatarInitials.js
```diff
@@ -0,0 +1,59 @@
+import React, { PureComponent } from 'react';
+import PropTypes from 'prop-types';
+import Box, { pickBoxProps } from '../box';
+import cx from 'classnames';
+import theme from './theme.css';
+import { Heading4 } from '../..';
```
kevinwaelkens: Not sure about this path syntax, so this is exported from the index file from 2 folders higher?

src/components/avatar/AvatarInitials.js
```diff
@@ -0,0 +1,59 @@
...
+
+class AvatarInitials extends PureComponent {
+  getInitials = () => {
+    const { name } = this.props;
+    const nameSplitted = name.split(' ');
```
kevinwaelkens: I think `splitted` is incorrect English, so PP: `splitName`

src/components/avatar/AvatarInitials.js
```diff
@@ -0,0 +1,59 @@
...
+    const firstLetter = nameSplitted[0].charAt(0);
+    if (nameSplitted.length > 1) {
+      const lastLetter = nameSplitted[nameSplitted.length - 1].charAt(0);
```
kevinwaelkens: It looks like it's not possible to get 3 initials or more, so I'd just go for `nameSplitted[1].charAt(0)`

src/components/avatar/AvatarInitials.js
mikeverf: I'd import that component from its own file, importing everything from one index file (that probably imports and exports this component too) can give circular dependencies, which can be quite annoying if you're more strict environments

src/components/avatar/AvatarInitials.js
TaraldRotsaert: We always want the first letter of the first and last name, so if the name would be: "Peter Maxim Karl Weytens" the result needs to be: "PW". Otherwise it would be "PM".

mikeverf:
❌Manually checked, I think the centering of the text is dependent on the length of the string;
image

src/components/avatar/AvatarInitials.js
kevinwaelkens: Aha fair enough, you're right 👍

lowiebenoot:
Can we make some kind of avatar component that either renders an image (if you provide one), or renders the letters (if you provide a name), and chooses a random color, based on an id you provide.

I did something similar to that, where I had an array with 5 placeholder avatars.
```js
getImage() {
    // just a temporary way of transforming the id into a number from 0 to 4, so we can get an image for it
    const number =
      this.props.deal.responsibleUser.id.split('').reduce((total, char) => total + char.charCodeAt(0), 0) % 5;
    return avatarList[number];
  }
```

PR #571, opened 2019-03-26
🐛 Fix missing locale values first day of week

kevinwaelkens:
### Description

There were some locale values missing in the firstDayOfWeek map object, this caused crashes.

### Breaking changes

- None


kevinwaelkens:
Should've started from `next-release`, remaking this PR.

PR #575, opened 2019-03-26
Fix: AvatarInitials set letter-spacing to 0

TaraldRotsaert:
### Description
- Set the `letter-spacing` of the initials to 0, otherwise the text jumps 1px to the right.

Screen Shot 2019-03-26 at 13 47 58

#### Screenshot after this PR
Screen Shot 2019-03-26 at 13 47 38

### Breaking changes
- None.

### DIFF
--- a/src/components/avatar/theme.css
+++ b/src/components/avatar/theme.css
@@ -95,7 +95,6 @@
 
   .content {
     padding-top: var(--content-padding-tiny);
-    letter-spacing: 0.4px;
     font-size: 8px;
     line-height: 16px;
   }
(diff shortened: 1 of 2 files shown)


src/components/avatar/theme.css
```diff
@@ -185,6 +184,7 @@
     color: white;
     text-align: center;
     text-transform: uppercase;
+    letter-spacing: 0px;
```
ArnaudWeyts: Q: is this rule really needed? Won't the browser default `normal` value already do the trick?

driesd: A: This rule can idd be dropped

TaraldRotsaert: The Heading4 has a letter-spacing of .6px by default so it needs to be "reset" manually in the css.

PR #577, opened 2019-03-27
AvatarInitials: center initials with flexbox instead of paddings

TaraldRotsaert:
### Description
- In this PR the initials get centred with flexbox instead of paddings and text-align.

### Breaking changes
- None


TaraldRotsaert:
Visually not, but right now we are dependent on different paddings for the vertical alignment.

PR #583, opened 2019-04-02
Maintenance: change bordered prop DataGrid

TaraldRotsaert:
### Description
- Changes to the css of the `DataGrid` so the `bordered` prop only triggers a top and bottom border.

Screen Shot 2019-04-02 at 10 07 41

#### Screenshot after this PR
Screen Shot 2019-04-02 at 10 08 10

### Breaking changes
- None


TaraldRotsaert:
@lowiebenoot the goal is to use the new Widget component if you want borders. If it should be really necessary you can manually add the borders to the side.

PR #584, opened 2019-04-02
DatePicker: Add bordered prop

TaraldRotsaert:
### Description
- Add a `bordered` prop to our DatePicker component.

Screen Shot 2019-04-02 at 13 28 26

#### Screenshot after this PR
Screen Shot 2019-04-02 at 13 28 37

### Breaking changes
- None


mikeverf:
Q: So this defaults to no border? Is this desired default styling? Because if it's not, this will break the UI everywhere DatePicker is used

TaraldRotsaert:
@mikeverf we are going to put the border on true by default so it doesn't break anything, good remark!

PR #592, opened 2019-04-08
New component: YearPicker

TaraldRotsaert:
### Description
- This PR adds a new component: the `YearPicker`. Our YearPicker can be used standalone or in a Widget.

#### Standalone
Screenshot 2019-04-08 at 16 48 16

#### Widget
Screenshot 2019-04-08 at 16 47 58

### Breaking changes
- None

### DIFF
--- a/src/components/datepicker/FullNavigationBar.js
+++ b/src/components/datepicker/FullNavigationBar.js
@@ -1,7 +1,6 @@
 import React, { PureComponent } from 'react';
 import PropTypes from 'prop-types';
-import { Box } from '../box';
-import { IconButton } from '../button';
+import { Button, ButtonGroup } from '../button';
 import {
   IconArrowLeftMediumOutline,
   IconArrowRightMediumOutline,
@@ -9,7 +8,7 @@ import {
   IconArrowRightSmallOutline,
 } from '@teamleader/ui-icons';
 
-class NavigationBar extends PureComponent {
+class FullNavigationBar extends PureComponent {
   handlePreviousClick = () => {
     this.props.onPreviousClick();
   };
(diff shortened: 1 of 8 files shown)


src/components/datepicker/FullNavigationBar.js
```diff
@@ -0,0 +1,90 @@
...
+import PropTypes from 'prop-types';
+import { Box } from '../box';
+import { Button, ButtonGroup } from '../button';
+import {
+  IconArrowLeftMediumOutline,
+  IconArrowRightMediumOutline,
+  IconArrowLeftSmallOutline,
+  IconArrowRightSmallOutline,
+} from '@teamleader/ui-icons';
+import { Section, Heading2 } from '../..';
```
kevinwaelkens: Can you not import this from the index file please? About the circular dependencies @mikeverf talked

src/components/datepicker/FullNavigationBar.js
```diff
@@ -0,0 +1,90 @@
...
+
+class FullNavigationBar extends PureComponent {
+  handlePreviousClick = () => {
+    this.props.onPreviousClick();
+    this.props.onClickPrevious();
+  };
+
+  handleNextClick = () => {
+    this.props.onNextClick();
+    this.props.onClickNext();
```
kevinwaelkens: Are these double functions just to provide both ways of writing it? 🤔

TaraldRotsaert: The `onNextClick` and `onPreviousClick` are functions of `react-day-picker` itself that update our calendar. The `onClickNext` and `onClickPrevious` are functions that we will use to update our `displayedMonth`, these functions are necessary to make the custom "today" button work.

src/components/datepicker/YearPicker.js
```diff
@@ -0,0 +1,173 @@
...
+            month={displayedMonth}
+            className={classNames}
+            classNames={theme}
+            locale={locale}
+            localeUtils={LocaleUtils}
+            modifiers={convertModifiersToClassnames(modifiers, theme)}
+            numberOfMonths={12}
+            pagedNavigation
+            navbarElement={
+              <FullNavigationBar
```
ArnaudWeyts: PP: put your conditional `showNavigationBar` in here

src/components/datepicker/FullNavigationBar.js
ArnaudWeyts: Can we maybe add a comment here to make this clear? Or destructure from the props and rename it to `onDayPickerNextClick`, otherwise it's a bit confusing

ArnaudWeyts:
Looks good, I noticed the performance takes a bit of a hit because the `mouseout` handler takes quite some time thought, but probably something we can optimize in the future

duivvv:
Q: Why are we adding a specific story 'in a widget', can't we wrap each component in a widget? 🤔

TaraldRotsaert:
@duivvv because we want to have the option to have a standalone version. In combination with 
our widgets we see them as "compositions" from a design perspective.

src/components/datepicker/FullNavigationBar.js
TaraldRotsaert: Good remark!

src/components/datepicker/FullNavigationBar.js
```diff
@@ -0,0 +1,100 @@
...
+};
+
+class FullNavigationBar extends PureComponent {
+  handlePreviousClick = () => {
+    // Function to trigger the next button of our calendar
+    const { onPreviousClick: onDayPickerPreviousClick } = this.props;
+    onDayPickerPreviousClick();
+
+    // Function passed from our parent component to update our state
+    this.props.onClickPrevious();
```
lowiebenoot: All the onNextClick, onPreviousClick, onClickNext, ... are not required according to the propTypes, but there is no check on whether they were provided or not.

duivvv:
> @duivvv because we want to have the option to have a standalone version. In combination with
our widgets we see them as "compositions" from a design perspective.

A Widget is just a container, right?

I mean, we could also create a story for a YearPicker in a Model, in a Box, in a Message, in an Island...

What am I missing here? 😬

src/components/datepicker/FullNavigationBar.js
TaraldRotsaert: Good remark, they should all be required.

TaraldRotsaert: We don't have a proper YearPicker in our library. We have a DatePicker with the option to skip 12 months but this picker misses some crucial items. Right now we have in mind that this component could be used to replace the year calendar in my account for example.

lowiebenoot:
@TaraldRotsaert 
> We don't have a proper YearPicker in our library. We have a DatePicker with the option to skip 12 months but this picker misses some crucial items. Right now we have in mind that this component could be used to replace the year calendar in my account for example.

If we don't know where we are going to use it, we shouldn't add it as a UI component. The only use case I have in mind for this, is indeed the "days off planner".

TaraldRotsaert:
@duivvv and i just discussed it, this component is right now way to specific. I am going to close this PR and in the near future i am going to add a story dedicated to the year view.

PR #594, opened 2019-04-10
New component: SplitButtonMenu

TaraldRotsaert:
### Description
- We already had a SplitButton in our Stories. But since we think this is something we will use a lot more i made a component for it. In a later PR i am going to delete the unnecessary story.

#### Screenshot
Screenshot 2019-07-08 at 14 42 05

### Breaking changes
- None


lowiebenoot:
Are you sure this is the way the component is supposed to work? I find it really weird that the button changes when you click an item in the menu. I would expect that it just immediately executes the action from the menu when clicking on it.

TaraldRotsaert:
We decided that we want it to be like this from a UX standpoint. If it wouldn't be like this the user could accidentally activate actions.

lowiebenoot:
I find that a poor UX standpoint. In my opinion, it just adds an unnecessary click. Dangerous actions (such as delete) should just have a confirmation dialog.

TaraldRotsaert:
Just discussed this with Stephen, Stephen is first going to do some research to split button behaviour.

duivvv:
(just as I would expect it, please ignore if it doesn't align with some guidelines)

- button has a default action (click button -> execute that action)
- other action? click on the arrow. On release -> execute that action

lowiebenoot:
@duivvv on release? So you should hold down your mouse button when clicking the arrow? I find that even weirder 😄

duivvv:
@lowiebenoot 

Yeah you're right

click to open - click to execute makes more sense idd 😄

TaraldRotsaert:
We had a discussion this midday and we came to the conclusion that we will make the split button as @lowiebenoot said. But this button will only have actions that don't need extra conformation or are destructive. If we have in the future a case where we need a destructive action as well we will update/expand the component. I hope this gives a clear answer.

duivvv:
👆 perfect 👌

I think the document template page is the perfect use case example here.

default - download current template
secondary option - download example template

lowiebenoot:
I think it can certainly contain destructive actions as well. Those actions should just have a confirmation dialog (which isn't a part of this component of course). 

If you would add a normal delete button (so not in the split button/menu), you would also have a confirmation dialog. So I don't really understand why we would treat it as something different here.

lowiebenoot:
Also wondering: should we still show the first action in the menu item as well? I guess it's not necessary to repeat that action?

duivvv:
specific reason why this one is on hold? I will probably need it this week.

TaraldRotsaert:
@duivvv there still needs to be done research for the behaviour of the split button.

TaraldRotsaert:
The component is updated wit a prop to decide if the list item action needs to be directly fired or not.

src/components/button/SplitButtonMenu.js
```diff
@@ -0,0 +1,112 @@
...
+  /** Determines which kind of button to be rendered. */
+  level: PropTypes.oneOf(['primary', 'secondary', 'destructive']),
+  /** Size of the button. */
+  size: PropTypes.oneOf(['small', 'medium', 'large']),
+  /** Boolean to decide if we want to fire directly after clicking on a list item. */
+  triggerListAction: PropTypes.bool,
+  /** The function executed, when we click on the main button. */
+  onButtonClick: PropTypes.func.isRequired,
+  /** Width of the main button in px. */
+  widht: PropTypes.string,
```
lowiebenoot: typo

src/components/button/SplitButtonMenu.js
```diff
@@ -0,0 +1,112 @@
+import React, { PureComponent } from 'react';
+import PropTypes from 'prop-types';
+import Box, { pickBoxProps } from '../box';
+import { Button, ButtonGroup } from '.';
+import { Menu } from '../menu';
+import Popover from '../popover';
+import { IconChevronDownSmallOutline } from '@teamleader/ui-icons';
+import pick from 'lodash.pick';
+import { button } from '@storybook/addon-knobs/dist/deprecated';
```
lowiebenoot: This shouldn't be here I guess

src/components/button/SplitButtonMenu.js
lowiebenoot: Why do we even need this prop? Why not just an automatic width?

TaraldRotsaert: Otherwise our button changes from width the moment our label gets changed.

lowiebenoot: Maybe that also just proves that the whole changing button feature is a weird feature.

Anyway: now we have to provide a certain width, which is not really feasible since labels can have completely different widths in different languages. So you might show a button with a lot of "padding" by default...

I find it also very weird that we actually implemented this feature with a prop. So the same looking component can now have 2 different behaviors. Sometimes it will do the action instantly, other times it will change the buttons. Isn't that confusing for the user? 

Do we already have some examples where we would use this split button? Maybe that will help me to wrap my mind around this.

TaraldRotsaert: The width prop was indeed a bad integration. To answer your second question i think the best way to explain this is by just linking you the guidelines(https://app.zeplin.io/project/599af68db2f4978a56dec839/screen/5cc1bc11bc9a132d5b674a50) for the split button.

duivvv: I still see this as a button with a default action and a lesser used secondary option

In the example of document layout templates:

- default action -> `download active template`
- click menu button -> secondary option `download example template`

TaraldRotsaert: The component is updated so it always fires an action on a menu item click and the main action is always displayed in the primary button.

duivvv:
Just spotted the Github implementation (draft PR button)

- select the option 
- change the label
- click the button

so now I'm confused on what route we should take. @stephen-tl ?

ArnaudWeyts:
Soooo, what's happening with this? This PR has been open for almost 2 months, @stephen-tl @driesd @LennertBlommaert? Let's just choose an approach and adjust if necessary?

TaraldRotsaert:
So we decided to reopen this PR once again since we actually want to use this component. Right now we will always fire the action on click, instead of prefilling the main button.
@lowiebenoot @ArnaudWeyts @driesd @mikeverf @kevinwaelkens @LennertBlommaert

PR #599, opened 2019-05-06
[AHOY-302] Dialog: scrollable content

driesd:
### Description

This PR implements the `scrollable` boolean prop in our `Dialog` & `DialogBase` components. Its value is `true` by default.

### Breaking changes

The `scrollable` prop is `true` by default, so this can result in unwanted scroll behavior when using our `Dialog` or `DialogBase` components in your projects.

### DIFF
--- a/src/components/dialog/Dialog.js
+++ b/src/components/dialog/Dialog.js
@@ -5,7 +5,7 @@ import cx from 'classnames';
 
 import theme from './theme.css';
 
-import { Banner, Button, ButtonGroup, DialogBase, Heading2, Heading3, Link } from '../../index';
+import { Banner, Box, Button, ButtonGroup, DialogBase, Heading2, Heading3, Link } from '../../index';
 import { COLORS } from '../../constants';
 
 class Dialog extends PureComponent {
@@ -32,7 +32,7 @@ class Dialog extends PureComponent {
   };
 
   render() {
-    const { children, className, title, ...otherProps } = this.props;
+    const { children, className, scrollable, title, ...otherProps } = this.props;
 
     const classNames = cx(theme['dialog'], className);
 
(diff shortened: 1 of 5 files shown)


src/components/dialog/Dialog.js
```diff
@@ -45,16 +45,24 @@ class Dialog extends PureComponent {
     ]);
 
     return (
-      <DialogBase className={classNames} {...restProps}>
+      <DialogBase className={classNames} {...restProps} scrollable={false}>
         {title && this.getHeader()}
-        {children}
+        {scrollable ? (
+          <Box display="flex" flexDirection="column" overflowY="auto">
```
rathesDot: Q: Any specific reason for `flex` here? This can just be block, right?

src/components/dialog/Dialog.js
```diff
@@ -45,16 +45,24 @@ class Dialog extends PureComponent {
     ]);
 
     return (
-      <DialogBase className={classNames} {...restProps}>
+      <DialogBase className={classNames} {...restProps} scrollable={false}>
```
kevinwaelkens: Q: Here you're providing `scrollable={false}` but the defaultProp is true, so why not `scrollable={scrollable}`?

src/components/dialog/Dialog.js
driesd: A: I need to set the same flex props like the `inner` container currently has. Otherwise, there will be the need to add a `class` or `ref` to the new scrollable container later on.

src/components/dialog/Dialog.js
driesd: A: In this case, we don't want the `DialogBase` component to be scrollable. We actually want is the `children's wrapper` to be scrollable. Therefore I'm explicitly passing `false` to `DialogBase` and let the overflow being handled by the inner Box.

PR #601, opened 2019-05-07
[AHOY-303] Input/Textarea: separate validation message from validation border

driesd:
### Description

This PR makes it possible to show a `validation border color` on `Input` fields & `Textarea` without showing the message underneath.

If you now just pass a `boolean` as `error`, `warning` or `success` prop value, it will only show a border around the input/textarea.

### Breaking changes

None.

### DIFF
--- a/src/components/validationText/ValidationText.js
+++ b/src/components/validationText/ValidationText.js
@@ -10,15 +10,15 @@ class ValidationText extends PureComponent {
   render() {
     const { error, inverse, help, success, warning } = this.props;
 
-    if (error) {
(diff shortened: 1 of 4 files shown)


src/components/validationText/ValidationText.js
```diff
@@ -10,15 +10,15 @@ class ValidationText extends PureComponent {
   render() {
     const { error, inverse, help, success, warning } = this.props;
 
-    if (error) {
+    if (error && typeof error !== 'boolean') {
```
lowiebenoot: You can just do `error === true`

PR #604, opened 2019-05-13
Banner: add size props

TaraldRotsaert:
### Description
- Add `size` prop to the `Banner` component

Screenshot 2019-05-13 at 16 45 07

#### Screenshot after this PR
Screenshot 2019-05-13 at 16 45 38

### Breaking changes
- None

### DIFF
--- a/src/components/banner/Banner.js
+++ b/src/components/banner/Banner.js
@@ -13,11 +13,11 @@ class Banner extends PureComponent {
   }
 
   render() {
-    const { children, className, icon, onClose, fullWidth, ...others } = this.props;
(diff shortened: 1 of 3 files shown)


src/components/banner/Banner.js
```diff
@@ -48,10 +48,13 @@ Banner.propTypes = {
   onClose: PropTypes.func,
   /** If true, component will take the full width of it's container. */
   fullWidth: PropTypes.bool,
+  /** Size prop to add a padding to the Section or Island. */
+  size: PropTypes.string,
```
lowiebenoot: You don't need to add it as a prop because it uses a `Section`/`Island` internally, which already supports the size.
Just adding it to the story knobs should be enough.

PR #627, opened 2019-07-08
ProgressTracker: clickable status bullet only

driesd:
### Description

This PR changes the `clickable area` of our `ProgressStep` component, part of the `ProgressTracker`. After this PR, only the status bullet will be hoverable & clickable instead of its whole wrapper.

### Breaking changes

None.


TaraldRotsaert:
Manually checked ✅

PR #647, opened 2019-07-15
MenuItem: set background-image to none

driesd:
### Description

This PR sets the `background-image` css rule to `none` in our `MenuItem` component. Before this PR, global styling in core would render a bullet background image to all `li` elements.

Screenshot 2019-07-15 12 38 47

#### Screenshot after this PR
Screenshot 2019-07-15 12 39 09

### Breaking changes

None.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ### Added
 
+- `MenuItem`: added `background-image: none;` to the styles. ([@driesd](https://github.com/driesd) in [#647](https://github.com/teamleadercrm/ui/pull/647))
+
 ### Changed
 
 ### Deprecated
--- a/src/components/menu/theme.css
+++ b/src/components/menu/theme.css
@@ -127,6 +127,7 @@
 }
 
 .menu-item {
+  background-image: none;
   height: var(--menu-item-height);
   overflow: hidden;
   position: relative;


kevinwaelkens:
Yeah I think we actually have it quite often where we change the style in core for this annoying reason.

lowiebenoot:
And how would you fix this global style without removing it from places where it should have the background image?

I would opt for putting it in here...

ArnaudWeyts:
@lowiebenoot I don't see why we can't overwrite it in core 🤷‍♂ you can attach a class to it where needed right? Maybe we should have a system in place to apply the overwrites, so they're all together and easy to track, because it doesn't really feel right to put overwrite rules per component in our generic ui-package

I'll unblock but I'm not a big fan of this approach

lowiebenoot:
@ArnaudWeyts well that's the problem. You don't know where it's needed because it's a global style.

lowiebenoot:
@ArnaudWeyts or do you mean overwriting it in just the place where the issue is happening?

rathesDot:
But we know when the background is set but do not want it, right? So we can set the `background-image: none` there, right?

In an ideal world that issue would not exist, so we should fix it at the place where it is introduced and not in this general package here

ArnaudWeyts:
@lowiebenoot Yep, that's what we're already doing right now in some places

lowiebenoot:
Ah sorry, I misunderstood it.

Yeah, you can easily fix it like that. But it can be a bit annoying when it's a component that is often used. Let's say for example that we had to overwrite some css for a `Button` component. Then we would always need to add an extra class to the buttons.

Another way could be that we overwrite the style generally by using the `data-teamleader-ui` attribute.

e.g.
```css
[data-teamleader-ui="button"] {
  background: pink;
}
```

ArnaudWeyts:
🤔 Hmm, that's also a solution indeed, maybe we can even change the original style to use a `:not()` selector? That way we don't even need to overwrite it

driesd:
I'm going to close this PR because I solved it like you guys were suggesting at the end of this PR discussion. I made a [core styling override for li items](https://github.com/teamleadercrm/core/pull/10845/commits/15bbd79486eeb0ddd2db5be744ce2a37833eae43), to not include MenuItem components.

PR #649, opened 2019-07-18
Fix loading spinner position

driesd:
### Description

This PR fixes the `loading spinner position` for our `Button` component.

Screenshot 2019-07-18 13 47 31

#### Screenshot after this PR
Screenshot 2019-07-18 13 46 31

### Breaking changes

None

### DIFF
--- a/src/components/button/theme.css
+++ b/src/components/button/theme.css
@@ -502,7 +502,7 @@
 
 .spinner {
   left: 50%;
-  position: absolute;
+  position: absolute !important;
   top: 50%;
   transform: translate(-50%, -50%);
 }
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,8 @@
 
 ### Fixed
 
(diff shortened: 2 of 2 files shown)


src/components/button/theme.css
```diff
@@ -502,7 +502,7 @@
 
 .spinner {
   left: 50%;
-  position: absolute;
+  position: absolute !important;
```
rathesDot: Is this also an overwrite for another repository?

ArnaudWeyts: Looks like it 😕

ArnaudWeyts: I propose we merge this for now as it is actually broken in core, but we add a `@TODO` comment to atleast know it needs to be fixed. I looked into it for a bit but I couldn't find the cause.

rathesDot: Thanks for the feedback! Let's get this merged for now!

PR #653, opened 2019-08-07
🐛 Make display overwritable for Label

kevinwaelkens:
### Description

When I added `display="flex"` to our Label component the changes weren't coming through. This was because there was a `theme` class overriding other classes which contained `display: block;`.

A fix for this is adding `display="block"` by default to the Box element and accept all `{...others}`. If we then pass `display="flex"` then we easily override the `block`.

Screenshot 2019-08-07 11 45 05

#### Screenshot after this PR

Screenshot 2019-08-07 11 45 19

### Breaking changes

None

### DIFF
--- a/src/components/label/Label.js
+++ b/src/components/label/Label.js
@@ -26,7 +26,7 @@ export default class Label extends PureComponent {
     const Element = size === 'large' ? TextDisplay : TextBody;
 
     return (
-      <Box display="block" element="label" marginBottom={3} className={className} {...others}>
+      <Box display="block" element="label" marginBottom={3} {...others}>
         {React.Children.map(children, child => {
           if (typeof child !== 'string') {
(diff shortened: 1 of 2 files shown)


src/components/label/Label.js
```diff
@@ -25,18 +23,10 @@ export default class Label extends PureComponent {
...
-        [theme['is-inverse']]: inverse,
-      },
-      className,
-    );
-
     const Element = size === 'large' ? TextDisplay : TextBody;
 
     return (
-      <Box element="label" marginBottom={3} className={classNames} {...others}>
+      <Box display="block" element="label" marginBottom={3} className={className} {...others}>
```
lowiebenoot: You can remove the whole `className` prop now and just let it be in the `...others`.

kevinwaelkens:
Ah you're right, sorry

PR #660, opened 2019-08-14
Fix: include postcss.config.js when installing as dependency

ArnaudWeyts:
### Description

In order to use yarn plug'n play, we need the postcss config to be included, adding the postcss config file in the published dependency solves this.
This might also solve the fact that you need a postcss config to use our ui package, though I'm not sure about that.

### Breaking changes

- None.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,8 @@
 
 ### Changed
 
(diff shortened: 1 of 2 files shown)


ArnaudWeyts:
@Kemosabert Wouldn't count on it 🙈 but it's definitely something we need to take a look at in the future 👍

PR #664, opened 2019-08-22
Don't change the date when the date is disabled

Kemosabert:
### Description

When disabling days in the date picker, the `onDayClick` is still triggered when clicking disabled days. In this PR I added a check in the onClick handler to check if the selected date is disabled, that way it does not trigger any state changes or handlers to be triggered further up.

The code and idea comes from: https://react-day-picker.js.org/examples/disabled-interaction

### Breaking changes

none


ArnaudWeyts:
Tried to manually check, it doesn't work, because we convert the modifiers to classnames so instead of `modifiers.disabled` it becomes something like `theme__disabled___`

ArnaudWeyts:
✅ Manually checked

PR #678, opened 2019-09-04
Avatar: refactor & add an editable state

driesd:
### Description

This PR refactors our existing `Avatar` component. Basically all duplicate logic from `AvatarImage` & `AvatarInitials` has moved to `Avatar`.

I also added the `AvatarOverlay` component, which is rendered when the newly introduced `editable` prop is true.

#### Screenshots after this PR

Screenshot 2019-09-04 17 35 13

Screenshot 2019-09-04 17 35 23

Screenshot 2019-09-04 17 36 17

Screenshot 2019-09-04 17 36 07

### Breaking changes

`AvatarImage` & `AvatarInitials` have been removed from export. `Avatar` should be used instead.

### DIFF
--- a/src/components/avatar/theme.css
+++ b/src/components/avatar/theme.css
@@ -177,7 +177,7 @@
     width: var(--large-size);
   }
 
-  .content {
+  .initials {
     font-size: 18px;
   }
 
--- a/src/components/avatar/theme.css
+++ b/src/components/avatar/theme.css
@@ -61,6 +61,16 @@
   padding: 0 var(--overflow-padding-horizontal);
 }
 
+.overlay {
+  background-position: bottom center;
+  background-repeat: no-repeat;
+  bottom: 0;
+  cursor: pointer;
+  left: 0;
+  padding-bottom: 5px;
+  position: absolute;
+}
+
 .dark {
   .overflow {
     background: color(var(--color-teal-darkest) a(10%));
@@ -161,7 +171,8 @@
 .is-large {
   &.avatar-image,
   &.avatar-initials,
-  .image {
+  .image,
+  .overlay {
     height: var(--large-size);
     width: var(--large-size);
   }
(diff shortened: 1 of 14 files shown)


src/components/avatar/theme.css
```diff
@@ -206,6 +222,10 @@
     height: var(--hero-size);
     min-width: calc(var(--hero-size) - var(--overflow-padding-horizontal) * 2);
   }
+
+  .overlay {
+    background-image: url(data:image/svg+xml;base64,PHN2ZyB4bWxucz0iaHR0cDovL3d3dy53My5vcmcvMjAwMC9zdmciIHdpZHRoPSIxMDgiIGhlaWdodD0iMjQiIHZpZXdCb3g9IjAgMCAxMDggMjQiPiAgPHBhdGggZmlsbD0iIzM0NEI2MyIgZmlsbC1vcGFjaXR5PSIuODUiIGQ9Ik0xMjUuNjY0NDY0LDEyMCBDMTEyLjQ4MDgzNiwxMzQuNzI4NjI4IDkzLjMyMjg2NzYsMTQzLjk5NzY2MyA3MiwxNDMuOTk3NjYzIEM1MC42NzcxMzI0LDE0My45OTc2NjMgMzEuNTE5MTYzOCwxMzQuNzI4NjI4IDE4LjMzNTUzNTgsMTIwIFoiIHRyYW5zZm9ybT0idHJhbnNsYXRlKC0xOCAtMTIwKSIvPjwvc3ZnPg==);
+  }
```
ArnaudWeyts: Not sure if we should just dump a base64 string in here 🤔Can't we just use a square background color and position the overlay down and hide the overflow?

rathesDot: I agree on that as well, this would give us more flexibility if we need to change the color or opacity of the overlay at some point

driesd: Agreed, looking into it!

PR #680, opened 2019-09-05
Dependencies: Replace postcss-cssnext

rathesDot:
### Description

This PR replaces the deprecated `postcss-cssnext` plugin with the package the author recommended. Kept the settings as is.

## IMPORTANT

Please try to import this package in your local environment of your microservice and check if the build processes still work as intended.

### Breaking changes

"None"


rathesDot:
@ArnaudWeyts But that is a problem then, right? Shouldn't it be possible to have this lib using `postcss-preset-env` while your service still uses `postcss-cssnext`?

rathesDot:
It seems to be resolved when you add the path to your config file to the postcss-loader in your webpack config ...

```
{
  loader: 'postcss-loader',
  options: {
    config: {
      path: `${__dirname}/postcss.config.js`,
    },
  },
},
```

Don't really understand why this is needed though ...

rathesDot:
Since the current state of the package breaks for everyone not using `postcss-cssnext` anyway, we decided to not investigate further for this PR but have a look at the issue separately.

For now, I added the change as a breaking change in the changelog and also added a not on how to upgrade. Hope that should be enough.

PR #681, opened 2019-09-05
Box: borders

driesd:
### Description

This PR makes it possible to add borders to our `Box` component.

#### Screenshot after this PR
Screenshot 2019-09-05 14 20 35

### Breaking changes

None.


src/components/box/Box.js
```diff
@@ -149,6 +173,8 @@ Box.propTypes = {
 };
 
 Box.defaultProps = {
+  borderColor: 'neutral',
+  borderTint: 'dark',
```
kevinwaelkens: Q: Why no borderWidth defaultprop? won't it be `undefinedpx` now?

driesd: The border styles are only applied when a border width is set, because of these lines:
`...(borderBottomWidth && { borderBottom: this.getBorder(borderBottomWidth) }),`

PR #699, opened 2019-10-03
💄 Remove padding on the right of banner content

rathesDot:
### Description

This PR removes some useless padding on the right of the banner which makes the spacing looking of when the content spans over multiple lines.

Screen Shot 2019-10-03 at 11 15 46 AM

#### Screenshot after this PR

Screen Shot 2019-10-03 at 11 16 05 AM

#### Note

This change has no impact on single-line banners (apart from that they can now have slightly more content in one line)

### Breaking changes

No breaking changes.


driesd:
If you pass a function to the `onClose` prop, it renders an `IconButton` to close the banner. That explains the extra spacing. So the padding should depend on this `onClose` prop to avoid this:
Screenshot 2019-10-03 11 46 40

rathesDot:
Fixed the remark and also added another story element to avoid that we miss this case in the future

PR #700, opened 2019-10-04
[AHOY-337] Radio button - change default background color

driesd:
### Description

This PR changes the `default background-color` of our `Radio` component, from neutral light to neutral lightest.

Screenshot 2019-10-04 11 29 03

#### Screenshot after this PR
Screenshot 2019-10-04 11 28 51

### Breaking changes

None.


rathesDot:
Will the background of the shape be changeable?

driesd:
> Will the background of the shape be changeable?

No, the background is not customizable. Can you clarify why you would need this?

rathesDot:
> > Will the background of the shape be changeable?
> 
> No, the background is not customizable. Can you clarify why you would need this?

In our use-case (and that is why you got the ticket I guess) we would have a grey background and that is why we need a radio/checkbox variation with the white shape.

But if we have a white background, we don't want the shape to be white, right?

PR #721, opened 2019-11-21
Menu item: allow anchor tags

driesd:
### Description

This PR adds an `element` prop which accepts either a `button` (default value) or an `anchor tag`. I also needed to adjust the markup and styling to make sure it still looks like the intended design.

Also, I made it keyboard accessible:

Screenshot 2019-11-21 10 37 09

### Breaking changes

None.

### DIFF
--- a/src/components/menu/MenuItem.js
+++ b/src/components/menu/MenuItem.js
@@ -1,6 +1,6 @@
 import React, { PureComponent } from 'react';
 import PropTypes from 'prop-types';
-import Box from '../box';
+import Box, { omitBoxProps, pickBoxProps } from '../box';
 import Icon from '../icon';
 import { TextBody } from '../typography';
 import cx from 'classnames';
@@ -16,7 +16,7 @@ class MenuItem extends PureComponent {
   };
 
   render() {
-    const { icon, caption, className, destructive, label, selected, disabled, ...others } = this.props;
+    const { icon, caption, className, destructive, element, label, selected, disabled, ...others } = this.props;
 
     const classNames = cx(
       theme['menu-item'],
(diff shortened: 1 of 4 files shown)


ArnaudWeyts:
⏳  Manually checking

src/components/menu/MenuItem.js
```diff
@@ -30,34 +30,42 @@ class MenuItem extends PureComponent {
...
+          textAlign="left"
+          {...(element === 'a' && disabled && { tabIndex: '-1' })}
+          {...restProps}
+        >
+          {icon && (
+            <Icon color={color} tint={tint} marginRight={3}>
+              {icon}
+            </Icon>
           )}
+          <Box flex="1 0 auto">
```
kevinwaelkens: Actually we only want to render this Box when there's a label or a caption right? 
```suggestion
          {label || caption (<Box flex="1 0 auto">
```

src/components/menu/MenuItem.js
```diff
@@ -30,34 +30,42 @@ class MenuItem extends PureComponent {
...
+              <TextBody color={color} tint={tint}>
+                {label}
+              </TextBody>
+            )}
+            {caption && (
+              <TextBody color="neutral" tint="darkest">
+                {caption}
+              </TextBody>
+            )}
+          </Box>
```
kevinwaelkens:
```suggestion
          </Box>)}
```

src/components/menu/MenuItem.js
ArnaudWeyts: Isn't the label required?

ArnaudWeyts:
✅ Manually checked

src/components/menu/MenuItem.js
kevinwaelkens: Doesn't seem required, it's indeed not really a logical use-case, but yeah 🤷‍♂

PR #727, opened 2019-11-27
Fix _blank click on core

kevinwaelkens:
### Description

On core, clicking on a menu item which has the `a` element would append the `href` to the current URL you were on. This only happened when clicking on the `div`/`p` element. When you clicked on the wrapping `a` tag (on the margin of it) it would open in a new tab, which is desired behaviour.

So because, for some reason on core, the `div`/`p` tag are doing something we don't want on click we should pass `pointer-events: none` on it.  (Or we could scour through the core codebase to check where the annoying part is, but that sounds very time-intensive.)

### Breaking changes

- none

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,8 @@
 
 ### Changed
 
+- `MenuItem`: when element is `a` the text container gets `pointer-events: none` ([@kevinwaelkens](https://github.com/kevinwaelkens) in [#727](https://github.com/teamleadercrm/ui/pull/727))
+
 ### Deprecated
 
 ### Removed
--- a/src/components/menu/MenuItem.js
+++ b/src/components/menu/MenuItem.js
@@ -54,7 +54,12 @@ class MenuItem extends PureComponent {
               {icon}
             </Icon>
           )}
-          <Box flex="1 0 auto">
+          <Box
+            className={cx({
+              [theme['menu-item-text-container']]: element === 'a',
+            })}
+            flex="1 0 auto"
+          >
             {label && (
               <TextBody color={color} tint={tint}>
                 {label}
--- a/src/components/menu/theme.css
+++ b/src/components/menu/theme.css
@@ -184,3 +184,7 @@
   -ms-user-select: none;
   user-select: none;
 }
+
+.menu-item-text-container {
+  pointer-events: none;
+}


rathesDot:
I think, we had this kind of discussion already: I don't really think that we should have a CORE specific bug handled in this package.

Can't you add this in the CSS of core?

kevinwaelkens:
@rathesDot You probably thought then I asked if anyone was opposing to add CSS in `core` [here](https://teamleader.slack.com/archives/G2T7PPV9T/p1574681523002500?thread_ts=1574675435.001400&cid=G2T7PPV9T) then, but I meant in the library.

No there's no possibility to fix it in core, as you can see I add the class on a div which has no properties passed to it. I can't access any of the CSS on that div because no classNames or whatever are specified. If I would go for another approach then we need to:
**OR** rework the component in its entirety
**OR** add core-specific properties, which also conflicts with your opinion about core specific bugs being handled in this package..

So I'm not sure if something else is possible here.

rathesDot:
> @rathesDot You probably thought then I asked if anyone was opposing to add CSS in `core` [here](https://teamleader.slack.com/archives/G2T7PPV9T/p1574681523002500?thread_ts=1574675435.001400&cid=G2T7PPV9T) then, but I meant in the library.
> 
> No there's no possibility to fix it in core, as you can see I add the class on a div which has no properties passed to it. I can't access any of the CSS on that div because no classNames or whatever are specified. If I would go for another approach then we need to:
> **OR** rework the component in its entirety
> **OR** add core-specific properties, which also conflicts with your opinion about core specific bugs being handled in this package..
> 
> So I'm not sure if something else is possible here.

I indeed thought, you were going to apply that to core 🙈 Sorry!

I know it is a dirty hack, but couldn't we do something like `.containerClass > div > div` ...

Otherwise, I would like to hear the opinion of others, I'm not entirely against it but I would prefer a solution in core if possible in any way ...

kevinwaelkens:
@rathesDot I think that is also a solution, but if we then change something in the structure, we also have to change something in core... And I think that's more dangerous? Also, always knowing that that is the issue sucks... Imagine this menu being used across all of core and then the structure of the MenuItem changes, then you have to change it everywhere as well... I think that also kind of sucks. I don't think `pointer-events: none;` can ever bring any harm... No?

rathesDot:
Let's stick with it then and think about a solution once we notice that we introduced an issue with this change.

PR #737, opened 2019-12-05
Box: border radius

driesd:
### Description

This PR implements the `borderRadius` prop in our `Box` component. It takes a string value and is undefined by default.

### Breaking changes

None.


rathesDot:
Shouldn't this be a number instead of a string? Or even better, should we have a fixed set of values instead of allowing any value?

driesd:
> Shouldn't this be a number instead of a string? Or even better, should we have a fixed set of values instead of allowing any value?

I agree with you. That would work very well if the values were only `4px` (rounded) and `50%` (circle).

But the caveat here is that we also need to be able to pass the values like`4px 4px 0 0`. This is why I decided to go for a string typed prop.

Another option here is to split up the borderRadius into `borderTopLeftRadius`, `borderTopRightRadius`, `borderBottomLeftRadius`, `borderBottomRightRadius`. And then I can work with fixed values again.

What do you think the best approach would be?

rathesDot:
It's just my personal preference but I would opt for `borderRadius` together with `borderTopLeftRadius` and so on. So `borderRadius` being a shortcut for setting `borderTopLeftRadius`, `borderTopRightRadius`, `borderBottomLeftRadius` and `borderBottomRightRadius`

driesd:
> It's just my personal preference but I would opt for `borderRadius` together with `borderTopLeftRadius` and so on. So `borderRadius` being a shortcut for setting `borderTopLeftRadius`, `borderTopRightRadius`, `borderBottomLeftRadius` and `borderBottomRightRadius`

Ok, I'll do it like this ☝️.
We also opted for this approach when we implemented the `margin` and `padding` props.

src/components/box/Box.js
```diff
@@ -7,6 +7,12 @@ import theme from './theme.css';
 const overflows = ['auto', 'hidden', 'scroll', 'visible'];
 const spacings = [0, 1, 2, 3, 4, 5, 6, 7, 8];
 
+const borderRadiuses = {
+  square: null,
```
rathesDot: Is there a difference of having `null` vs `undefined`?

PR #755, opened 2019-12-17
🔧 Don't require library users to use postcss

ArnaudWeyts:
### Description

This PR allows our UI-library to be used without any further setup.

* Added rollup
* PostCSS gets compiled and bundled
* Added very rudimentary types for each component

### Breaking changes

- There shouldn't really be any breaking changes. Everything should still work as before.

### Manual check

* Run `yarn build` in your local version of the UI
* Start a fresh project (f.e. use create-react-app)
* Run `npm link @teamleader/ui` in your new project
* Import a UI component and use it ✨


driesd:
This guy! 🎉

driesd:
Manually tested with `Button` and `StatusBullet` component. Works like a charm!

PR #756, opened 2019-12-17
NumericInput: refactor and extend spinner

driesd:
### Description

Before this PR it was only possible to render de `spinner controls` inside the input field, on the left. This PR not only renames the `spinner` prop to `stepper`, but also introduces a new way of rendering these controls.

#### No stepper at all
Screenshot 2019-12-17 16 13 12

#### Stepper as a suffix
Screenshot 2019-12-17 16 12 54

#### Stepper as two connected buttons (new)
Screenshot 2019-12-17 16 13 04

### Breaking changes

- changed prop type `spinner` from `boolean` (default `true`) to `oneOf` (default `suffix`)
- renamed `spinner` prop to `stepper`

### DIFF
--- a/src/components/input/NumericInput.js
+++ b/src/components/input/NumericInput.js
@@ -9,6 +9,7 @@ import {
 import Button from '../button';
 import Icon from '../icon';
 import SingleLineInputBase from './SingleLineInputBase';
+import omit from 'lodash.omit';
 import theme from './theme.css';
 
 class StepperControls extends PureComponent {
@@ -154,7 +155,8 @@ class NumericInput extends PureComponent {
   };
 
   render() {
-    const { onChange, suffix, ...others } = this.props;
+    const { onChange, ...others } = this.props;
+    const restProps = omit(others, ['suffix']);
 
     return (
       <SingleLineInputBase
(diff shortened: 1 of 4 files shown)


src/components/input/NumericInput.js
```diff
@@ -89,35 +96,80 @@ class NumericInput extends PureComponent {
...
+        <Button
+          disabled={this.isMaxReached()}
+          icon={<IconAddSmallOutline />}
+          onClick={this.handleIncreaseValue}
+          size={size}
+        />
+      );
+    }
+
+    return connectedRight;
```
sanderbrugge: Q: not entirely sure what this returns? wouldn't `null` suffice?

src/components/input/NumericInput.js
```diff
@@ -89,35 +96,80 @@ class NumericInput extends PureComponent {
...
+        <Button
+          disabled={this.isMinReached()}
+          icon={<IconMinusSmallOutline />}
+          onClick={this.handleDecreaseValue}
+          size={size}
+        />
+      );
+    }
+
+    return connectedLeft;
```
sanderbrugge: Same remark as `getConnectedRight`.

src/components/input/NumericInput.js
driesd: A: `connectedRight` and `connectedLeft` are props that can contain f.e. a `Button`. But, when the stepper controls are rendered left and right, we don't want any extra connected component being rendered.

rathesDot: Simply speaking, if we have a `<NumericInput stepper="connected" connectedRight={<Button />}>` the `connectedRight` would be overwritten.

sanderbrugge: @driesd @rathesDot thanks for explaining, missed that context. Still feels like an odd architecture though.

PR #759, opened 2019-12-18
Input & Select: focus state should not overrule any other state

driesd:
### Description

This PR fixes all `Input` & `Select` states in a way that the focus state cannot overrule any other state (error, warning, success & disabled).

Screenshot 2019-12-18 10 55 41

#### Screenshot after this PR
Screenshot 2019-12-18 10 58 46

### Breaking changes

None.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,9 @@
 
 ### Fixed
 
+- `Input`: focus state should not overrule any other state. ([@driesd](https://github.com/driesd) in [#759](https://github.com/teamleadercrm/ui/pull/759))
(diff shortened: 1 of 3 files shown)


rathesDot:
Wait that isn't the desired behavior? I think it is correct that you have a proper visible focus state no matter what state the input is in. Especially since we do not have outlines set

driesd:
> Wait that isn't the desired behavior? I think it is correct that you have a proper visible focus state no matter what state the input is in. Especially since we do not have outlines set

According to our UX/UI team, It should be like this 🙈. I saw the same behavior in other tools as well. And actually also, it doesn't feel weird to do it like this :)

rathesDot:
I would then argue that we should have a focus state that is separate from the other states but I'm leaving it to the UI/UX team, they will know better than me, I guess.

PR #791, opened 2020-01-08
DatePicker: remove obsolete .overlay styles

driesd:
### Description

This PR removes the obsolete `.overlay` styling from our `DatePicker` component. By doing this, we also eliminate the usage of `composes` completely from our UI library.

### Breaking changes

None.


ArnaudWeyts:
Double checked for usages, looks good to me 👍

PR #810, opened 2020-01-14
Upgrade all babel dependencies at once

driesd:


rathesDot:
Merging this will also close #811

PR #841, opened 2020-02-06
DatePicker: MonthPicker with single select field

mikeverf:
### Description

Add a smaller variant of MonthPicker that uses one selector instead of two fields for when `DatePicker` `size` is set the `small`

Screenshot 2020-02-06 at 11 16 18

#### Screenshot after this PR

Screenshot 2020-02-06 at 11 15 56

### Breaking changes

None

### DIFF
--- a/src/components/datepicker/DatePicker.js
+++ b/src/components/datepicker/DatePicker.js
@@ -4,6 +4,7 @@ import DayPicker from 'react-day-picker';
 import Box, { omitBoxProps, pickBoxProps } from '../box';
 import NavigationBar from './NavigationBar';
 import WeekDay from './WeekDay';
+import MonthPicker from './MonthPicker';
 import { convertModifiersToClassnames } from './utils';
 import cx from 'classnames';
 import theme from './theme.css';
@@ -12,6 +13,7 @@ import uiUtilities from '@teamleader/ui-utilities';
 class DatePicker extends PureComponent {
   state = {
     selectedDate: null,
+    selectedMonth: null,
   };
 
   static getDerivedStateFromProps(props, state) {
@@ -37,9 +39,13 @@ class DatePicker extends PureComponent {
     );
   };
 
+  handleYearMonthChange = selectedMonth => {
+    this.setState({ selectedMonth });
+  };
+
   render() {
-    const { bordered, className, modifiers, size, ...others } = this.props;
-    const { selectedDate } = this.state;
+    const { bordered, className, modifiers, size, withMonthPicker, showWeekNumbers, ...others } = this.props;
+    const { selectedDate, selectedMonth } = this.state;
     const boxProps = pickBoxProps(others);
     const restProps = omitBoxProps(others);
     const classNames = cx(
@@ -56,13 +62,27 @@ class DatePicker extends PureComponent {
       <Box {...boxProps}>
         <DayPicker
           {...restProps}
+          month={selectedMonth}
           className={classNames}
           classNames={theme}
           modifiers={convertModifiersToClassnames(modifiers, theme)}
-          navbarElement={<NavigationBar size={size} />}
+          navbarElement={<NavigationBar size={size} withMonthPicker={withMonthPicker} />}
           onDayClick={this.handleDayClick}
           selectedDays={selectedDate}
           weekdayElement={<WeekDay size={size} />}
+          showWeekNumbers={showWeekNumbers}
+          captionElement={
+            withMonthPicker
+              ? ({ date, localeUtils }) => (
+                  <MonthPicker
+                    date={date}
+                    localeUtils={localeUtils}
+                    onChange={this.handleYearMonthChange}
+                    small={!showWeekNumbers}
+                  />
+                )
+              : undefined
+          }
         />
       </Box>
     );
--- a/src/components/datepicker/DatePicker.js
+++ b/src/components/datepicker/DatePicker.js
@@ -43,6 +43,16 @@ class DatePicker extends PureComponent {
     this.setState({ selectedMonth });
   };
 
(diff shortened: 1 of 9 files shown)


src/components/datepicker/MonthPicker.js
```diff
@@ -0,0 +1,74 @@
...
+
+  const years = [];
+  for (let i = fromMonth.getFullYear(); i <= toMonth.getFullYear(); i += 1) {
+    years.push(i);
+  }
+
+  useEffect(() => {
+    setSelectedMonth({ value: date.getMonth(), label: localeUtils.formatMonthTitle(date) });
+    setSelectedYear(date.getFullYear());
+  }, [date]);
```
ArnaudWeyts: I would just use a memoized value instead of using state and and changing with a `useEffect`, to avoid async issues (because useEffect doesn't block a render):

```js
const selectedMonth = useMemo(() => ({ value: date.getMonth(), label: localeUtils.formatMonthTitle(date)}), [date])
const selectedYear = useMemo(() => date.getFullYear(), [date])
```

src/components/datepicker/theme.css
```diff
@@ -82,7 +82,7 @@
 }
 
 .month-picker-field {
-  width: 87px;
+  width: auto !important;
```
ArnaudWeyts: 🤔

src/components/input/SingleLineInputBase.js
```diff
@@ -117,6 +118,8 @@ SingleLineInputBase.propTypes = {
   warning: PropTypes.oneOfType([PropTypes.bool, PropTypes.node]),
   /** A custom width for the input field */
   width: PropTypes.string,
+  /** Whether to disable styling that hints being able to type in the input field */
+  noInputStyling: PropTypes.bool,
```
ArnaudWeyts: PP: rename this to what it does; f.e. a variable called `typeable` which is set to `true` by default

src/components/datepicker/DatePicker.js
```diff
@@ -43,6 +43,16 @@ class DatePicker extends PureComponent {
...
+  getMonthPickerSize = () => {
+    const { showWeekNumbers, size } = this.props;
+    const monthPickerSizeByDatePickerSize = {
+      small: 'smallest',
+      medium: showWeekNumbers ? 'medium' : 'small',
+      large: 'large',
+    };
+    return monthPickerSizeByDatePickerSize[size];
+  };
+
```
ArnaudWeyts: PP: use `useMemo` for this to avoid recalculating on every render, especially since the size won't really change

src/components/input/SingleLineInputBase.js
mikeverf: The thing is, it's a bit of a weird case; 
it doesn't actually prevent typing in it, it just hides the styling that makes it look like  you can type in it
This all stems from the issue that we're using an Input field here, but don't use the input part of it
So I wanted to put an emphasis on the _styling_ part so that it doesn't seem like it would disable the typing behaviour

src/components/datepicker/theme.css
mikeverf: Would be more helpful to put your thoughts into words here, since I don't know what you're alluring to
So I'll just explain:
As per commit message,
> Fix 100% being inherited from `.wrapper.input`

The `Input` component has general styling on `.wrapper.input` that sets the width to `100%`. In our case, we don't want that; we want that width to be set based on the children's width, so I used `auto`. And since that general `.wrapper.input` styling always overrides, I had to use `!important`

src/components/input/SingleLineInputBase.js
ArnaudWeyts: So there's a case where we want to not show the styling which suggests the input is typeable, but allow users to type in it? Can we not use the input and create a different input for it in that case? Just trying to avoid adding props like these because the issue seems to be somewhere else and our case just doesn't fit this abstraction

src/components/datepicker/theme.css
ArnaudWeyts: Thanks for the explanation! I suppose this could also be fixed by using a more specific input instead of using the abstraction.

src/components/datepicker/DatePicker.js
mikeverf: This is a class component though, so do you mean a general memoization function (like from lodash or something)?

src/components/input/SingleLineInputBase.js
mikeverf: Yeah, we were thinking "button" might be a better fit here, but otherwise we considered this another step below disabled and read-only, where you can't input anything, but you can still click in it (and thus trigger the datepicker)
I think the whole issue really just stems from the use of Input, and I don't know why that choice was made, so I can't just change it. Design just asked to remove the stuff that makes it look like it's editable

src/components/datepicker/theme.css
mikeverf: Sure, but I wouldn't know what would conceptually set it apart from other input types, since it's really specific to our use case, that's why I left it like this until we need a better abtraction

src/components/datepicker/DatePicker.js
ArnaudWeyts: Ah no, didn't realise it was a class component, you can leave it as is then

ArnaudWeyts:
⏳ Manually checking

ArnaudWeyts:
✅ The component itself works

I noticed that the text inside of the small sized month picker doesn't really fit though, should we also just use the short version of the month (first 3 letters)?

Screenshot 2020-02-11 at 09 08 37

mikeverf:
That's fixed by @driesd in the current release, I guess the base branch just isn't updated yet

PR #875, opened 2020-02-18
Dialog: expose a Header, Body and Footer component

lowiebenoot:
### Description

This PR will expose a Header, Body and Footer component on the DialogBase component. You can use for building custom dialogs, without having to copy/mimic the internal parts of the Dialog component

`<DialogBase.Header>`
`<DialogBase.Body>`
`<DialogBase.Footer>`

### Breaking changes

None.

### DIFF
--- a/src/components/dialog/Body.js
+++ b/src/components/dialog/Body.js
@@ -1,0 +1,29 @@
+import React, { PureComponent } from 'react';
+import PropTypes from 'prop-types';
+
+import { Box } from '../../index';
+
+class Body extends PureComponent {
+  render() {
+    const { scrollable, children, ...rest } = this.props;
+
+    if (!scrollable) {
+      return children;
+    }
+
+    return (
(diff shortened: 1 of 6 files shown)


src/components/dialog/Body.js
```diff
@@ -0,0 +1,29 @@
...
+import PropTypes from 'prop-types';
+
+import { Box } from '../../index';
+
+class Body extends PureComponent {
+  render() {
+    const { scrollable, children, ...rest } = this.props;
+
+    if (!scrollable) {
+      return children;
```
kevinwaelkens: Q: Just wondering, why exactly do you need to early return here then? Is it because it might just get scrollable too early (when it's not as big as you want it) that you could force it to not become scrollable? Cause I'd think those properties on Box wouldn't matter as long as it's not too big yet and it would automatically arrange everything when needed huh?

PR #879, opened 2020-02-19
NumericInput: text align right

driesd:
### Description

This PR adds the `textAlignRight` boolean prop to our `InputBase` component. If true, the input text is aligned to the right.

#### Screenshot with textAlignRight FALSE

Screenshot 2020-02-19 16 18 35

#### Screenshot with textAlignRight TRUE

Screenshot 2020-02-19 16 18 28

### Breaking changes

None.

### DIFF
--- a/src/components/input/InputBase.js
+++ b/src/components/input/InputBase.js
@@ -6,7 +6,7 @@ import theme from './theme.css';
 
 class InputBase extends PureComponent {
   render() {
-    const { bold, className, element, innerRef, inverse, size, ...otherProps } = this.props;
+    const { bold, className, element, innerRef, inverse, size, textAlignRight, ...otherProps } = this.props;
 
     const classNames = cx(
       theme['input'],
@@ -14,6 +14,7 @@ class InputBase extends PureComponent {
       {
         [theme['is-inverse']]: inverse,
         [theme['is-bold']]: bold,
+        [theme['has-text-align-right']]: textAlignRight,
       },
       className,
     );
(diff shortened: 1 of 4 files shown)


src/components/input/InputBase.js
```diff
@@ -6,14 +6,15 @@ import theme from './theme.css';
 
 class InputBase extends PureComponent {
   render() {
-    const { bold, className, element, innerRef, inverse, size, ...otherProps } = this.props;
+    const { bold, className, element, innerRef, inverse, size, textAlignRight, ...otherProps } = this.props;
```
ArnaudWeyts: PP: name the property `textAlign` with a default value of `left`

driesd: Well, I also thought about this, but here's what was holding me back:

- `textAlign` already is a `Box` prop, which would be omitted in `SingleLineInputBase` if we don't adjust this component as well.
- it would make `SingleLineInputBase` and `InputBase` less readable
- it would look a bit weird to pass `textAlign` input prop, but omit the `textAlign` Box prop 🤔 
- aligning text to the left is already the default, so the only override here is to align text to the right
- finally, we don't need to center align text for now

So, that's why I shifted lanes and go for `textAlignRight` explicitly.

ArnaudWeyts: Not sure if it's the way to go, because it can be confusing to know you can set a `textAlign` property and it doesn't actually do anything (because it gets set on the `Box`) but either way, it's only a PP.

PR #884, opened 2020-02-25
Storybook: upgrade to 5.3.14

driesd:
### Description

This PR upgrades Storybook to 5.3.13. We do this in preparation for Storybook 6.0 because the current way of configuring is deprecated and will no longer be supported.

So in this PR, we convert the config to its new format.

### Breaking changes

None.

### Next up
Replace `addon-info` & `addon-readme` with `Addon-docs`. Therefore all stories will be ported from CSF to `MDX` format.


ArnaudWeyts:
🔴 When clearing localstorage (so when you land on the documentation for the first time) the dark mode is enabled and everything looks quite weird

driesd:
> 🔴 When clearing localstorage (so when you land on the documentation for the first time) the dark mode is enabled and everything looks quite weird
> 
> Screenshot 2020-02-26 at 09 36 15

Good catch @ArnaudWeyts! I re-added this `base: 'light'` to the Storybook theme config so it prevents switching to dark mode automatically `because of the OS settings`.

PR #899, opened 2020-03-03
♻️ Render AvatarInitials without providing id

kevinwaelkens:
### Description

This PR makes sure the AvatarInitials still gets rendered even when an `id` is not provided. The only thing you actually need is the full name to create the initials. There's already a check inside the component that when the `id` is not provided it should return the `neutral` color. However, the if clause never allowed an `id` to be undefined so it just would skip that component and return AvatarAnonymous.

image

#### Screenshot after this PR

image

### Breaking changes

none

### DIFF
--- a/src/components/avatar/Avatar.js
+++ b/src/components/avatar/Avatar.js
@@ -49,7 +49,7 @@ class Avatar extends PureComponent {
       );
     }
 
-    if (fullName && id) {
+    if (fullName) {
       return (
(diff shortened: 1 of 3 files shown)


driesd:
I did a small change though. When no id is set, the text color should be neutral darkest.

Screenshot 2020-03-04 13 17 16

kevinwaelkens:
✅ Approve

PR #928, opened 2020-03-17
Link: badged

driesd:
### Description

This PR introduces a `badged` style for our `Link` component.

Screenshot 2020-03-17 19 15 04

### Breaking changes

None.


driesd:
> Is this supposed to replace this? It's something we got from design, but when you hover it, it gets a blue background:
> 
> Screenshot 2020-03-17 at 21 07 31

Correct. I discussed this with @stephen-tl  and this badged link will replace the one of your screenshot.

PR #933, opened 2020-03-19
Avatar stack: overlapping avatars

driesd:
### Description

This PR introduces overlapping avatars in an `AvatarStack`. They stop overlapping when `selectable` prop is `true`.

#### AvatarStack (overlapping by default)
Screenshot 2020-03-19 12 49 07

#### AvatarStack with selectable avatars
Screenshot 2020-03-19 12 49 26

#### AvatarStack with selected avatars
Screenshot 2020-03-19 12 49 47

### Breaking changes

🔥 `AvatarStack`: changed so that the avatars now overlap each other by default, until `selected` is `true`.

### DIFF
--- a/src/components/avatar/Avatar.js
+++ b/src/components/avatar/Avatar.js
@@ -78,14 +78,14 @@ class Avatar extends PureComponent {
   };
 
   render() {
-    const { className, onClick, selected, size, shape, ...others } = this.props;
+    const { className, selectable, selected, size, shape, ...others } = this.props;
 
     const avatarClassNames = cx(
       theme['wrapper'],
       theme[`is-${size}`],
       theme[`is-${shape}`],
       {
-        [theme['is-selectable']]: onClick,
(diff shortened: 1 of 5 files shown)


src/components/avatar/Avatar.js
```diff
@@ -78,14 +78,14 @@ class Avatar extends PureComponent {
   };
 
   render() {
-    const { className, onClick, selected, size, shape, ...others } = this.props;
+    const { className, selectable, selected, size, shape, ...others } = this.props;
```
lowiebenoot: I think you actually fixed a bug "by accident" 😄. Before this change, the `onClick` prop was actually not put on the component, it was only used for the is-selectable class. 

Now that you replaced it by selectable, it's in the `others` and will be spread onto the box later on.

driesd: Omg, this is so true! 🙈 😄

ArnaudWeyts:
👏

PR #940, opened 2020-03-22
Widget: update header height

driesd:
### Description

This PR lowers the height of the `Widget.Header` for all sizes (small, medium & large). Also, in stories, we use `Heading3` as the widget title instead of Heading2 (design specs).

#### Small
Screenshot 2020-03-23 09 40 10

#### Medium
Screenshot 2020-03-23 09 39 43

#### Large
Screenshot 2020-03-23 09 39 32

#### Small
Screenshot 2020-03-23 09 47 41

#### Medium
Screenshot 2020-03-23 09 47 52

#### Large
Screenshot 2020-03-23 09 48 05

### Breaking changes

`Widget headers` might be less high since we decreased vertical padding and switched to min-height & flexbox alignment instead.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,8 @@
 
 ### Changed
 
+- [Breaking] `Widget.Header`: decreased vertical padding and switched to min-height & flexbox alignment. ([@driesd](https://github.com/driesd) in [#940](https://github.com/teamleadercrm/ui/pull/940)
+
 ### Deprecated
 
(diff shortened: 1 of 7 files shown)


driesd:
> In deals-frontend we currently have an "issue" where our widget headers don't have the same height because some of them contain a button and the others don't. The button is higher than the title, so it changes the height.
> 
> Do you think this will solve this issue?

This PR would idd fix that issue :)

PR #953, opened 2020-03-23
Avatar: add AvatarTeam component

driesd:
### Description

This PR adds the `AvatarTeam` component and will be rendered by the `Avatar` component when its `team` prop is `true`.

#### Screenshots
Screenshot 2020-03-23 16 36 28

Screenshot 2020-03-23 16 36 39

### Breaking changes

None.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,9 @@
 
 ### Added
 
+- `Avatar`: added `team` prop (default `false`). ([@driesd](https://github.com/driesd) in [#953](https://github.com/teamleadercrm/ui/pull/953)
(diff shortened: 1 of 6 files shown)


driesd:
> Question (probably more for the designers): I know we want to show a tooltip with the person's name/team name/company name, when hovering over the Avatar. Should that _always_ be the case or not? Because, if it is, then I feel like it would be a good thing to make that abstraction inside of Avatar, since it's then defined as inherent behaviour of the component.
> If not, then I'd like to know the ruleset for that
> 
> Either way, out of scope of this PR, just wanted to raise the question :)

I'll raise this question to our designers :)

PR #971, opened 2020-03-26
Fix: Don't limit onClick on BodyRow children

mikeverf:
### Description

This check was apparently added to prevent the onClick of `BodyRow` to be triggered when clicking on a child that triggers an onClick event.
But that check also means that when you abstract away the children of `BodyRow` into their own component, the children's `onClick` would not be called, because the immediate parent wasn't a `BodyRow`. 

Since that limits the way we compose our components and the issue it's solving is easily fixable by the user by calling `event.stopPropagation` themselves, I propose to get rid of this check.

### Breaking changes

- Removed check that stops `onClick` of `BodyRow` to be triggered when its child `onClick` is triggered. Add `event.stopPropagation()` to the onClick of any children of `BodyRow` to ensure same behaviour.


joaoinacio:
LGTM, I have no idea though if this could affect existing behaviour 🤷‍♂

driesd:
✅ Manually checked

PR #977, opened 2020-03-31
Container component

driesd:
### Description

This PR adds a `Container` component with responsive behavior. By default, its width is fluid. When passing the `fixed` boolean prop, a maximum width will be applied and the container will be centered.

### Fluid
Screenshot 2020-04-01 12 27 26

### Fixed
Screenshot 2020-04-01 12 27 38

### Breaking changes

None.

### DIFF
--- a/src/components/container/Container.js
+++ b/src/components/container/Container.js
@@ -17,7 +17,7 @@ class Container extends PureComponent {
     );
 
     return (
-      <Box {...others} className={classNames}>
+      <Box {...others} boxSizing="content-box" className={classNames}>
         {children}
       </Box>
     );
--- a/src/components/container/Container.js
+++ b/src/components/container/Container.js
@@ -25,6 +25,7 @@ class Container extends PureComponent {
 }
 
 Container.propTypes = {
+  children: PropTypes.node,
   fixed: PropTypes.bool,
 };
 
--- a/src/components/container/Container.js
+++ b/src/components/container/Container.js
@@ -1,0 +1,31 @@
+import React, { PureComponent } from 'react';
+import PropTypes from 'prop-types';
+import Box from '../box';
+import cx from 'classnames';
+import theme from './theme.css';
+
+class Container extends PureComponent {
+  render() {
+    const { children, className, fixed, ...others } = this.props;
+
+    const classNames = cx(
+      theme['container'],
+      {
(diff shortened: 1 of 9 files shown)


kevinwaelkens:
Shouldn't this also have margin-bottom? Cause otherwise it'll just stick to the bottom right? And the `upgrade now` banner will be overlapping it?

driesd:
> Shouldn't this also have margin-bottom? Cause otherwise it'll just stick to the bottom right? And the `upgrade now` banner will be overlapping it?

In a following PR, I'll add a `Page` component which will be responsible for this.

src/components/container/theme.css
```diff
@@ -0,0 +1,19 @@
...
+
+.container {
+  min-width: 990px;
+  padding-left: var(--spacer-big);
+  padding-right: var(--spacer-big);
+}
+
+.is-fixed {
+  margin: 0 auto;
+  max-width: 1080px;
```
lowiebenoot: Why max-width? Now it will still be kind of fluid, no?

src/components/container/Container.js
```diff
@@ -0,0 +1,31 @@
+import React, { PureComponent } from 'react';
+import PropTypes from 'prop-types';
+import Box from '../box';
+import cx from 'classnames';
+import theme from './theme.css';
+
+class Container extends PureComponent {
```
ArnaudWeyts, line 7: PP: don't use PureComponent, it's premature optimisation

lowiebenoot: All the components are PureComponents

src/components/container/theme.css
driesd: You're right @lowiebenoot, `max-width` should equal to `min-width` when the container is fixed. 

Also, the padding should not increase on larger screens because margin: 0 auto; should take care of this. And I also changed the `box-sizing` of the container to be sure the width is always correct regardless of the paddings on each side.

driesd:
> There's no story? I wanted to verify some issues that I had when building the same thing inside our project.

A story is added to demo the container behavior.

src/components/container/Container.js
ArnaudWeyts: Yeah they shouldn't be

PR #1000, opened 2020-04-07
Container: increase the width of the container

driesd:
### Description

This PR increases the min & max-width values from `990px` to `1056px` as mentioned in [this PR](https://github.com/teamleadercrm/ui/pull/977).

### Breaking changes

None.


kevinwaelkens:
PR #1000

PR #1027, opened 2020-04-20
Link (badged): visual state updates

driesd:
### Description

Besides **adjusting** the `visual states` of our `badged Link` component,  this PR is also **adding** a `selected` prop, which renders the `Link in a selected state` when `true`.

#### Screenshot after this PR
Screenshot 2020-04-20 10 03 27

### Breaking changes

None.


driesd:
> Wondering what the use case is for the selected state though.

Good question! I checked this with our designers and for now, it's mainly because of consistency across other button-like components. But they already have future use-cases in mind...

PR #1109, opened 2020-05-18
SilentBanner: component

driesd:
### Description

This PR adds the new `SilentBanner` component.
**Small disclaimer:** the name of this new component is temporary and will be changed.

#### Basic
Screenshot 2020-05-18 16 38 55

#### Dismissable
Screenshot 2020-05-18 16 39 20

#### With actions
Screenshot 2020-05-18 16 39 31

#### With status
Screenshot 2020-05-18 16 39 43

#### With status and icon
Screenshot 2020-05-18 16 42 28

#### Full option
Screenshot 2020-05-18 16 39 53

### Breaking changes

None.

### DIFF
--- a/src/components/silentBanner/SilentBanner.js
+++ b/src/components/silentBanner/SilentBanner.js
@@ -1,0 +1,135 @@
+import React, { PureComponent } from 'react';
+import PropTypes from 'prop-types';
+import {
+  IconBellMediumOutline,
+  IconCheckmarkBadgedMediumFilled,
+  IconCloseMediumOutline,
+  IconWarningBadgedMediumFilled,
+} from '@teamleader/ui-icons';
+import Box from '../box';
+import Button from '../button';
+import ButtonGroup from '../buttonGroup';
+import Icon from '../icon';
+import { IconButton } from '../iconButton';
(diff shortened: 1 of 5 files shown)


src/components/silentBanner/SilentBanner.js
```diff
@@ -0,0 +1,150 @@
...
+            )}
+            {children && <TextBody color="teal">{children}</TextBody>}
+            {hasActions && (
+              <ButtonGroup display="flex" justifyContent="flex-end" marginTop={4}>
+                {secondaryAction && (
+                  <Button level="link" size="small">
+                    {secondaryActionLabel}
+                  </Button>
+                )}
+                {primaryAction && <Button size="small">{primaryActionLabel}</Button>}
```
lowiebenoot, line 114: Not very handy if we can only set the label 😉 

You can use the same approach as in the Dialog component.

```js
{secondaryAction && <Button level="link" size="small" {...secondaryAction} />}
{primaryAction && <Button level="primary" size="small" {...primaryAction} />}
```

This way you also don't need a separate prop (primaryAction & primaryActionLabel)

PR #1144, opened 2020-06-02
Container: fixed & fluid padding should be consistent

driesd:
### Description

This PR makes sure that our `Container` paddings are consistent for fluid and fixed on each breakpoint.

### Breaking changes

None.


ArnaudWeyts:
✅ Manually checked

PR #1160, opened 2020-06-15
Scrollable tabs

driesd:
### Description

This PR makes our `TabGroup` component scrollable when tabs are horizontally overflowing their container. It is also possible to scroll with the mouse wheel because of the `overflowX="scroll"` prop. The scrollbar is hidden with CSS.

**Tested in:**
- Chrome
- Firefox
- Safari
- Opera
- IE11

#### Medium sized
Screen Recording 2020-06-15 at 16 27 51

#### Small sized
Screen Recording 2020-06-15 at 16 28 08

#### Resize behavior
Screen Recording 2020-06-15 at 16 28 38

### Breaking changes

None.

### Coming up, 2 PR's
- `TabGroup` will auto-scroll to the active `TitleTab`
- `IconTab` will be removed

### DIFF
--- a/src/components/tab/TabGroup.js
+++ b/src/components/tab/TabGroup.js
@@ -1,5 +1,6 @@
 import React, { createRef, PureComponent } from 'react';
 import ReactResizeDetector from 'react-resize-detector';
+import smoothScroll from 'smoothscroll-polyfill';
 import PropTypes from 'prop-types';
 import Box from '../box';
 import IconButton from '../iconButton';
@@ -18,6 +19,7 @@ class TabGroup extends PureComponent {
   };
 
   componentDidMount() {
+    smoothScroll.polyfill();
     this.scrollContainerRef.current.boxNode.current.addEventListener('scroll', this.handleScroll);
   }
 
--- a/src/components/tab/TabGroup.js
+++ b/src/components/tab/TabGroup.js
@@ -5,14 +5,26 @@ import cx from 'classnames';
 import theme from './theme.css';
 
 class TabGroup extends PureComponent {
+  scrollContainerRef = createRef();
+
   render() {
     const { children, className, inverted, size, ...others } = this.props;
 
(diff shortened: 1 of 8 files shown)


driesd:
> Nice approach with using the native scroll.
> 
> Only thing I'm a bit hesitant about is the fixed scrolling distance of 200. Maybe we can make it a bit more dynamic and use something like 50% of the scroll container (with a minimum of 200).
> I think if you use 200, you will always have to click multiple times if there are a few items hidden by the scroll. If you use half of the container, it will probably show more hidden items at once.
> 
> Just an idea that I have, certainly not blocking.

Valid concern indeed. I'll tackle this in my follow-up PR, which will handle auto-scroll to the active tab.

src/components/tab/TabGroup.js
```diff
@@ -1,18 +1,95 @@
...
+  componentWillUnmount() {
+    this.scrollContainerRef.current.boxNode.current.removeEventListener('scroll', this.handleScroll);
+  }
+
+  checkForScrollPosition = () => {
+    const { scrollLeft, scrollWidth, clientWidth } = this.scrollContainerRef.current.boxNode.current;
+
+    this.setState({
+      canScrollLeft: scrollLeft > 0,
+      canScrollRight: scrollLeft !== scrollWidth - clientWidth,
```
lorgan3, line 35: I think you should use a `<` instead,  I think you can scroll further than the width on mac. I'll share a video with you on slack

driesd:
> Looks good 👍
> 
> I found 1 small issue / weird behaviour and I have 1 design suggestion: I feel like the buttons should have a little bit of transparency or a fade animation because it's a little weird when they pop in when manually scrolling

Good suggestion, I'll discuss this with our designers!

src/components/tab/TabGroup.js
driesd: Will try to fix in my follow-up PR. Thanks for the manual check!

PR #1173, opened 2020-06-18
Menu: let the component grow inside of its container

ArnaudWeyts:
### Description

- The menu component now grows to the width of its container

Screenshot 2020-06-18 at 15 34 24

#### Screenshot after this PR

Screenshot 2020-06-18 at 15 35 41

### Breaking changes

- None

### Manual check

On base branch:

- On the popover with menu story, open up the menu, and remove the last item in the inspector
- You'll notice the menu shrinks to a certain min width but its items are allowed to shrink even more, causing the selected states and highlights to be too small for its container

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,8 @@
 
 ### Changed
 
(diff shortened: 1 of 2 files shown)


ketlane:
Yes please! :)

PR #1231, opened 2020-07-15
[guild] [experiment] Better typescript support

lorgan3:
### Description

This is an experiment to see how hard it to improve the support for TypeScript. There don't appear to be any (working / usable) tools to generate proper declaration files from a non typescript source so I went the other way around and enabled typescript in the library.

I configured ts to not be very strict, so almost all code keeps working with no changes required.

### Pros & cons
✅ no need to keep a `.d.ts` file in sync. This gets generated automatically and all exported components are included with no additional configuration.
✅ with very little work the existing components can be converted to ts, which results in proper typehinting for the props.
✅ I'm not noticing any regression in the storybook (Though I do see some issues that exist in the current version also)

😐 Some files I HAD to convert to ts (the ones using `forwardRef` and functions that return private objects)
😐 Errors are pretty hard to read because the types are inferred from the proptypes.
😐 Even though it should work according to [the library](https://github.com/taehwanno/props-type#optional-without-isrequired) I used, I couldn't differentiate between required and optional props so I made everything optional for now.

So overal I don't think there are any big blockers and adopting this change would already provide a (small) benefit.
If you want typehinting for all components they should be converted to ts like I did with the Box / Text / ... components 

I tried using it in the `deals-frontend` project using `npm link` and besides some glitches because of it (I think `isComponentOfType` checks don't work when using `npm link`) it was working 💪 
image

Let me know what you think!


src/components/hoc/withTheme/withTheme.tsx
```diff
@@ -0,0 +1,46 @@
...
+  className: PropTypes.string,
+};
+
+const defaultProps = {
+  color: 'neutral',
+  size: 'medium',
+  tint: 'normal',
+};
+
+type Props = Partial<PropsType<typeof props, typeof defaultProps>>;
```
ArnaudWeyts, line 24: I'm not entirely sure I'm following on why we need to use this library and can't just define the types in typescript itself 🤔

lorgan3: Yes you can definitely do that! I used this library because this way you don't have to rewrite the proptypes so this should be easier to do.

Storybook also uses the proptypes to generate the documentation so we would have to use something like this to generate them again https://github.com/milesj/babel-plugin-typescript-to-proptypes
☝️ The resulting docs were not as 'good' though, the enums that show all options on hover just became the word `union` when I gave it a try.

lorgan3:
@lowiebenoot I meant in the types. I'm wrapping them in a `Partial<>` like here https://github.com/teamleadercrm/ui/pull/1231/files#diff-e6b1bc1aeb592793ebf47f20713ccee0R24

(this doesn't change the behaviour, just the static type checking)

I'll create a second branch where I'll write an interface instead of converting the proptypes, that should address the 2nd and 3rd con. The tradeoff is that it's more work.

lorgan3:
I tried using interfaces instead of proptypes but I struggled a lot getting storybook to detect these. I got it kind of working but it's not working very well (comments on the props are only sometimes working, types are not as accurate)

Here's a branch if you want to see it for yourself https://github.com/teamleadercrm/ui/pull/new/guild-experiment-typescript-v2

It looks like storybook 6 has ts support out of the box with might work better

PR #1239, opened 2020-07-27
Storybook 6: replace addon-info with addon-docs

driesd:
### Description

This PR replaces the deprecated `addon-info` with the new `addon-docs`.

### Breaking changes

None.


driesd:
> I find it kind of confusing to use with tabs on the left and the right 😕

The `Knobs`, as we know them in the panel on the right, will be removed and entirely replaced by the new Storybook `Controls`.

driesd:
> Is there a way to show multiple components on 1 page? For a couple components that have child components it's not possible to see the props (like LabelValuePair.Label, LabelValuePair.Value, ...)

Working on this, but there's a chance we'll have to get rid of these dot notation components... 🤷‍♂️

driesd:
@lorgan3 I'll make sure the warning goes away in a follow-up PR 🙂

PR #1262, opened 2020-08-13
[CHORE] Bumped kind-of to 6.0.3

nielsvermaut:


ArnaudWeyts:
Closing since dependabot should create a PR for this instead.

PR #1264, opened 2020-08-13
[CHORE] Bumped prismjs to 1.21

nielsvermaut:


ArnaudWeyts:
Closed since dependabot should create a PR for this instead.

PR #1296, opened 2020-09-07
[CHORE] Remove "@storybook/addon-docs" as Dependency, and add it as DevDependency

sanderbrugge:
### Description

- The newest version of "@storybook/addon-docs" is requiring a lot of dependencies, which are not needed. They now pollute `yarn.lock`-files from repositories that use this package with unnecessary dependencies.
- It's also throwing vulnerability reports that aren't needed on consuming repo's.

### Breaking changes

- This should not be a breaking change.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,8 @@
 
 ### Changed
 
+- `Dependencies`: Set "@storybook/addon-docs" as a DevDependency instead of a reguler dependency.
+
 ### Deprecated
 
 ### Removed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@
 
 ### Changed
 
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -4,6 +4,8 @@
 
 ### Changed
 
+- `Dependencies`: Set "@storybook/addon-docs" as a DevDependency instead of a reguler dependency.
```
ArnaudWeyts, line 7: Small typo 🙈 

```suggestion
- `Dependencies`: Set "@storybook/addon-docs" as a DevDependency instead of a regular dependency.
```

ArnaudWeyts (approved):

mikeverf (approved):

lowiebenoot (approved):

driesd (approved):

PR #1361, opened 2020-12-08
Dialog: remove redundant wrapper in dialog rendering implementation

ArnaudWeyts:
### Description

The `DialogBase` component creates a root div (which is empty) for some reason when rendering itself (trough a React portal). This causes issues since the actual overlay is then a child of that div, and the `position: fixed` set on it is relative to it's containing block (the root div).

### Manual check

- The dialog still mounts and unmounts

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,8 @@
 
 ### Changed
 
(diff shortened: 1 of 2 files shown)


timdegroote (approved):

driesd (approved):

driesd:
✅  Manually checked

PR #1488, opened 2021-02-11
Remove husky as a dependency as we don't use that anymore

sanderbrugge:
### Description

- Removed the `husky` dependency as we don't use this anymore. see #1487


driesd (approved):

lorgan3 (approved):

ArnaudWeyts (changes requested):
🤔 We do use it though, there's a `.huskyrc` file in the root directory

driesd:
@ArnaudWeyts 's got a point here... ☝️  Closing this PR.

PR #1490, opened 2021-02-12
eslint-config-standard-jsx should be a dev dependency

lowiebenoot:
`eslint-config-standard-jsx` got installed as a dependency, while it should be a dev dependency.


lorgan3 (approved):
Thanks!

ArnaudWeyts (approved):

sanderbrugge (approved):

PR #1566, opened 2021-03-17
Remove dependency on `@sambego/storybook-state`

timdegroote:
### Description

The `@sambego/storybook-state` package still loaded in previous versions of several storybook packages, packages which also rely on React 16. This package also doesn't seem that actively maintained anymore, so an update doesn't seem close on the horizon. Since its functionality can relatively easily be replaced by using `useState`, it makes sense to remove it.

By removing storybook-state we now only load the storybook package versions which `@teamleader/ui` lists (lots of packages are removed from `yarn.lock`) and will make the switch to React 17 easier in the future.

### DIFF
--- a/.storybook/stories/playground/dialogs.stories.js
+++ b/.storybook/stories/playground/dialogs.stories.js
@@ -1,6 +1,5 @@
-import React from 'react';
+import React, { useState } from 'react';
 import { addStoryInGroup, PLAYGROUND } from '../../utils';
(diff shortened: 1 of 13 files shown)


lorgan3 (approved):

JorenSaeyTL (approved):

sanderbrugge (approved):

driesd (approved):

ArnaudWeyts (approved):
🎖️ Great job

PR #1617, opened 2021-04-21
Input: remove width 100% from the `.wrapper`.

driesd:
This PR removes the `width: 100%` from the component's wrapper because this is causing troubles with flexbox layout.

### Screenshot before this PR
Screenshot 2021-04-21 at 13 07 59

### Screenshot after this PR
Screenshot 2021-04-21 at 13 08 19

### Breaking changes

Should not break anything 🤞

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@
 
 - :boom: `CompactMessage`: removed component. ([@driesd](https://github.com/driesd) in [#1612])
 - :boom: `Message`: removed component. ([@driesd](https://github.com/driesd) in [#1612])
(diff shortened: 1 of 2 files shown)


lorgan3 (approved):

timdegroote (approved):

EOSullivanBerlin (approved):

lowiebenoot:
Quite a high chance that this will break some layouts, no?

driesd:
Well, I did some quick testing on `Input` components used in `core`, by overriding the `.wrapper`'s `width: 100%` with `width: auto`. I didn't see layouts breaking. Of course, there will always be those few cases which we overlooked... 🤷‍♂️

PR #1646, opened 2021-05-18
Introduce `DurationInput` & improvements to the `NumericInput`

ArnaudWeyts:
### Description

This PR introduces a new `DurationInput` component. It consists of an hour and minute NumericInput

Screenshot 2021-05-18 at 11 23 38

### Added

- `DurationInput` component

### Changed

- `NumericInput`: Add support for clicking and holding on the steppers to increment / decrement
- `NumericInput` [_Internal_]:  removed redundant internal state

### Breaking changes

- None

### Manual check

- The NumericInput allows clicking and holding the increment / decrement buttons
- Test out the DurationInput component

### DIFF
--- a/src/components/input/DurationInput.js
+++ b/src/components/input/DurationInput.js
@@ -4,11 +4,11 @@ import { Box, NumericInput } from '../../';
 const DurationInput = ({ value, onChange, onKeyDown }) => {
   return (
     <Box display="flex" alignItems="center">
-      <NumericInput placeholder="00" min={0} onKeyDown={onKeyDown} size="small" />
+      <NumericInput placeholder="00" min={0} onKeyDown={onKeyDown} size="small" flexGrow={1} />
       <Box element="span" marginHorizontal={2}>
         :
       </Box>
-      <NumericInput placeholder="00" min={0} onKeyDown={onKeyDown} size="small" />
+      <NumericInput placeholder="00" min={0} onKeyDown={onKeyDown} size="small" flexGrow={1} />
     </Box>
   );
 };
--- a/src/components/input/DurationInput.js
+++ b/src/components/input/DurationInput.js
@@ -11,11 +11,30 @@ const DurationInput = ({ value, onChange, onKeyDown }) => {
   };
 
   const handleMinutesChange = (_, minutes) => {
-    onChange &&
+    if (!onChange) {
+      return;
+    }
+
+    if (minutes >= 60) {
       onChange({
-        ...value,
-        minutes,
+        hours: (value?.hours || 0) + 1,
+        minutes: 0,
       });
+      return;
+    }
+
+    if (minutes < 0) {
+      onChange({
+        hours: (value?.hours || 1) - 1,
+        minutes: 59,
+      });
+      return;
+    }
+
+    onChange({
+      ...value,
+      minutes,
+    });
   };
 
   return (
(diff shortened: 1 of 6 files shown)


ArnaudWeyts:
⚠️ Still need to implement the onBlur handler correctly

lowiebenoot (approved):

src/components/input/DurationInput.js
```diff
@@ -11,19 +11,38 @@ const DurationInput = ({ value, onChange, onKeyDown }) => {
...
-        minutes,
+        hours: (value?.hours || 0) + 1,
+        minutes: 0,
+      });
+      return;
+    }
+
+    if (minutes < 0) {
+      onChange({
+        hours: (value?.hours || 1) - 1,
```
lorgan3, line 56: Should it also make sure to not output negative hours here?

src/components/input/DurationInput.js
```diff
@@ -34,8 +53,8 @@ const DurationInput = ({ value, onChange, onKeyDown }) => {
       </Box>
       <NumericInput
         placeholder="00"
-        min={0}
-        value={value?.minutes || ''}
+        {...(!value?.hours ? { min: 0 } : {})}
```
lorgan3, line 91: What is this for? 🤔

src/components/input/DurationInput.js
ArnaudWeyts: It will actually take it into account, if `value.hours = 0`, it will fall back to 1 🙈 Bit weird to read though so I understand the confusion

src/components/input/DurationInput.js
ArnaudWeyts: This makes sure that you can't go below 0 when the hour counter is 0. (If you have `1 hour`, you can decrement the minute picker and it will subtract the hour and set the minutes to `59`)

src/components/input/DurationInput.js
lorgan3: right I see 😅

sanderbrugge (approved):

src/components/input/DurationInput.js
```diff
@@ -0,0 +1,16 @@
+import React from 'react';
+import { Box, NumericInput } from '../../';
+
+const DurationInput = ({ value, onChange, onKeyDown }) => {
```
jnstr, line 6: Is this the best fitting name for this component? 

For the current use case, bulk time tracking modal, this will be used to set the duration. But in the future, this might be used to set the start and end time of something. In that case, it's not a duration anymore.
On the other hand, keeping this name might prevent misusing this component for things it isn't meant for.

jnstr (approved):
Nice work! :+1:

src/components/input/DurationInput.js
ArnaudWeyts: Good question, for the start and end time the `TimeInput` component should be used, this would then be limited to 24h, where as this one isn't

jnstr: makes sense! :+1:

lorgan3 (approved):

lowiebenoot (approved):
Should you also implement an onFocus?

ArnaudWeyts:
> Should you also implement an onFocus?

Yes! I'll do that in a follow-up since I've discovered some more missing things while implementing the component

PR #1652, opened 2021-05-20
`DurationInput`: Consistent empty values

ArnaudWeyts:
### Fixed

- `DurationInput` would resort to a weird `{ hours: undefined }` or `{ minutes: undefined }` value when clearing either of the values

### DIFF
--- a/src/components/input/DurationInput.js
+++ b/src/components/input/DurationInput.js
@@ -13,6 +13,11 @@ const DurationInput = ({ value, onChange, onBlur, onFocus, onKeyDown, autoFocus,
     }
 
     if (hours === '') {
+      if (typeof value?.minutes === 'undefined') {
+        onChange();
+        return;
+      }
+
       onChange({
         minutes: value?.minutes,
       });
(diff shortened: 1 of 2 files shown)


jnstr (approved):

driesd (approved):

src/components/input/DurationInput.js
```diff
@@ -13,9 +13,15 @@ const DurationInput = ({ value, onChange, onBlur, onFocus, onKeyDown, autoFocus,
     }
 
     if (hours === '') {
+      if (typeof value?.minutes === 'undefined') {
+        onChange();
```
lowiebenoot, line 17: Can you explain why it's triggering an onChange with no params? Don't really understand what's going on here.

lowiebenoot (approved):

lorgan3 (approved):

src/components/input/DurationInput.js
ArnaudWeyts: Here's the different states of the value:

```js
// No values in either hours or minutes
value: undefined

// No value in hours
value: {
  minutes: 30
}

// No value in minutes
value: {
  hours: 1
}

// Value in both minutes and hours
value: {
  hours: 1
  minutes: 30
}
```

Because each input can be cleared independently, we want to pass no object at all if each input is empty

PR #1659, opened 2021-06-01
FIX: Make stepper events trigger the onChange event

JorenSaeyTL:
### Description

I did this so that the stepper can be used when
passing native value and onChange props to
the NumericInput

For example a form library that passes value and onChange
to this component will now be able to work with the
stepper out of the box.

Current behavior is still supported (onChange(event, value)).

- Same styling though so N/A

#### Screenshot after this PR

- Same styling though so N/A

### Breaking changes

- Current behavior should still be supported onChange(event, value) instead of onChange(event)

Based on this https://stackoverflow.com/questions/42550341/react-trigger-onchange-if-input-value-is-changing-by-state =>
**Solution Working in the Year 2020 and 2021:**

Also in this post top answer: https://stackoverflow.com/questions/23892547/what-is-the-best-way-to-trigger-onchange-event-in-react-js

### Manual check
 
- NumericInput => typing in the input field still works with (onChange(event, value)) (both event and value are present)
- NumericInput =>  Clicking the stepper now works with onChange(event, value) (both event and value are present, instead of event being null)

### DIFF
--- a/src/components/input/NumericInput.js
+++ b/src/components/input/NumericInput.js
@@ -181,7 +181,7 @@ class NumericInput extends PureComponent {
 
     return (
       <SingleLineInputBase
-        innerRef={this.inputElement}
+        ref={this.inputElement}
         type="number"
         value={value}
         onChange={this.handleOnChange}
--- a/src/components/input/NumericInput.js
+++ b/src/components/input/NumericInput.js
@@ -40,9 +40,22 @@ class StepperControls extends PureComponent {
 class NumericInput extends PureComponent {
   constructor(props) {
     super(props);
+    this.inputElement = React.createRef();
     this.timer = React.createRef();
     this.timeout = React.createRef();
   }
+
+  setNativeValue = (element, value) => {
+    const valueSetter = Object.getOwnPropertyDescriptor(element, 'value').set;
+    const prototype = Object.getPrototypeOf(element);
+    const prototypeValueSetter = Object.getOwnPropertyDescriptor(prototype, 'value').set;
+
+    if (valueSetter && valueSetter !== prototypeValueSetter) {
+      prototypeValueSetter.call(element, value);
+    } else {
+      valueSetter.call(element, value);
+    }
+  };
 
   handleOnChange = (event) => {
     const { onChange } = this.props;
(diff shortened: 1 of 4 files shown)


src/components/input/SingleLineInputBase.js
```diff
@@ -79,6 +80,7 @@ class SingleLineInputBase extends PureComponent {
       onBlur: this.handleBlur,
       onFocus: this.handleFocus,
       readOnly,
+      innerRef,
```
ArnaudWeyts, line 83: Strong PP: Implement ref forwarding similarly how it's done in other places, by using a forwardRef, so we can just pass `ref` to the component:

Example:
https://github.com/teamleadercrm/ui/blob/287612a6ffb295e628ede33fe41c7ef2a058457d/src/components/box/Box.js#L209

JorenSaeyTL: Jep also like that :) will do

ArnaudWeyts (approved):

driesd (approved):

ArnaudWeyts (approved):

src/components/input/NumericInput.js
```diff
@@ -50,13 +51,18 @@ class NumericInput extends PureComponent {
...
     const newValue = parseValue(currentValue + step * n, min, max);
 
-    if (newValue !== currentValue) {
-      onChange && onChange(null, newValue);
+    const inputElement = this.inputElement.current;
+    if (inputElement) {
+      const prototype = Object.getPrototypeOf(inputElement);
+      const prototypeValueSetter = Object.getOwnPropertyDescriptor(prototype, 'value').set;
+
+      prototypeValueSetter.call(inputElement, newValue);
```
lorgan3, line 64: This might be a stupid question but why do you need to do it this way?

JorenSaeyTL (commented):
Just setting `inputElement.value = newValue` doesn't work anymore in React so we need to do it like this, Better explanation in the link above.

The problem is normally a native <input type="number" step="1"/> also dispatches the onChange event when stepping, but because we use our own buttons (for styling purposes) we only passed the value to the onChange function. But react and most libraries expect an event as the first param of the onChange function.

sanderbrugge (approved):

lorgan3 (approved):
Did you add the links later or was I blind? Anyways thanks for the explanation 😅

PR #1672, opened 2021-06-07
Fix Toggle layout

ArnaudWeyts:
### Description

When using the Toggle component inside of a flex container, the track (slider control) wasn't taking up the space it should take up (30px for small), it looks like the `flex-basis` property doesn't suffice, so I just added explicit width for all sizes as well. I've also set the input to hidden to prevent the extra 4 pixels before the slider seen in the screenshot

### Fixed

- Toggle takes up the correct amount of space in a flex container
- Toggle input that's used under the hood is hidden

Screenshot 2021-06-07 at 15 40 03

#### Screenshot after this PR

Screenshot 2021-06-07 at 15 42 40

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,8 @@
 
 ### Fixed
 
+- `Toggle`: Fix layout ([@ArnaudWeyts](https://github.com/ArnaudWeyts) in [#1672])
+
 ### Dependency updates
 
 ## [6.2.0]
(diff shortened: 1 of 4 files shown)


driesd (approved):

EOSullivanBerlin:
✅ Looks good ✅

jnstr (approved):

EOSullivanBerlin (approved):

PR #1689, opened 2021-06-16
Add `max` and `error` props to `DurationInput`

jnstr:
### Added
- `DurationInput`: Added `error` property
- `DurationInput`: Added `max` property

Screenshot 2021-06-14 at 13 40 37

#### Screenshot after this PR

image

### Breaking changes

- none

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,8 @@
 ## [unreleased]
 
 ### Added
+- `DurationInput`: Added `error` property ([@jnstr](https://github.com/jnstr) in [#1689] 
+- `DurationInput`: Added `max` property ([@jnstr](https://github.com/jnstr) in [#1689]
 
 ### Changed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,8 +1,9 @@
 ## [unreleased]
 
 ### Added
-- `DurationInput`: Added `error` property ([@jnstr](https://github.com/jnstr) in [#1689] 
-- `DurationInput`: Added `max` property ([@jnstr](https://github.com/jnstr) in [#1689]
+
+- `DurationInput`: Added `error` property ([@jnstr](https://github.com/jnstr) in [#1689](https://github.com/teamleadercrm/ui/pull/1689))
+- `DurationInput`: Added `max` property ([@jnstr](https://github.com/jnstr) in [#1689](https://github.com/teamleadercrm/ui/pull/1689))
 
 ### Changed
 
--- a/src/components/input/DurationInput.js
+++ b/src/components/input/DurationInput.js
@@ -183,6 +183,4 @@ DurationInput.propTypes = {
   max: PropTypes.number,
 };
 
-DurationInput.defaultProps = {};
-
 export default DurationInput;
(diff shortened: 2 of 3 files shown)


src/components/input/DurationInput.js
```diff
@@ -1,12 +1,13 @@
...
 import { Box, NumericInput } from '../../';
 import theme from './theme.css';
+import cx from "classnames";
 
 const transformToPaddedNumber = (number) => (number < 10 ? `0${number}` : number.toString());
 
 const MINUTES_STEP = 15;
 
-const DurationInput = ({ value, onChange, onBlur, onFocus, onKeyDown, autoFocus, textAlignRight, className }) => {
+const DurationInput = ({ value, onChange, onBlur, onFocus, onKeyDown, autoFocus, textAlignRight, className, error, max }) => {
```
jnstr, line 10: Q: How could we make it more clear the max is in minutes? `maxMinutes` could cause confusion.

lorgan3: I think for this one an explanation in storybook would be the best solution
see https://github.com/teamleadercrm/ui/blob/next-release/src/components/input/Input.js#L42 -> https://components.teamleader.design/?path=/docs/4-low-level-blocks-form-elements-input--input

Though it seems only the main component in the story has this, that might be a configuration issue.

Otherwise you could maybe opt for `maxInMinutes`

lorgan3 (approved):

src/components/input/DurationInput.js
ArnaudWeyts: These both sound like reasonable solutions, I think I would prefer just going for the documentation option

src/components/input/DurationInput.js
```diff
@@ -140,6 +149,7 @@ const DurationInput = ({ value, onChange, onBlur, onFocus, onKeyDown, autoFocus,
         placeholder="00"
         step={MINUTES_STEP}
         {...(!value?.hours ? { min: 0 } : {})}
+        {...(maxHours && value?.hours && value.hours >= maxHours ? { max: maxMinutes } : {})}
```
ArnaudWeyts, line 152: Not sure I'm following with this one. Doesn't this mean that if you have a max with less than 59 minutes (thus being able to not fill in the hours), the max will never matter?

jnstr: Oh, you're right. Will fix!

CHANGELOG.md
```diff
@@ -1,6 +1,8 @@
 ## [unreleased]
 
 ### Added
+- `DurationInput`: Added `error` property ([@jnstr](https://github.com/jnstr) in [#1689] 
+- `DurationInput`: Added `max` property ([@jnstr](https://github.com/jnstr) in [#1689]
```
lorgan3, line 5:
```suggestion
- `DurationInput`: Added `error` property ([@jnstr](https://github.com/jnstr) in [#1689](https://github.com/teamleadercrm/ui/pull/1689)) 
- `DurationInput`: Added `max` property ([@jnstr](https://github.com/jnstr) in [#1689](https://github.com/teamleadercrm/ui/pull/1689))
```

lorgan3: Just noticed this 😅

src/components/input/DurationInput.js
ArnaudWeyts: We've added a comment, but there seems to be something wrong with the docblocks rendering, will look at it in a different PR

ArnaudWeyts (approved):

ArnaudWeyts:
✅ Manually checked

PR #1690, opened 2021-06-16
Clean up build process

ArnaudWeyts:
### Description

This is a draft PR to clean up some issues with the configured build process. I've also manually checked this by locally linking the package.

### Fixed

- Infinite building for storybook due to incorrect story path configuration
- Babel warnings when building for storybook
- PostCSS deprecation warning when building for storybook

### Manual check

- [x] Storybook builds without issues (`yarn start`)
- [x] A build of the UI-library still works when locally linking it

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -2,7 +2,6 @@
   "presets": ["@babel/preset-env", "@babel/preset-react"],
   "plugins": [
     "@babel/plugin-proposal-export-default-from",
-    "@babel/plugin-proposal-class-properties",
     "@babel/plugin-transform-runtime"
   ],
   "env": {
--- a/.babelrc
+++ b/.babelrc
@@ -5,9 +5,6 @@
     "@babel/plugin-transform-runtime"
   ],
   "env": {
-    "development": {
-      "plugins": ["@babel/plugin-transform-react-jsx-source"]
-    },
     "cjs": {
       "plugins": ["transform-react-remove-prop-types"],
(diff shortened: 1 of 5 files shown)


jnstr (approved):

lowiebenoot (approved):

lorgan3 (approved):

lowiebenoot (commented):
Not sure how relevant it actually is, but I get an error when trying to render a Select component.

```
Uncaught Error: Invalid hook call. Hooks can only be called inside of the body of a function component. This could happen for one of the following reasons:
1. You might have mismatching versions of React and the renderer (such as React DOM)
2. You might be breaking the Rules of Hooks
3. You might have more than one copy of React in the same app
See https://fb.me/react-invalid-hook-call for tips about how to debug and fix this problem.
    React 2
        resolveDispatcher
        useContext
    withEmotionCache emotion-element-a8309070.browser.esm.js:26
    React 6
        renderWithHooks
        updateForwardRef
        beginWork
        callCallback
        invokeGuardedCallbackDev
        invokeGuardedCallback
```

ArnaudWeyts:
https://github.com/teamleadercrm/ui/pull/1690#pullrequestreview-687109964

@lowiebenoot Yup, that's because of the different react instances between UI and core, we can have a call on how to properly test it

sanderbrugge (approved):

driesd (approved):

PR #1704, opened 2021-06-28
Fix duration input overlapping other elements

ArnaudWeyts:
I haven't applied this change to any of the other inputs yet since I'm unsure what the impact will be.

### Fixed

- DurationInput conflicting with existing stacking contexts

### Manual check

Let's zoom for this one.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,8 @@
 
 ### Fixed
 
+- `DurationInput`: Fix duration input overlapping other elements ([@ArnaudWeyts](https://github.com/ArnaudWeyts) in [#1704](https://github.com/teamleadercrm/ui/pull/1704))
+
 ### Dependency updates
 
 ## [6.5.0] - 2021-06-18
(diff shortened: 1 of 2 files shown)


jnstr (approved):

lowiebenoot (commented):
I wonder why it needs a `z-index: 1;` on hover (a few lines above your change).

EOSullivanBerlin (approved):

adognameddog (approved):

ArnaudWeyts:
> I wonder why it needs a `z-index: 1;` on hover (a few lines above your change).

@lowiebenoot  Because of the way they are positioned, e.g. the second input overlaps the first one by 1 pixel (to make sure the middle border is aligned), but when the first input is hovered, we want to make sure the focus state isn't overlapped

ArnaudWeyts:
I changed my mind, let's manually check this where it's implemented.

lowiebenoot (approved):

PR #1717, opened 2021-06-30
[CORE-4804] Filter selection component

JorenSaeyTL:
### Description

Added a FilterSelection component

#### Screenshot after this PR

Schermafbeelding 2021-06-30 om 16 21 22

For the other states check storybook locally

### DIFF
--- a/src/components/filterSelection/FilterSelection.js
+++ b/src/components/filterSelection/FilterSelection.js
@@ -20,8 +20,26 @@ export const Status = {
   BROKEN: 'broken',
 };
 
-const FilterSelection = ({ label, modificationText, status, amount, onClick, onClearClick }, ref) => {
-  const showAmount = typeof amount === 'number';
+const ColorByStatus = {
+  [Status.DEFAULT]: 'neutral',
+  [Status.ACTIVE]: 'aqua',
+  [Status.FOCUSED]: 'aqua',
+  [Status.DISABLED]: 'neutral',
+  [Status.INVALID]: 'gold',
+  [Status.BROKEN]: 'ruby',
+};
+
+const BackgroundTintByStatus = {
+  [Status.DEFAULT]: undefined,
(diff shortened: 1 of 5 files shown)


driesd (approved):

src/components/filterSelection/FilterSelection.js
```diff
@@ -0,0 +1,13 @@
...
+
+const FilterSelection = () => {
+  return <Box>Hello FilterSelection!</Box>;
+};
+
+FilterSelection.propTypes = {};
+
+FilterSelection.defaultProps = {};
+
+export default memo(FilterSelection);
```
ArnaudWeyts: Let's not memoize components by default, I know we used to do this by default using PureComponents as well but this isn't a standard thing to default to, I've said this several times before, as it can worsen performance. Build first, optimise later 😄

src/components/filterSelection/FilterSelection.js
```diff
@@ -13,7 +18,70 @@ export const Status = {
 };
 
 const FilterSelection = ({ label, status, amount, onClick, onClearClick }, ref) => {
-  return <Box ref={ref}>Hello FilterSelection!</Box>;
+  const handleClearClick = useCallback(
```
ArnaudWeyts: Again, don't memoize be default, you should only memoize inline functions if they turn out to be really expensive. It can be cheaper just to create functions every time than to memoize them

ArnaudWeyts (approved):
Doesn't seem like this component is keyboard accessible in storybook. Are you planning on following up on that?

src/components/filterSelection/FilterSelection.js
JorenSaeyTL: Sounds like a good rule 👍

JorenSaeyTL:
Gonna make it keyboard accessible in a follow up 👍 since its a new component it shouldn't be breaking :)

PR #1721, opened 2021-07-01
Fix update filter selection

JorenSaeyTL:
### Description

Visually nothing changed.

Added keyboard accessibility and removed an usused style,
Also stop propagation of event when clicking the clear button

### DIFF
--- a/src/components/filterSelection/FilterSelection.js
+++ b/src/components/filterSelection/FilterSelection.js
@@ -52,6 +52,13 @@ const FilterSelection = ({ label, modificationText, applied, status, amountAppli
       }
     : {};
 
+  const handleClearClick = (event) => {
+    event.stopPropagation();
+    if (status !== Status.DISABLED) {
+      onClearClick();
+    }
+  };
+
   return (
     <Box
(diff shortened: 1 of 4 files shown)


driesd (approved):

src/components/filterSelection/FilterSelection.js
```diff
@@ -63,6 +70,8 @@ const FilterSelection = ({ label, modificationText, applied, status, amountAppli
     <Box
       ref={ref}
       className={cx(theme['select-control'])}
+      role="button"
+      tabIndex={0}
```
lorgan3, line 74: Q: Would it make sense to just make this element a button instead of a div or did that cause other issues?

lorgan3 (approved):

src/components/filterSelection/FilterSelection.js
JorenSaeyTL: Tried that as well, but then I have to get rid of the default styling of the button, normally the role="button" will have the same effect for accessibility (aria)

PR #1722, opened 2021-07-02
[FIX] height and hover state of FilterSelection

JorenSaeyTL:
### Description

Component with/without border should have the same height

#### Screenshot after this PR

Schermafbeelding 2021-07-02 om 11 14 37

Both have same height and introduced the hover state

### DIFF
--- a/src/components/filterSelection/FilterSelection.js
+++ b/src/components/filterSelection/FilterSelection.js
@@ -69,7 +69,7 @@ const FilterSelection = ({ label, modificationText, applied, status, amountAppli
   return (
     <Box
       ref={ref}
-      className={cx(theme['select-control'])}
+      className={cx(theme['select-control'], status === Status.DEFAULT && theme['select-control--hovered'])}
       role="button"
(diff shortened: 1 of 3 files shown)


ArnaudWeyts (approved):

driesd (approved):

src/components/filterSelection/FilterSelection.js
```diff
@@ -69,7 +69,7 @@ const FilterSelection = ({ label, modificationText, applied, status, amountAppli
   return (
     <Box
       ref={ref}
-      className={cx(theme['select-control'])}
+      className={cx(theme['select-control'], status === Status.DEFAULT && theme['select-control--hovered'])}
```
lorgan3, line 72: silly pp so feel free to ignore
`--hovered` threw me off, it's not actually being hovered yet so something like `--hoverable` seems more appropriate

lorgan3 (approved):

sanderbrugge (approved):

PR #1731, opened 2021-07-09
`Dialog` and `Popover`: make aria compliant

ArnaudWeyts:
### Description

This is the second and last part in making our dialogs and popovers aria compliant.

### Fixed

- `Dialog`: has been made aria compliant, enabling keyboard navigation
- `Popover`: has been made aria compliant, enabling keyboard navigation

### Changed

- [BREAKING] `Popover`: Popovers now mount straight on the body instead of a self-made div

### Manual check

- [x] I can navigate popovers and dialogs using the keyboard


src/components/dialog/useFocusTrap.js
```diff
@@ -0,0 +1,82 @@
...
+      <div ref={topFocusBumperRef} tabIndex="0" />
+      {children}
+      <div ref={bottomFocusBumperRef} tabIndex="0" />
+    </>
+  );
+};
+
+const openDialogs = new Set();
+
+const useFocusTrap = ({ active, sourceRef }) => {
```
lorgan3: Is this a common thing to do?

ArnaudWeyts: This pattern is based on https://www.w3.org/TR/wai-aria-practices-1.1/examples/dialog-modal/dialog.html

I've looked into the snippets provided by w3, as well as investigated some other libraries and this is usually what ends up being implemented. It makes sense to me, and it's a whole lot better than leaving the focus behind in the background 🙈

driesd (approved):

src/components/dialog/useFocusTrap.js
lorgan3: Yes it makes sense! I was just a bit surprised by the amount of code required just to be compliant

src/components/datepicker/DatePickerInput.js
```diff
@@ -141,6 +141,7 @@ class DatePickerInput extends PureComponent {
           position="end"
           offsetCorrection={30}
           zIndex={500}
+          returnFocusToSource={false}
```
lowiebenoot, line 144: As your commit message says, the datepicker input is a bit weird right now. You actually can't get past the datepicke input right now, because when it closes, it goes back to the body.

lowiebenoot (commented):
I also think we should stop blurring buttons after clicking on them. Right now when you click a button to open a dialog for example, it will use the body as the "current" active element, because it blurs immediately. So if you close the dialog, it doesn't return back to the button.

The downside is that it will always show the focus state on buttons after clicking on them, but that's native behaviour, so we probably shouldn't change it 🤷

lorgan3 (approved):

src/components/dialog/useFocusTrap.js
```diff
@@ -0,0 +1,82 @@
+import React, { useEffect, useRef, useCallback } from 'react';
+
+import { focusOnFirstDescendent, focusOnLastDescendent } from './utils';
+
+const FocusRing = ({ topFocusBumperRef, bottomFocusBumperRef, children }) => {
+  return (
+    <>
+      <div ref={topFocusBumperRef} tabIndex="0" />
+      {children}
```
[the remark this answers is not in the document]
ArnaudWeyts: Hmm, I don't think that would solve breaking the :first-child and :last-child selectors though, since you're still adding a wrapping div?

src/components/popover/Popover.js
```diff
@@ -39,7 +39,7 @@ const Popover = (props) => {
     returnFocusToSource,
   } = props;
 
-  const { ref, FocusRing } = useFocusTrap({ active, returnFocusToSource });
+  const { ref, FocusRing } = useFocusTrap({ active, returnFocusToSource, initialFocusRef: false });
```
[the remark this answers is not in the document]
ArnaudWeyts: I'm still in doubt about this, it felt a bit better to not have an initial focus in popovers, since f.e. the split button has some interesting focus states:

Screenshot 2021-07-12 at 09 22 44

However, the focus will still be set the first focus bumper, which means that when you perform your first tab, the next focusable element in the popover will receive focus

src/components/datepicker/DatePickerInput.js
ArnaudWeyts: Yup, most of it is caused by the date picker opening when focused, and the datepicker doesn't pass correct events to its handlers. I think a separate focus state for the date picker input without opening the popover and a secondary action like f.e. pressing enter could solve this.

src/components/dialog/useFocusTrap.js
ArnaudWeyts: Yup, sadly it's still very hard to do dialogs and popovers right. There's a new `dialog` html element in the works but support is still very sparse: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/dialog

src/components/dialog/useFocusTrap.js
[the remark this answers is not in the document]
lowiebenoot: If you have this for example:

```jsx
<Popover>
  <div/>
  <div/>
  <div/>
</Popover>

```

With current implementation those pseudo selectors would change, because you're adding a different first and last child.

src/components/popover/Popover.js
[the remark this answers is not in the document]
lowiebenoot: Hmm yeah, it depends a bit on the case I guess.
I recently built a Popover with some fields, and there it would make sense to receive an initial focus.
Maybe we should be able to overwrite it? Enable it by default, but disable it for some other components, e.g. the menu here. But then you could argue that it's not consistent...

src/components/dialog/useFocusTrap.js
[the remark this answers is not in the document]
ArnaudWeyts: I see, I'll add a wrapping div in this case!

src/components/popover/Popover.js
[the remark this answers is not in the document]
ArnaudWeyts: Planned a meeting to discuss this, in the meantime I don't think this is blocking for this story.

src/components/datepicker/DatePickerInput.js
ArnaudWeyts: Planned a meeting to discuss this, in the meantime I don't think this is blocking for this pull request.

ArnaudWeyts:
> I also think we should stop blurring buttons after clicking on them. Right now when you click a button to open a dialog for example, it will use the body as the "current" active element, because it blurs immediately. So if you close the dialog, it doesn't return back to the button.
> 
> The downside is that it will always show the focus state on buttons after clicking on them, but that's native behaviour, so we probably shouldn't change it 🤷

Yeah, I don't know why we implemented this behaviour in the first place, we should stick to the default button behaviour as much as possible.

jnstr (approved):

sanderbrugge (approved):
👏

jnstr:
MC ok

lowiebenoot:
> Yeah, I don't know why we implemented this behaviour in the first place, we should stick to the default button behaviour as much as possible.

@ArnaudWeyts this was implemented to avoid the "ugly" focus state after clicking on a button. But it seems like browsers changed this behaviour and don't focus anymore on clicking (i know chrome used to focus on buttons after clicking). This used to be very inconsistent between browsers, but seems like it's better now. 

So we can remove that behaviour.

PR #1737, opened 2021-07-13
Fix regression bugs caused by ref forwarding and aria-compliancy

ArnaudWeyts:
### Fixed

- `Overlay`: the overlay now stops propagating the escape key event when active, even when no escape handler is attached
- `NumericInput`: fix stepper controls not updating the value of the input

PR #1738, opened 2021-07-13
Don't allow more characters than needed in duration input 

jnstr:
We're limiting the length of the input when manually typing in the duration input instead of using the stepper.

Minutes should always be limited to maximum 2 characters.

We only limit the length of the hour input if there is a maximum duration.

--- 

**:rotating_light: Hold merge until #1737 is merged :rotating_light:**

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,8 @@
 
 ### Changed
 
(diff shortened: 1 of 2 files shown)


adognameddog (approved):

ArnaudWeyts (commented):
Don't forget to update the changelog

jnstr:
@ArnaudWeyts what else needs to happen? 7e15843

PR #1762, opened 2021-08-13
Chromatic visual regression testing

lowiebenoot:
This week I have been experimenting with [Chromatic](https://www.chromatic.com) visual regression testing.
It's a tool built by the storybook team, to do visual regression testing on your storybook stories.

Sadly, just testing our existing stories wouldn't give us much coverage, since it barely tests variations of the components, while most of our components have quite some variations.

With some small trickery, I was able to run these tests on a separate group of stories (`*.stories.spec.js`). These stories are fairly simple, and just contain a bunch of variations from the component, usually by mapping over one or more arrays. 
I sometimes added some spacers and wrappers with min-height, to make the screenshots a bit more appealing, and to avoid that the diff between 2 screenshots would be completely messed up when a component changes a bit.
I didn't add stories for all components yet because I first want to check how it's working, before spending even more time on creating these stories.

I added it as a step in our CI pipeline, but I'm not sure how this will work out. When there is a visual regression, we should go to chromatic and accept (or refuse) the changes in their tool. I'm wondering what happens when you accept these changes. Currently, I feel like the step will just remain as "failed", and we would have to re-run the task, which wouldn't be very handy of course (certainly because the payment modal is based on the amount of snapshots taken).

Here are some screenshots of how it looks like (no worries, I'll send invites as well so you can check it out).

List of builds
Screenshot 2021-08-13 at 16 07 18

List of stories inside a build that have regressions
Screenshot 2021-08-13 at 16 12 34

The diff between story snapshots from 2 different builds
Screenshot 2021-08-13 at 16 13 25

Screenshots from the stories
Screenshot 2021-08-13 at 16 12 46


package.json
```diff
@@ -86,6 +86,7 @@
     "babel-eslint": "^10.0.2",
     "babel-loader": "^8.0.0-beta.6",
     "babel-plugin-transform-react-remove-prop-types": "^0.4.13",
+    "chromatic": "^5.10.0-canary.0",
```
ArnaudWeyts, line 89: Any specific reason why we're using a canary build instead of a stable build?

ArnaudWeyts (approved):

package.json
lowiebenoot: No, didn't even notice it. I just did `yarn add --dev chromatic`. Wouldn't expect it to install a non-stable build then 🤔

ArnaudWeyts: Huh, maybe they configured it as a default version? Though judging from npm it doesn't look like it

Screenshot 2021-08-13 at 19 23 50

lowiebenoot: Changed it to use the latest stable version instead.

ArnaudWeyts (approved):

timdegroote (approved):

driesd (approved):

lorgan3 (approved):

lowiebenoot:
I'm going to merge this. Curious if the CI setup will work 😅

PR #1767, opened 2021-08-16
Copy extra files while building

lorgan3:
### Description

The babel build only includes files that were built by it so I've added `--copy-files` to include the required svgs
Then I've also added `--no-copy-ignored` so the stories are excluded.
And finally `--extensions` was added to get around [an issue](https://github.com/babel/babel/issues/11394) with the `no-copy-ignored` parameter


lorgan3:
We decided to inline the svgs instead

PR #1791, opened 2021-09-07
Fix date picker locale

lorgan3:
### Description

The `DatePickerInput` passes `localeUtils` to the day-picker so it is translated but the basic `DatePicker` does not, which results in it always being english unless you implement them yourself.
This fixes that behaviour by passing the utils in the DatePicker instead of DatePickerInput

/

#### Screenshot after this PR

/

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,8 @@
 
 ### Fixed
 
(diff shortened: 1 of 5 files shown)


sanderbrugge (approved):

ArnaudWeyts (approved):

EOSullivanBerlin (approved):

JorenSaeyTL (approved):
Nice!

ArnaudWeyts:
✅ Still works like it used to based on the stories

PR #1794, opened 2021-09-12
`DurationInput`: add `small`, `medium` (default) and `large` size property

ArnaudWeyts:
### Added

- `DurationInput`: add `small`, `medium` (default) and `large` size property

#### Screenshot after this PR

Example of the small input

Screenshot 2021-09-12 at 12 27 13

### DIFF
--- a/src/components/input/DurationInput.js
+++ b/src/components/input/DurationInput.js
@@ -212,6 +212,7 @@ DurationInput.propTypes = {
 };
 
 DurationInput.defaultProps = {
+  textAlignRight: false,
   size: 'medium',
 };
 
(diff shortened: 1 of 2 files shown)


lowiebenoot (approved):

lowiebenoot:
Checked ✅

jnstr (approved):

src/components/input/DurationInput.js
```diff
@@ -203,6 +208,12 @@ DurationInput.propTypes = {
   error: PropTypes.bool,
   /** In minutes **/
   max: PropTypes.number,
+  size: 'small' | 'medium' | 'large',
```
lorgan3, line 211: Does this actually work or did you mix up proptypes and typescript? 😅

ArnaudWeyts: Woops, you're right, I'm not getting any errors but this doesn't seem to be the right way to define this 🙈 Fixed!

EOSullivanBerlin (approved):

lorgan3 (approved):

PR #1802, opened 2021-09-29
Experiment/changie

ArnaudWeyts:
This was just an experiment, the dockerfile definitely shouldn't live inside of this repository


lowiebenoot (commented):
Looks cool. This would mean that we have to run it for every dependabot PR as well, right?

ArnaudWeyts:
>Looks cool. This would mean that we have to run it for every dependabot PR as well, right?

@lowiebenoot Not necessarily, dependabot updates are always tricky, I'm a proponent of not including these updates all together. Some of the dependency updates only relate to storybook/dev dependencies that don't affect the end users of the library. Only for packages like react-day-picker, of which we make the internal API available to end users, we can make an exception

lowiebenoot:
> Some of the dependency updates only relate to storybook/dev dependencies that don't affect the end users of the library. Only for packages like f.e. react-day-picker, of which we make the internal API available to end users, we can make an exception and generate a changelog entry

@ArnaudWeyts that would make sense, yes. 👍

PR #1806, opened 2021-10-07
DatePickerInput: Remove zIndex overwrite

lorgan3:
### Description

If a select is rendered before a DatePickerInput is opened, the month picker would appear below the date picker. This is because both are currently rendered at z-index 500. 
This pr just removes the overwrite from the input so it defaults to 300 (below any select)

image

#### Screenshot after this PR
image

### Breaking changes

- Technically this is a breaking change but the odds of this actually breaking something are so low that creating a new major version seems overkill. Let me know if you disagree and I'll update the changelog.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,8 @@
 
 ### Fixed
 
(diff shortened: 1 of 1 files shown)


JorenSaeyTL:
🎉

PR #1823, opened 2021-10-29
`Overlay, Dialog, Popover` Fix dragging outside of the Overlay causing the Overlay to close

ArnaudWeyts:
### Description

When clicking inside of overlay content, and dragging outside before releasing the click. Chrome, Safari and Edge caused the Overlay to be closed. This is because the `click` event target is handled differently in these browsers.

https://user-images.githubusercontent.com/10011712/139430013-40dc1ce0-288f-4343-a800-e4e7dea6210c.mov

### Breaking changes

- `Overlay`: `onClick`'s behaviour has been replaced by `onInteractOutside`
- `Dialog, DialogBase, Popover`: `onOverlayClick` has been replaced by `onInteractOutside`

### Manual check

- Recreate the video in Chrome, Safari or Edge


lowiebenoot (commented):
Personally I feel like `onInteractOutside` is not a correct name, since you're not interacting outside of the Overlay. Why not just keep it as `onClick`? Because now it's actually a breaking change, no? It's not set as breaking in the changelog, but you did add the breaking tag on this PR.

lorgan3 (commented):
I agree with Lowie, the name change doesn't seem required. I feel like this is a fix to make sure `onClick` works as expected ( =not a breaking change)
Renaming the prop does make it breaking though

ArnaudWeyts:
@lowiebenoot @lorgan3 You're both right and wrong, I wanted to rename the handler because this seems to just be how `onClick` works in Chrome, Safari and Edge. I want to prevent people from expecting a regular onClick event even though it's going to be a mouse down event now.

I would still want to call it something different and introduce a breaking change on the `Overlay`, to make it very explicit that we have custom behavior for this handler, but the naming could indeed be improved, maybe something like:

- onOverlayInteract
- onOverlayClick (too close to onClick for me, but I could settle)
- onInteractOutsideOfContent

The missing breaking change indication in the changelog was indeed an oversight

sanderbrugge:
Is this fix in response of an actual issue? Because, to me, this looks like expected behaviour?

ArnaudWeyts:
> Is this fix in response of an actual issue? Because, to me, this looks like expected behaviour?

@sanderbrugge Yes, this is in response to a customer report

jnstr (approved):

lorgan3 (approved):

JorenSaeyTL (approved):

lowiebenoot (approved):
Not sure if it should still be a breaking change. Using `onClick` on the `Overlay` component still works. But if we don't make it a breaking change, then people will not change it to `onOverlayClick`... I'm a bit in between.

lorgan3:
✅ checked it works on chrome and still works of firefox

ArnaudWeyts:
> Not sure if it should still be a breaking change. Using `onClick` on the `Overlay` component still works. But if we don't make it a breaking change, then people will not change it to `onOverlayClick`... I'm a bit in between.

@lowiebenoot I'll make it a breaking change, since if someone is already using Overlays, they need to make swap out the handler (the onClick will just get fired any time you click inside, beating the point of an overlay)

PR #1828, opened 2021-11-04
[ATL-460] Fix `Overlay` conflicting with `Select` component

ArnaudWeyts:
### Description

- The `Select` component was fresh rendering components when opening, causing the `clickOrigin.current` to no longer be present in the DOM and the `Overlay` component to detect the click as outside of the overlay content

### Manual check

- Let's do an over the shoulder check

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,9 @@
 
 ### Fixed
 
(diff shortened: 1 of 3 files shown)


sanderbrugge (approved):

jnstr (approved):

lorgan3 (approved):
Thanks!

PR #1836, opened 2021-11-08
Prevent click events in the `Overlay` from propagating to the parent

ArnaudWeyts:
### Description

- Prevent click events in the `Overlay` from propagating to the parent

### Manual check

Let's do an over the shoulder check

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,8 @@
 
 ### Fixed
 
+- `Overlay`: Prevent click events in the `Overlay` from propagating to the parent ([@ArnaudWeyts](https://github.com/ArnaudWeyts) in [#1836](https://github.com/teamleadercrm/ui/pull/1836))
+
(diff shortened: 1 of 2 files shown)


jnstr (approved):

lowiebenoot (approved):
Can you explain why the event shouldn't shouldn't bubble up? The Overlay is injected into the body, so there's not much to bubble up to.

ArnaudWeyts:
> Can you explain why the event shouldn't shouldn't bubble up? The Overlay is injected into the body, so there's not much to bubble up to.

@lowiebenoot I was kinda stunned by this as well, my guess at the moment is that React still sees the overlay as a `child` of its virtual DOM tree, even though we do some custom mounting on the body, and it still bubbles up the click events to the parent components

lowiebenoot:
@ArnaudWeyts waw, that's quite remarkable 🤯

PR #1877, opened 2021-12-16
`Widget`, `IslandGroup`: ref forwarding, `boxProps`: style property

ArnaudWeyts:
### Added

- `Widget`, `IslandGroup`: added ref forwarding

### Changed

- [BREAKING] `Checkbox`, `Input`, `TimeInput`, `NumericInput`, `SingleLineInputBase`, `Textarea`, `RadioButton`, `Select`, `Toggle`: due to the `style` prop now being part of `boxProps`, the property will no longer get passed to the internal elements, but rather the wrapping `Box`, same as `classname`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,11 @@
 
 ### Added
 
+- `Widget`: added ref forwarding ([@ArnaudWeyts](https://github.com/ArnaudWeyts) in [#1877](https://github.com/teamleadercrm/ui/pull/1877))
+
 ### Changed
+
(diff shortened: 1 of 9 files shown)


lorgan3 (approved):

ArnaudWeyts:
Holding because I'm evaluating if I need ref forwarding on the widget body or just on the widget itself

lorgan3 (approved):
Can you add some more context why you need ref forwarding on the widget and IslandGroup (and not on the widget body & Island) ?

ArnaudWeyts:
> Can you add some more context why you need ref forwarding on the widget and IslandGroup (and not on the widget body & Island) ?

@lorgan3 Context can more or less be found inside of https://teamleader.atlassian.net/browse/ATL-744, I'm opening a PR on our core repository soon that will make use of this ref forwarding which will give more context.

JorenSaeyTL (approved):

PR #1893, opened 2022-01-05
Fix security issues and dependency updates

qubis741:
https://teamleader.atlassian.net/browse/ATL-736

### Description

Updated all dependencies to latest except `react`, `react-dom`, `react-day-picker` (latest is beta) and vulnerable ones in `resolutions`.
Fixed security issues


ArnaudWeyts:
Taking a look at the linting errors and caught UI changes

package.json
```diff
@@ -113,19 +113,19 @@
     "postcss-cli": "^9.1.0",
     "postcss-color-function": "^4.1.0",
     "postcss-custom-media": "^8.0.0",
-    "postcss-each": "^0.10.0",
+    "postcss-each": "^1.1.0",
```
lorgan3, line 116: Just a heads up, this version seems to be missing a lot of functionality that we need
https://github.com/teamleadercrm/ui/pull/1742

package.json
```diff
@@ -67,66 +67,78 @@
...
+    "**/trim": "1.0.0",
+    "prismjs": "1.25.0",
+    "glob-parent": "6.0.1",
+    "browserslist": "4.19.1",
+    "trim-newlines": "3.0.1",
+    "color-string": "1.5.5",
+    "semver-regex": "3.1.3",
+    "nth-check": "2.0.1",
+    "ansi-regex": "5.0.1"
   },
```
lorgan3, line 142: Should we document somewhere for each of these dependencies why we did this and when they can be removed again? Otherwise these will never get cleaned up after the vulnerabilities are fixed (if ever)

lorgan3 (commented):
I'm also a little bit sceptical you can do all those major bumps without any code or configuration changes. But it's possible of course. 😅

package.json
ArnaudWeyts: Indeed, I don't think we'll move forward with these fixed resolutions, I'd rather wait for the main dependency to update / find a work around to not use the dependency that requires these anymore

ArnaudWeyts:
⏳ Manually checking...

ArnaudWeyts (approved):

lorgan3 (approved):

lowiebenoot (approved):

JorenSaeyTL (approved):

PR #1925, opened 2022-01-18
Create separate badged link component instead of link with icon and badged property

qubis741:
https://teamleader.atlassian.net/browse/ATL-719

### Description

`Link` component props `icon` and `badged` are removed. New component `BadgedLink` is created, to supply them.

### Breaking changes

- `icon` and `badged` prop is removed from Link
- added `BadgedLink` component

### Manual Check
Checkout branch, open storybook
- [x] Link doesn't have `icon` and `badged` prop
- [ ] `BadgedLink` looks okay
- [ ] `DetailPage` header backlink looks okay


src/components/badgedLink/BadgedLink.js
```diff
@@ -83,6 +84,8 @@ BadgedLink.propTypes = {
   // disabled: PropTypes.bool,
   /** The icon displayed inside the button. */
   icon: PropTypes.element.isRequired,
+  /** The aligment of the icon inside the button. */
+  iconAligment: PropTypes.oneOf(['center', 'bottom']),
```
ArnaudWeyts, line 87: I'm assuming there's a use case for this somewhere? Could you show us a use case that would need "bottom" icon alignment instead of centered?

src/components/detailPage/theme.css
```diff
@@ -8,4 +8,8 @@
   position: absolute;
   top: 0;
   left: -25px;
+
+  &__icon {
+    margin-top: 1px;
+  }
```
ArnaudWeyts, line 14: Why is this needed? Shouldn't this alignment already be part of the BadgedLink?

ArnaudWeyts:
Let's also add screenshot tests for all new components, for an example you can look at the .spec files

qubis741:
Thanks for your feedback. I have updated styling, so `iconAligment` is not needed anymore, added snapshot and changelog.

CHANGELOG.md
```diff
@@ -2,6 +2,8 @@
 
 ### Added
 
+- [BREAKING] `BadgedLink`: ([@qubis741](https://github.com/qubis741) in [#1925](https://github.com/teamleadercrm/ui/pull/1925))
```
lorgan3, line 5: I don't think adding a component is a breaking change 😄

src/components/badgedLink/BadgedLink.js
```diff
@@ -0,0 +1,105 @@
...
+        ref={badgedLinkRef}
+        className={classNames}
+        data-teamleader-ui="badgedLink"
+        onMouseUp={handleMouseUp}
+        onMouseLeave={handleMouseLeave}
+        {...others}
+      >
+        {icon && iconPlacement === 'left' && icon}
+        {children}
+        {icon && iconPlacement === 'right' && icon}
```
lorgan3, line 71: If the icon is actually required like the proptypes say

```suggestion
        {iconPlacement === 'left' && icon}
        {children}
        {iconPlacement === 'right' && icon}
```

lorgan3 (commented):
Can you also add the `BadgedLink` to the `index.d.ts` file? 🙏

lorgan3 (approved):

ArnaudWeyts (approved):

ArnaudWeyts:
⏳ Manually checking

lowiebenoot (approved):

driesd (approved):

PR #1938, opened 2022-01-25
Restored mouse event propagation in `Overlay` component

jelledc:
### Description

Restored event bubbling for click, mousedown and mouseup events in the Overlay component because the use of event.stopPropagation prevented the handling of said events outside of the overlay (eg. routing listening to onclick events).

Note: the handling of ESC key presses will still stop propagation.

### DIFF
--- a/src/components/overlay/Overlay.js
+++ b/src/components/overlay/Overlay.js
@@ -51,18 +51,11 @@ class Overlay extends PureComponent {
     }
   };
 
-  handleClick = (event) => {
-    event.stopPropagation();
-    this.props.onClick?.(event);
-  };
-
   handleMouseDown = (event) => {
-    event.stopPropagation();
     this.clickOriginRef.current = event.target;
   };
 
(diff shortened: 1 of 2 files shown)


lowiebenoot (dismissed):

src/components/overlay/Overlay.js
```diff
@@ -90,7 +83,6 @@ class Overlay extends PureComponent {
               onKeyDown={this.handleEscKey}
               onMouseDown={this.handleMouseDown}
               onMouseUp={this.handleMouseUp}
-              onClick={this.handleClick}
```
qubis741, line 93: Missing of `props.onClick` is okay?

jelledc: Yes, if passed it will be set through `{...other}`.

qubis741 (approved):

lowiebenoot (approved):

PR #1939, opened 2022-01-25
[FEATURE] fade out dialog

JorenSaeyTL:
### Description

- Add `fadeOutOnClose` property
- Add property in storybook

This will make it possible to make the dialog fade out on close (Use case: https://www.figma.com/proto/ViF1ZZVAmlae4lCLZkrSne/Global-search?page-id=7088%3A25977&node-id=7088%3A25978&viewport=241%2C48%2C0.65&scaling=scale-down&starting-point-node-id=7088%3A25978)

### Screenshot after this PR

- My screen recording software crashes on me, perhaps best to manual check this through storybook

### Breaking changes

- N/A

### DIFF
--- a/src/components/dialog/DialogBase.js
+++ b/src/components/dialog/DialogBase.js
@@ -1,4 +1,4 @@
-import React from 'react';
+import React, { useState, useEffect } from 'react';
 import { createPortal } from 'react-dom';
 import PropTypes from 'prop-types';
 import cx from 'classnames';
@@ -12,6 +12,8 @@ import theme from './theme.css';
 import uiUtilities from '@teamleader/ui-utilities';
 import useFocusTrap from '../../utils/useFocusTrap';
 
+const FADE_OUT_TIME = 500;
+
 export const DialogBase = ({
   active,
   backdrop,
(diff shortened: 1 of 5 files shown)


src/components/dialog/DialogBase.js
```diff
@@ -20,31 +22,58 @@ export const DialogBase = ({
...
   onOverlayClick,
   scrollable,
+  fadeOutOnClose,
   size,
   initialFocusRef,
 }) => {
   const { ref, FocusRing } = useFocusTrap({ active, initialFocusRef });
 
-  if (!active) {
+  const [delayedActive, setDelayedActive] = useState(false);
```
lorgan3, line 31: PP: react-transition-group has an event when the dialog has exited, can you use that instead of the `setTimeout`?

http://reactcommunity.org/react-transition-group/transition#Transition-prop-onExited

lorgan3 (approved):

src/components/dialog/DialogBase.js
JorenSaeyTL: Gonna try it out!

JorenSaeyTL: Works 👍

lorgan3 (approved):

sanderbrugge (approved):
Great stuff!

lowiebenoot (approved):

ArnaudWeyts:
Hmm, seems like a very one-off implementation for that kind of dialog 🤔

sanderbrugge:
it just works ✅

ArnaudWeyts:
Guess it's fine for now, as there is no 100% solution for it without making a complete copy of the Dialog component with this implementation, just looking to avoid these magic properties that will only be used once

ArnaudWeyts (approved):

JorenSaeyTL:
I'll hold it until the discussion is done

ArnaudWeyts (changes requested):
After a short discussion with Stephen and @JorenSaeyTL, I don't think this belongs in the UI-library (yet) since it's still a very one-off, if we want to continue with it, let's create a separate component closer to the implementation until the pattern has matured.

PR #1958, opened 2022-02-01
[LIG-835] Prevent clicks in underlying `Dialog` from closing parent `Overlay`

jelledc:
### Link to JIRA issue

https://teamleader.atlassian.net/browse/LIG-835

### Description

Added extra condition that checks if the click event was sent from the current portal's DOM tree. Nested overlays will be rendered in a different tree and won't trigger their parent's onOverlayClick event anymore.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,8 @@
 
 ### Fixed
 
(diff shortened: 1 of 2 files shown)


lorgan3 (approved):
Sounds like it should work

ArnaudWeyts (approved):

lowiebenoot (approved):
Seems to be fine. Tested it by adding a Menu inside a Popover inside a Dialog.

✅ Clicking next to the Menu only closes the Menu
✅ Dragging inside the dialog and releasing outside the dialog still doesn't hide the dialog

PR #1971, opened 2022-02-08
Investigate dialog on esc key down not triggering

qubis741:
https://teamleader.atlassian.net/browse/FRAF-183

### Description

By adding fallback by @ArnaudWeyts  in `useFocusTrap`, that focuses `topFocusBumper` is nothing focusable is found in `FocusRing`, Dialogs are now closable with ESC key right away.
`DialogBase` without any focusable element needs to have explicitly set `initialFocusRef`, so that ESC key can work.

_Proper focus passing after closing Dialog is not working right now, more in ticket comment_

### Manual Check
- [x] All 3 `Dialogs` in `Dialog (mid level block)` are closable with ESC key right after open (no additional clicking or focus needed)

### DIFF
--- a/src/components/dialog/dialog.stories.js
+++ b/src/components/dialog/dialog.stories.js
@@ -1,4 +1,4 @@
-import React, { useState } from 'react';
+import React, { useRef, useState } from 'react';
 import { addStoryInGroup, MID_LEVEL_BLOCKS } from '../../../.storybook/utils';
 import { IconWarningBadgedMediumOutline } from '@teamleader/ui-icons';
 import { Box, Button, Dialog, TextBody } from '../../index';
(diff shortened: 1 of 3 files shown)


ArnaudWeyts (approved):

ArnaudWeyts:
✅

src/components/dialog/dialog.stories.js
```diff
@@ -137,7 +137,7 @@ DialogWithFocusStates.parameters = {
 
 export const DialogBaseStory = (args) => {
   const [active, setActive] = useState(false);
-
+  const bodyRef = useRef()
```
lorgan3, line 140: Linting looks off here

lorgan3 (approved):

src/components/dialog/dialog.stories.js
qubis741: No automatic linting here :/

lorgan3 (approved):

ArnaudWeyts (approved):

PR #1979, opened 2022-02-11
[FRAF-289] Implement draggable Dialogs

sanderbrugge:
### Description
This PR makes it possible to move around the `Dialog` in the viewport.

Added a `draggable` prop:
When true, the `Dialog` header gets a draggable indicating icon, which you can grab.

Screenshot 2022-03-15 at 15 13 58

#### Screenshot after this PR

Screenshot 2022-03-15 at 15 11 36

### Breaking changes

None.


src/components/dialog/DialogBase.js
```diff
@@ -70,6 +75,39 @@ export const DialogBase = ({
   return createPortal(dialog, document.body);
 };
 
+const makeDraggable = () => {
+  let x = 0;
+  let y = 0;
+
+  const ele = document.getElementById('dragMe');
```
ArnaudWeyts, line 82: ref should be used here, which means we should probably also create a hook for it instead (similar to the package), but I think I prefer maintaining it here than inside of that package, especially since it's not a very "established" package

sanderbrugge: yup, fair enough, thought the package was neat because you can set a drag target and handle ref, which can mimic the existing implementation of "handling drag bij clicking only the icon, but targeting the full popup".

ArnaudWeyts: Doesn't seem super hard to implement ourselves either

qubis741: If its easy to implement it && that package in other draft is not reliable, then I guess we should go with our implementation

ArnaudWeyts:
Closing this PR since it will be followed up later

lowiebenoot (approved):

qubis741 (approved):

lorgan3 (approved):
Looks good. What happened to the lockfile though 😨

ArnaudWeyts:
Going to close this PR and recreate without the lock file changes, the original PR shouldn't have been re-opened since it was a POC

ArnaudWeyts:
> Checked ✅
> 
> You can drag the dialog outside of your window, and then if you release you can't get it back. But I guess we don't really care about that (or at least not now I would say).

@lowiebenoot Indeed, this is a slight behavioral change as apposed to our current legacy implementation, but I'll check this off with Stephen to see if it suffices

PR #2074, opened 2022-04-08
Allow passing rest props to DialogBase component

lorgan3:
### Description

This allows the `Dialog` / `DialogBase` to become a form element.
I did this so the primaryAction / secondaryAction are inside the form and they can submit it. Putting the form around the dialog did not work because the dialog is rendered outside of it using portals.

example code:
```
        <Dialog
          title="Form"
          active
          primaryAction={{ type: 'submit', label: 'Save'}}
          secondaryAction={{ onClick: onClose, label: 'Cancel' }}
          onCloseClick={onClose}
          onEscKeyDown={onClose}
          onOverlayClick={onClose}
          form
          onSubmit={handleSubmit}
        >
         {/* content here */}
        </Dialog>
```

#### Screenshot after this PR

there are no visual changes

### Breaking changes

N/a

### DIFF
--- a/src/components/dialog/DialogBase.js
+++ b/src/components/dialog/DialogBase.js
@@ -24,6 +24,7 @@ export const DialogBase = ({
   size,
   initialFocusRef,
   dragHandleRef,
+  ...rest
 }) => {
   const { ref, FocusRing } = useFocusTrap({ active, initialFocusRef });
   useDraggable({ active, dragTargetRef: ref, dragHandleRef });
@@ -52,7 +53,7 @@ export const DialogBase = ({
             onEscKeyDown={onEscKeyDown}
           >
             <FocusRing>
-              <div ref={ref} data-teamleader-ui="dialog" className={dialogClassNames}>
+              <Box ref={ref} data-teamleader-ui="dialog" className={dialogClassNames} {...rest}>
                 <div className={theme['inner']}>
                   {scrollable ? (
                     <Box display="flex" flexDirection="column" overflowY="auto">
@@ -62,7 +63,7 @@ export const DialogBase = ({
                     children
                   )}
                 </div>
-              </div>
+              </Box>
             </FocusRing>
           </Overlay>
         );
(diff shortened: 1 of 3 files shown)


src/components/dialog/DialogBase.js
```diff
@@ -24,6 +24,7 @@ export const DialogBase = ({
   size,
   initialFocusRef,
   dragHandleRef,
+  ...rest
```
ArnaudWeyts, line 27: I'm generally not a fan of just spreading all properties down, especially if there's only one property needed, can we just spread the element property for now?

src/components/dialog/DialogBase.js
```diff
@@ -52,7 +53,7 @@ export const DialogBase = ({
             onEscKeyDown={onEscKeyDown}
           >
             <FocusRing>
-              <div ref={ref} data-teamleader-ui="dialog" className={dialogClassNames}>
+              <Box ref={ref} data-teamleader-ui="dialog" className={dialogClassNames} {...rest}>
```
ArnaudWeyts, line 56: I'm also wondering if this could break something, the default display value for a div differs from a form

`div: display: block`
`form: display: inline-block`

src/components/dialog/DialogBase.js
ArnaudWeyts: Let's do that, because I don't see the value of ever using a different element for a `Dialog`

lorgan3: I had that at first but I also needed to pass the `onSubmit` function to make it work.
Since that's kind of a random property to only applies to the `form` element I preferred just spreading all the props

src/components/dialog/DialogBase.js
lorgan3: It only becomes a form element if you actually pass `element="form"` to the `Dialog` so I don't think this is an issue

driesd (approved):

jelledc (approved):

stefaandevylder (approved):

ArnaudWeyts (approved):

PR #2085, opened 2022-04-14
Add menu vertical offset

lorgan3:
### Description

If you use react-select together with `menuWidth` near the right edge of the screen, the menu will go off the screen.
I've added a `menuVerticalOffset` prop to compensate for this. Unfortunately because we use portals a simple `alignRight`/`alignLeft` setting is not possible

image

#### Screenshot after this PR

image

### Breaking changes

N/a

### DIFF
--- a/src/components/select/Select.js
+++ b/src/components/select/Select.js
@@ -183,11 +183,20 @@ class Select extends PureComponent {
     };
   };
 
-  getMenuStyles = (base) => ({
-    ...base,
-    marginTop: 3,
-    marginBottom: 3,
-  });
+  getMenuStyles = (base) => {
+    const { menuVerticalOffset } = this.props;
+    return {
+      ...base,
+      marginTop: 3,
+      marginBottom: 3,
+      ...(menuVerticalOffset
+        ? {
+            right: menuVerticalOffset,
+            position: 'absolute',
+          }
+        : {}),
+    };
+  };
 
   getMenuPortalStyles = (base) => {
     const { inverse, menuWidth } = this.props;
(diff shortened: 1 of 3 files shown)


stefaandevylder (approved):

src/components/select/Select.js
```diff
@@ -183,11 +183,20 @@ class Select extends PureComponent {
...
-    marginTop: 3,
-    marginBottom: 3,
-  });
+  getMenuStyles = (base) => {
+    const { menuVerticalOffset } = this.props;
+    return {
+      ...base,
+      marginTop: 3,
+      marginBottom: 3,
+      ...(menuVerticalOffset
```
lowiebenoot, line 192: Isn't this horizontal instead of vertical?

lowiebenoot (commented):
I don't fully understand how this prop works. It's absolute positioned against what?
In the screenshot, what did you use to fix the issue? An offset of 0?

src/components/select/Select.js
lorgan3: 💩 you're right

lorgan3:
@lowiebenoot it's a relative offset (it's `position: absolute` in the css due to how that menu works)
So the value that I used in the screenshot is the amount of px that I wanted to move it to the left, eg
* menuWidth prop: '300px'
* select itself is 200px wide
* menuVerticalOffset: '100px'

(Will rename it to menuHorizontalOffset when this conversation is resolved, maybe you have a better solution that I can use instead)

jelledc:
Still needs the changelog update :see_no_evil:.

lorgan3:
@jelledc Thanks, I did update the changelog but I forgot pushing it 🙈

stefaandevylder (approved):

driesd (approved):

jelledc (approved):

PR #2089, opened 2022-04-15
[FRAF-552] Add onClick event handler to the Widget Header

farazatarodi:
### Description

Adds onClick event handler to the Widget Header.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,12 @@
 
 ### Dependency updates
 
+## [14.2.0] - 2022-04-12
+
+### Added
+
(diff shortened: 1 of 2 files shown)


ArnaudWeyts (dismissed):

jelledc (changes requested):
I believe you still need to update the changelog :slightly_smiling_face:

jelledc (approved):

qubis741 (approved):

qubis741:
Q: Shouldn't there be `cursor: pointer` style when header is clickable? Also some story with `onClick` would be useful. Otherwise MCed, works.

driesd (approved):

PR #2090, opened 2022-04-19
Add incremental typescript support

ArnaudWeyts:


src/components/box/Box.tsx
```diff
@@ -4,7 +4,9 @@ import { COLOR, COLORS, TINTS } from '../../constants';
 import theme from './theme.css';
 
 const margins = [-8, -7, -6, -5, -4, -3, -2, -1, 0, 1, 2, 3, 4, 5, 6, 7, 8] as const;
+export type Margin = typeof margins[number];
```
qubis741, line 7: Q: Won't it be more convenient to have `Margin` and `Padding` in separate file? just from hierarchy POV

qubis741 (approved):

src/components/box/Box.tsx
```diff
@@ -1,78 +1,129 @@
...
+  paddingRight = paddingHorizontal,
+  paddingTop = paddingVertical,
+  style,
+  textAlign,
+  ...others
+}: Props) => {
+  const getBorder = (value: number) => {
+    return `${value}px solid ${
+      COLOR[borderColor.toUpperCase() as Uppercase<keyof typeof COLOR>][
+        borderTint.toUpperCase() as Uppercase<keyof typeof COLOR['AQUA']>
```
lorgan3: PP define the colors using enums instead of having to try really hard to make the types work

example: https://www.typescriptlang.org/play?#code/KYOwrgtgBAKgliALgZygbwFBSgGTgcwAtFhlEoBeKAchwEkBxACRgFEBlGagGi1wOKUa9Zl17YAcgHsAThACGAGyHUJAeQBKAWQCCOHnwAi8mQGsVhnRoDSB7MbOlyVapZscxGAL4YMoSFAAwlKKsqiY2DoAjmDyKjoAigCqOnZQDCEAJioMajiGaRLAYIgySioSrEkwGnppWgjONFp0Ep7YGmAARgCeKhpJAEIAmmkwwOUubHXiUABqcCHATdRzdHmsnj4YAMZSIGRBeZoAXFAawHsymQA8waEyyNznl7K38EhPUGQyCPgAfP8hBEoABte5hAB00ViAF0ziDsKCPihIXgiCQyPCaABiABmACY8QAOPF4tJIlHINECRDY6g4nY7YAAFjJFLBVMh0jkSnpOIAnAKdgL2bNKY1qQ5TPyAAyygDsAEZCRzkZLIdKnHLZSzMvIAGxpLyzAD0pqghmAeIQwCgUkQhGAMigeweyG8AG5fHsDuR8MtBm9nUIABQAdzgmUdZ3AEC6zuebtkZwhj2eiEaZypAEpKECQTJlmAZCAoAADAAkaEj0cIXgADgAPb4hKNQauBY4aUHJmSw0GZpCwrzl7w+-bIJaQ0L4UMBxBB67O0NK55p6kw+TPLk8hSKHM5z1AA

jelledc (approved):

src/components/box/Box.tsx
ArnaudWeyts: I inlined them because they were defined inline before will need to keep an eye on components that do not go through the Box component to set margins or paddings, but I think most components use this base component at the moment

src/components/box/Box.tsx
```diff
@@ -14,57 +14,59 @@ const borderRadiuses = {
...
+    padding: Padding;
+    paddingHorizontal: Padding;
+    paddingVertical: Padding;
+    paddingBottom: Padding;
+    paddingLeft: Padding;
+    paddingRight: Padding;
+    paddingTop: Padding;
+    style: CSSProperties;
+    textAlign: 'center' | 'left' | 'right';
+  } & HTMLProps<HTMLElement>
```
lorgan3, line 68: Would be cool if we could narrow this down based on the element type but I haven't found a way to do this

eg
```
<Box element="a" /> // now only `HTMLAnchorElement` props are allowed instead of `HTMLElement`

File: src/components/box/Box.tsx (hunk 2 above)
Comment (lorgan3) [reply src/components/box/Box.tsx]: On second thought the lowercase / uppercase will still be required to prevent changing the api. It's still slightly nicer than the `keyof typeof` thing IMO

```
const getBorder = (width: number, color: Lowercase<Colors> | Colors, tint: Lowercase<Tints> | Tints) => {
  return `${width}px solid ${COLOR[color.toUpperCase() as Colors][tint.toUpperCase() as Tints]}`
}

// New ts way
console.log(getBorder(1, Colors.Aqua, Tints.Normal));

// Old js way
console.log(getBorder(1, 'aqua', 'normal'));
```

**Review by stefaandevylder — APPROVED** [review]

File: src/components/box/Box.tsx (hunk 3 above)
Comment (ArnaudWeyts) [reply src/components/box/Box.tsx:68]: Yeah I've looked into it a bit but also couldn't find a clean solution 😕

File: src/components/box/Box.tsx (hunk 2 above)
Comment (ArnaudWeyts) [reply src/components/box/Box.tsx]: Good suggestion, was planning on providing good generics in a separate PR, created something in the backlog for it

**Review by lowiebenoot — APPROVED** [review]

File: src/components/box/Box.tsx [hunk 4]
```diff
@@ -14,129 +14,127 @@ const borderRadiuses = {
   rounded: '4px',
 };
 
-class Box extends PureComponent {
-  getBorder = (value) => {
-    const { borderColor, borderTint } = this.props;
+const Box = (props) => {
```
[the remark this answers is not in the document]
lowiebenoot: Guess it's not really necessary to have these as PureComponents.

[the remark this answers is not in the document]
ArnaudWeyts: Indeed, I would prefer to not use memoized components by default, only when optimizing 🙈

PR #2092, opened 2022-04-20
Make teal default color in Text elements

qubis741:
### Changed

- `Text`: Default value of `color` property is `teal` ([@qubis741](https://github.com/qubis741) in [#2092](https://github.com/teamleadercrm/ui/pull/2092))


qubis741:
Wont do

PR #2115, opened 2022-05-10
[GUILD] Typescript utils

jelledc:
### Description
Converted `utils/getOS` and `utils/useFocusTrap` files to typescript.


driesd (approved):

CHANGELOG.md
```diff
@@ -4,6 +4,8 @@
 
 ### Changed
 
+- [INTERNAL]: Converted utils to typescript. ([@jelledc](https://github.com/jelledc) in [#2115](https://github.com/teamleadercrm/ui/pull/2115))
```
ArnaudWeyts, line 7: I don't think we need to mention internal things like these in the changelog, since a library user will never be affected by it

ArnaudWeyts (dismissed):

CHANGELOG.md
driesd: Yeah that actually makes sense indeed... So we won't be updating the Changelog for TypeScript conversions. But we will update it when we remove the auto blurring logic. 👍

stefaandevylder (dismissed):

stefaandevylder (approved):

PR #2123, opened 2022-05-10
[GUILD] Typescript Button

driesd:
This PR converts the `Button` component to TypeScript. It also gets rid of the auto blur functionality.


src/components/button/Button.tsx
```diff
@@ -22,33 +21,58 @@ const textComponentMap = {
...
+  /** If true, component will take the full width available. */
+  fullWidth?: boolean;
+  /** The icon displayed inside the button. */
+  icon?: ReactNode;
+  /** The position of the icon inside the button. */
+  iconPlacement?: 'left' | 'right';
+  /** The textual label displayed inside the button. */
+  label?: string;
+  /** Determines which kind of button to be rendered. */
+  level?: BUTTON_LEVELS;
```
lowiebenoot, line 46: You're gonna have to "cast' it to a type, otherwise you can't use ith with string literals (`<Button level="primary" />` is going to have an error).

src/components/button/Button.tsx
```diff
@@ -22,33 +21,58 @@ const textComponentMap = {
   large: UITextDisplay,
 };
 
-/** @type {React.ComponentType<any>} */
+interface Props {
```
lowiebenoot, line 24: I think it should extend BoxProps, since it's using Box internally and spreading rest props on it. I just realised this. Need to do it myself as well for the Avatar component.

lowiebenoot (approved):

BeirlaenAaron (approved):

PR #2130, opened 2022-05-11
[GUILD] Typescript Dialog

driesd:
This PR converts our `Dialog`, `DialogBase`, and their child components to TypeScript.


src/components/dialog/Body.tsx
```diff
@@ -1,8 +1,9 @@
-import React, { ReactNode, forwardRef } from 'react';
+import React, { ReactNode, forwardRef, RefObject } from 'react';
 
 import { Box } from '../../index';
+import { BoxProps } from '../box/Box';
 
-interface BodyProps {
+interface BodyProps extends Omit<BoxProps, 'ref'> {
```
ArnaudWeyts, line 6: I would type the ref explicitly instead of extending from Box, since it's not always a Box, but can also be a div

src/components/dialog/DialogBase.tsx
```diff
@@ -81,7 +81,12 @@ export const DialogBase = ({
             onEscKeyDown={onEscKeyDown}
           >
             <FocusRing>
-              <Element ref={ref} data-teamleader-ui="dialog" className={dialogClassNames} {...(form && { onSubmit })}>
+              <Element
+                ref={ref as RefObject<HTMLFormElement> & RefObject<HTMLDivElement>}
```
ArnaudWeyts, line 86: This needs to be an `|` operator instead of an `&`, no?

package.json
```diff
@@ -87,6 +87,7 @@
     "@types/lodash.omit": "^4.5.7",
     "@types/lodash.without": "^4.4.6",
     "@types/react": "^17.0.45",
+    "@types/react-dom": "^18.0.4",
```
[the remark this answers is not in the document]
lowiebenoot: afaik it uses the correct version automatically (/cc @lorgan3). It contains the types for other versions as well and does some magic I guess.
https://github.com/DefinitelyTyped/DefinitelyTyped/tree/master/types/react-dom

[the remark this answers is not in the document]
ArnaudWeyts: Aha 💡 In that case we should probably still update @types/react?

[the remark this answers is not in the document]
lorgan3: I _think_ that's how it works, though I would actually prefer keeping on the 17.X.X version.

src/components/dialog/Body.tsx
BeirlaenAaron: Thats what we're doing now right? We're extending everything from BoxProps except ref, and then do `as RefObject<HTMLDivElement>` on the div's ref

src/components/dialog/DialogBase.tsx
BeirlaenAaron: I've changed the way this element is rendered into a form or div a bit, since TS really wanted the ref's type to be `RefObject<HTMLFormElement> & RefObject<HTMLDivElement>`

JorenSaeyTL (dismissed):

src/components/dialog/Body.tsx
ArnaudWeyts: Ah yes you're right, I got confused 🙈

qubis741 (approved):

lorgan3 (approved):

ArnaudWeyts:
✅ Manually checked

ArnaudWeyts (approved):

BeirlaenAaron (approved):

PR #2131, opened 2022-05-11
Add missing prop for box component

eniskraasniqi1:
### Description

Added the missing prop `to` for box component and also option `grid` for display, needed for upgrading dependencies in `service-auth-frontend` `react 17.0.2`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,7 @@
 ## [unreleased]
+### Fixed
+
+- `Box`: Added missing props for Box Component ([@eniskrasniqi1](https://github.com/eniskraasniqi1) in [#2131](https://github.com/teamleadercrm/ui/pull/2131))
 
 ### Added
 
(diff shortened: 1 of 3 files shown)


driesd (dismissed):

CHANGELOG.md
```diff
@@ -10,6 +10,12 @@
 
 ### Dependency updates
 
+## [14.5.1] - 2022-05-11
```
qubis741, line 13: you should place your change in top part of changelog (unreleased). Here it goes when UI library is updated and deployed

src/components/box/Box.tsx
```diff
@@ -65,6 +65,7 @@ export type BoxProps = Partial<
     paddingTop: Padding;
     style: CSSProperties;
     textAlign: 'center' | 'left' | 'right';
+    [key: string]: unknown;
```
qubis741, line 68: its safer to explicitly set props, so I would say use `to: string`

lorgan3: The problem with that is the `element` prop can be used for both html elements and React components.
So if you have a component `Foo` that accepts a prop `bar: string`
and you do `<Box element={Foo} />` it's not possible to add the required `bar` prop.

lorgan3 (approved):

lorgan3 (approved):

stefaandevylder (approved):

lowiebenoot (approved):

PR #2132, opened 2022-05-11
[GUILD] Typescript `DatePicker`

BeirlaenAaron:
This PR converts the `DatePicker` component to typescript


src/components/datepicker/DatePicker.tsx
```diff
@@ -0,0 +1,125 @@
...
+import MonthPicker from './MonthPicker';
+import { convertModifiersToClassnames } from './utils';
+import cx from 'classnames';
+import theme from './theme.css';
+import uiUtilities from '@teamleader/ui-utilities';
+import LocaleUtils from './localeUtils';
+
+type Size = 'small' | 'medium' | 'large';
+
+interface DatePickerProps {
```
lowiebenoot, line 15: It should extend from Box props and from the DayPicker props

src/components/datepicker/DatePickerInput.tsx
```diff
@@ -0,0 +1,197 @@
...
+import DatePicker from '.';
+import Icon from '../icon';
+import Input from '../input';
+import Popover from '../popover';
+import cx from 'classnames';
+import theme from './theme.css';
+import { formatDate } from './localeUtils';
+import { IconCalendarSmallOutline } from '@teamleader/ui-icons';
+import { DateTimeFormatOptions } from 'luxon';
+import { InputProps } from 'react-select';
```
lowiebenoot, line 12: This should be our own Input props from `'../input'` (which doesn't exist yet, I guess). Maybe just use `TSFixMe` and add a todo.

src/components/datepicker/DatePickerInput.tsx
```diff
@@ -0,0 +1,197 @@
...
+    if (openPickerOnFocus) {
+      setPopoverAnchorEl(event.currentTarget);
+      setIsPopoverActive(true);
+      inputProps.onFocus && inputProps.onFocus(event);
+    } else {
+      inputProps.onFocus && inputProps.onFocus(event);
+    }
+  };
+
+  const handleInputClick = (event: React.MouseEvent<HTMLInputElement> | React.FocusEvent<HTMLInputElement>) => {
```
lowiebenoot, line 110: It will never be a FocusEvent, right?

src/components/datepicker/DatePickerInput.tsx
```diff
@@ -0,0 +1,197 @@
...
+    if (!openPickerOnFocus) {
+      setPopoverAnchorEl(event.currentTarget);
+      setIsPopoverActive(true);
+      inputProps.onFocus && inputProps.onFocus(event as React.FocusEvent<HTMLInputElement>);
```
lowiebenoot, line 114: Don't think we should to this "casting", cause you cast it to something it isn't.
I guess it complains now because you're using the InputProps from react-select.
If we use our own, we can actually change the onFocus handler types, and use `event: React.MouseEvent<HTMLInputElement> | React.FocusEvent<HTMLInputElement>`

src/components/datepicker/DatePickerInput.tsx
```diff
@@ -0,0 +1,197 @@
...
+import Input from '../input';
+import Popover from '../popover';
+import cx from 'classnames';
+import theme from './theme.css';
+import { formatDate } from './localeUtils';
+import { IconCalendarSmallOutline } from '@teamleader/ui-icons';
+import { DateTimeFormatOptions } from 'luxon';
+import { InputProps } from 'react-select';
+
+interface DatePickerInputProps {
```
lowiebenoot, line 14: Needs to extend from BoxProps.
All of the props are now required as well. Pretty sure they are not.

src/components/datepicker/MonthPicker.tsx
```diff
@@ -1,13 +1,31 @@
 import React, { useMemo, useState, useEffect } from 'react';
-import PropTypes from 'prop-types';
 
 import Box from '../box';
 import { Select } from '../select';
 import { NumericInput } from '../input';
 import theme from './theme.css';
+import { LocaleUtils } from 'react-day-picker';
+
+interface MonthPickerProps {
```
lowiebenoot, line 9: Not sure, but I don't think any of the props is required actually.

lowiebenoot (changes requested):

lowiebenoot (approved):

src/components/datepicker/localeUtils.ts
```diff
@@ -1,7 +1,7 @@
 import { DateTime, Info } from 'luxon';
 
 const defaultLocale = 'en-GB';
-const firstDayOfWeek = {
+const firstDayOfWeek: Record<string, number> = {
```
ArnaudWeyts, line 4: PP: use `as const` instead

```tsx
const firstDayOfWeek = {
 'da-DK': 1,
  'de-DE': 1,
  'de-DE': 1,
  'en-GB': 1,
  'en-GB': 1,
  'en-US': 0,
  'en-US': 0,
  'es-ES': 1,
  'es-ES': 1,
  'fi-FI': 1,
  'fi-FI': 1,
  'fr-BE': 1,
  'fr-BE': 1,
  'fr-FR': 1,
  'fr-FR': 1,
  'it-IT': 1,
  'it-IT': 1,
  'nb-NO': 1,
  'nb-NO': 1,
  'nl-BE': 1,
  'nl-BE': 1,
  'nl-NL': 1,
  'nl-NL': 1,
  'pl-PL': 1,
  'pl-PL': 1,
  'pt-PT': 1,
  'pt-PT': 1,
  'sv-SE': 1,
  'sv-SE': 1,
  'tr-TR': 1,
  'tr-TR': 1,
} as const;
```

BeirlaenAaron: I can't find a way to fix this when using `as const`

image

ArnaudWeyts: locale will have to be typed as `keyof firstDayOfWeek` then 🙈 Probably best to put all the locales in a different type

lorgan3:
Looks like a valid failure to me. There aren't more days (it's the same month) but something seems to have broken with the 'first day of week' part. In the screenshots Sunday is suddenly the first day instead of Monday.

qubis741 (dismissed):

src/components/datepicker/localeUtils.ts
```diff
@@ -59,10 +59,30 @@ export const formatWeekdayLong = (dayOfWeekNumber: number, locale = defaultLocal
   return Info.weekdays('long', { locale })[dayIndex];
 };
 
-export const getMonths = (locale = defaultLocale) => Info.months('long', { locale });
+export const getMonths = (locale = defaultLocale) =>
+  Info.months('long', { locale }) as [
```
lorgan3, line 63: Is this absolutely necessary, does `string[]` not suffice? 😅

lorgan3 (commented):
Looks good already just some remarks for possible improvements

ArnaudWeyts (dismissed):

stefaandevylder (dismissed):

JorenSaeyTL (approved):

qubis741 (approved):

qubis741:

src/components/datepicker/DatePicker.tsx
```diff
@@ -0,0 +1,122 @@
...
+  size = 'medium',
+  withMonthPicker,
+  showWeekNumbers,
+  onChange,
+  ...others
+}) => {
+  const [selectedDate, setSelectedDate] = useState<Date>();
+  const [selectedMonth, setSelectedMonth] = useState<Date>();
+
+  useEffect(() => {
```
eniskraasniqi1, line 47: I think this useEffect can be replaced with a default state on useState 🤔, something like:

`const [selectedDate, setSelectedDate] = useState<Date>(others.selectedDate);`

src/components/datepicker/DatePickerInput.tsx
```diff
@@ -0,0 +1,198 @@
...
+  /** If true, component will be rendered in inverse mode. */
+  inverse?: boolean;
+  /** The language ISO locale code ('en-GB', 'nl-BE', 'fr-FR',...). */
+  locale?: string;
+  /** Callback function that is fired when the date has changed. */
+  onChange?: (selectedDate: Date) => void;
+  /** Callback function that is fired when the popover with the calendar gets closed (unfocused) */
+  onBlur?: () => void;
+  /** Object with props for the Popover component. */
+  popoverProps?: object;
```
eniskraasniqi1, line 35: Probably putting this as TSFixMe would be better, so we don't forget after we transform Popover to TS.

src/components/datepicker/WeekDay.tsx
```diff
@@ -0,0 +1,25 @@
...
+interface WeekDayProps {
+  weekday?: number;
+  className?: string;
+  localeUtils?: LocaleUtils;
+  locale?: string;
+  size?: Size;
+}
+
+const WeekDay: GenericComponent<WeekDayProps> = ({ weekday, className, localeUtils, locale, size }) => {
+  const weekdayName = localeUtils?.formatWeekdayLong(weekday!, locale);
```
stefaandevylder, line 16: Weekday is nullable, you can't force it here

src/components/datepicker/datePicker.stories.spec.tsx
```diff
@@ -10,15 +10,15 @@ export default {
...
+    {(['small', 'medium', 'large'] as const).map((size) => (
       <DatePicker key={size} size={size} marginBottom={4} selectedDate={new Date('1990-07-30')} onChange={() => {}} />
     ))}
   </div>
 );
 
 export const DatePickerInputStory = () => (
   <div>
-    {['small', 'medium', 'large'].map((size) => (
+    {(['small', 'medium', 'large'] as const).map((size) => (
```
stefaandevylder, line 21: If you use it twice, just create a const for it

src/components/datepicker/datePicker.stories.tsx
```diff
@@ -110,9 +111,9 @@ export const inputSingleDate = () => {
       locale={select('Locale', languages, 'nl-BE')}
       onChange={handleOnChange}
       selectedDate={preSelectedDate}
-      size={select('Size', sizes, 'medium')}
-      inputSize={select('Input size', optionalSizes, null)}
-      datePickerSize={select('Date picker size', optionalSizes, null)}
+      size={select('Size', sizes, 'medium') as 'small' | 'medium' | 'large'}
```
stefaandevylder, line 114: Lets type this

stefaandevylder (changes requested):
See comment about weekday

src/components/datepicker/DatePickerInput.tsx
```diff
@@ -0,0 +1,189 @@
...
+  const [popoverAnchorEl, setPopoverAnchorEl] = useState<EventTarget | null>(null);
+  const [selectedDate, setSelectedDate] = useState<Date | undefined>(others.selectedDate);
+
+  const getFormattedDate = () => {
+    if (!selectedDate) {
+      return '';
+    }
+
+    if (!customFormatDate) {
+      return formatDate(selectedDate, locale as DateTimeFormatOptions);
```
eniskraasniqi1, line 82: Q: Is `as DateTimeFormatOptions` needed here ?

BeirlaenAaron: Not anymore apparently

BeirlaenAaron:
Input single date story somehow broke after the inputs were converted to typescript

BeirlaenAaron:
> Input single date story somehow broke after the inputs were converted to typescript

The problem was here that `value` was added to the props in the `Input` component, which was not the case in the old code so the value kept being set as it's initial value.

qubis741 (approved):

qubis741:
Manual check failed :rotating_light:, discussed in slack

qubis741 (approved):

qubis741:

PR #2138, opened 2022-05-13
Fix icon menu

lorgan3:
### Description

It seems that in React 17 a click renders the menu, which then receives the same click again causing the menu to be hidden again right away.

### Manual check
 - Check you can open and close the `IconMenu` component

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,8 @@
 
 ### Fixed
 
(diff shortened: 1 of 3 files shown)


qubis741 (approved):

farazatarodi (approved):

lowiebenoot (approved):

lowiebenoot:
Checked ✅

PR #2143, opened 2022-05-17
Toggle button to TS and make width adaptive

qubis741:
### Changed

- `Toggle`: Changed width from full to adaptive


JorenSaeyTL (dismissed):
MCing ⌛  ...

JorenSaeyTL:
I'll read that branch name before I'll go to bed 😅

JorenSaeyTL:
MC ✅ 
- Toogle still works
- When I make the label `width: 100%` I can still click on the white space of the label to change the toggle, so my question:
Is it ok that the label is clickable to change toggle? Perhaps something to check off with Stephen and someone from QA?

src/components/toggle/Toggle.tsx
```diff
@@ -1,105 +1,93 @@
...
+  disabled = false,
+  maxLines,
+  className,
+  label,
+  onChange,
+  size = 'medium',
+  ...others
+}: ToggleProps) => {
+  const ref = useRef<HTMLInputElement>(null);
+  const restProps = omit(others, ['onChange']);
```
lorgan3, line 36: I don't think this `omit` does anything, `onChange` is already destructured on line 31

src/components/toggle/theme.css
```diff
@@ -22,8 +22,10 @@
 
 .toggle {
   align-items: flex-start;
-  display: flex;
+  display: inline-flex;
   user-select: none;
+  /* Fix for whitespace bellow, caused be inline-flex */
```
lorgan3, line 27:
```suggestion
  /* Fix for whitespace below, caused be inline-flex */
```

qubis741:
> MC ✅
> 
> * Toogle still works
> * When I make the label `width: 100%` I can still click on the white space of the label to change the toggle, so my question:
>   Is it ok that the label is clickable to change toggle? Perhaps something to check off with Stephen and someone from QA?
>   ⚠️

Yes that is desired outcome

JorenSaeyTL (approved):

lorgan3 (approved):

PR #2170, opened 2022-05-25
Email selector update

qubis741:
https://teamleader.atlassian.net/browse/FRAF-644

### Added

- `EmailSelector`: `warning` property and submit on SPACE key 

### Manuel Check

- You can pass `warning` prop that overrides internal `warning`
- Email can be also submitted with spacebar

### DIFF
--- a/src/components/emailSelector/Label.tsx
+++ b/src/components/emailSelector/Label.tsx
@@ -19,7 +19,6 @@ const UP_ARROW = 'ArrowUp';
 const DOWN_ARROW = 'ArrowDown';
 const SEMI = ';';
 const COMMA = ',';
-const SPACE = ' ';
 
 interface LabelProps {
   option: Suggestion;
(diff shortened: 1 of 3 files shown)


src/components/emailSelector/Label.tsx
```diff
@@ -132,6 +133,7 @@ const Label = ({
 
         case COMMA:
         case SEMI:
+        case SPACE:
```
lorgan3, line 136: The reason we didn't do this on space is because you can also type a customer name (first and last name or if the company name has spaces) and then you can select the email from the autocomplete

Could you double check with design it is ok to remove this or perhaps add an option for this functionality?

qubis741: Good point, discussing it with design

lorgan3 (approved):

ArnaudWeyts (approved):

lowiebenoot (approved):

PR #2174, opened 2022-05-26
Convert Icon component to typescript

eniskraasniqi1:
### Description

This PR converts icon component to typescript and also from class component to functional component.

**Note when Manual Checking**:
Not sure if I should leave children props to -> _ReactElement | ReactElement[]_, since the .type property doesn't exists on _ReactNode_(string, boolean, {}...). Leaving this for discussion for better approach.


src/components/icon/Icon.tsx
```diff
@@ -0,0 +1,46 @@
...
+  /** Classname for the Icon component */
+  className?: string;
+  /** Color of the icon */
+  color?: COLOURS;
+  /** Tint of the icon */
+  tint?: TINTS;
+  /** Opacity of the icon */
+  opacity?: number;
+  /** For other props */
+  [key: string]: unknown;
```
qubis741, line 21: remove this

src/components/icon/Icon.tsx
```diff
@@ -0,0 +1,46 @@
+import React, { ReactNode } from 'react';
+import Box from '../box';
+import cx from 'classnames';
+import theme from './theme.css';
+
+export type COLOURS = 'neutral' | 'mint' | 'violet' | 'ruby' | 'gold' | 'aqua' | 'teal';
+export type TINTS = 'lightest' | 'light' | 'normal' | 'dark' | 'darkest';
+
+interface IconProps {
```
qubis741, line 9: add behind `IconProps` this: `extends Omit<BoxProps , {{props you are overriding}}>`

This is for `...others` spread that is used on line 28

src/components/icon/Icon.tsx
```diff
@@ -2,11 +2,12 @@ import React, { ReactNode } from 'react';
 import Box from '../box';
 import cx from 'classnames';
 import theme from './theme.css';
+import { BoxProps } from '../box/Box';
 
 export type COLOURS = 'neutral' | 'mint' | 'violet' | 'ruby' | 'gold' | 'aqua' | 'teal';
 export type TINTS = 'lightest' | 'light' | 'normal' | 'dark' | 'darkest';
 
-interface IconProps {
+interface IconProps extends Omit<BoxProps, 'ref'> {
```
qubis741, line 10: I would say omit types should be `children | className` since you are overriding them

qubis741 (dismissed):

JorenSaeyTL (dismissed):

src/components/icon/Icon.tsx
```diff
@@ -29,13 +29,15 @@ const Icon = ({ children, className, color = 'teal', tint = 'normal', opacity =
...
+          }
 
-        return React.cloneElement(child, {
-          opacity,
-        });
+          return React.cloneElement(child, {
+            opacity,
+          });
+        }
       })}
```
lorgan3, line 40: I think in order to keep the old behaviour you need to return the child if it's not a valid element.
And you can probably drop the `child.type` bit since there should always be a type if it's a valid element

src/components/icon/Icon.tsx
```diff
@@ -0,0 +1,45 @@
...
+
+const Icon = ({ children, className, color = 'teal', tint = 'normal', opacity = 0.84, ...others }: IconProps) => {
+  const classNames = cx(theme[color], theme[tint], className);
+
+  return (
+    <Box className={classNames} alignItems="center" data-teamleader-ui="icon" display="flex" element="span" {...others}>
+      {React.Children.map(children, (child) => {
+        // Check if child is an actual React component
+        // if so, pass the needed props. If not, just render it.
+        if (React.isValidElement(child)) {
```
farazatarodi, line 31: You are not rendering the child if it is not a React component.

farazatarodi (changes requested):

src/components/icon/Icon.tsx
eniskraasniqi1: Yeah, thanks for the remark. Dropped the child type and checked first if its not a valid element 👍

farazatarodi (dismissed):

stefaandevylder (approved):

qubis741 (approved):

PR #2185, opened 2022-05-31
[FRAF-544] Flex component

farazatarodi:
### Link to issue

[Jira](https://teamleader.atlassian.net/browse/FRAF-544)

### Description

Adds a Flex component to the UI library that will be used for layouts. For the full information and the properties, check the Jira ticket.

### Manual check

Check the storybook.

### DIFF
--- a/src/components/flex/Flex.tsx
+++ b/src/components/flex/Flex.tsx
@@ -9,13 +9,23 @@ interface FlexProps {
   direction?: 'row' | 'column';
   gap?: Gap;
   alignItems?: 'center' | 'flex-start' | 'flex-end' | 'baseline' | 'stretch';
+  justifyContent?: 'center' | 'flex-start' | 'flex-end' | 'space-around' | 'space-between' | 'space-evenly';
   style?: CSSProperties;
 }
 
-const Flex = ({ children, direction = 'row', gap = 0, alignItems, style, ...otherProps }: FlexProps) => {
+const Flex = ({
+  children,
+  direction = 'row',
+  gap = 0,
+  alignItems,
+  justifyContent,
+  style,
+  ...otherProps
+}: FlexProps) => {
   const classNames = cx(theme['flex'], theme[`${direction}`], {
     [theme[`gap-${gap}`]]: gap > 0,
     [theme[`align-items-${alignItems}`]]: alignItems,
+    [theme[`justify-content-${justifyContent}`]]: justifyContent,
   });
 
   return (
@@ -24,5 +34,7 @@ const Flex = ({ children, direction = 'row', gap = 0, alignItems, style, ...othe
     </div>
   );
 };
+
+Flex.displayName = 'Flex';
 
 export default Flex;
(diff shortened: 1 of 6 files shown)


src/components/flex/Flex.tsx
```diff
@@ -0,0 +1,43 @@
...
+export type Gap = typeof gaps[number];
+
+type FlexProps = Partial<
+  {
+    children: ReactNode;
+    direction: 'row' | 'column';
+    gap: Gap;
+    alignItems: 'center' | 'flex-start' | 'flex-end' | 'baseline' | 'stretch';
+    justifyContent: 'center' | 'flex-start' | 'flex-end' | 'space-around' | 'space-between' | 'space-evenly';
+    style: CSSProperties;
```
qubis741, line 15: Q: This is not defined in ticket, do we need it? If we want to add additional styling, wont it be more efficient to do it via `className` ?

src/components/flex/Flex.tsx
```diff
@@ -4,10 +4,11 @@ import theme from './theme.css';
 
 interface FlexProps {
   children: ReactNode;
+  direction?: 'row' | 'column';
```
lorgan3: Since this is a new component we have the option to use enums here without introducing a breaking change.
Is this perhaps something we can discuss in the ui guild when a type or an enum is preferred? (Or was this already discussed?)

src/components/flex/Flex.tsx
```diff
@@ -2,13 +2,16 @@ import React, { ReactNode } from 'react';
 import cx from 'classnames';
 import theme from './theme.css';
 
+const gaps = [0, 1, 2, 3, 4, 5, 6, 7, 8] as const;
+export type Gap = typeof gaps[number];
```
lorgan3, line 6: Q: since `gaps` is not used, why not `export type Gap = 0 | 1 | 2 | 3 | 4 | 5 | 6 | 7 | 8;` directly?

src/components/flex/Flex.tsx
farazatarodi: Let's discuss this in the standup.

src/components/flex/Flex.tsx
farazatarodi: I remember while converting some components to TS, we decided to use strings instead of enums. Importing enums into components were not logical at that point and you will still get code completion with strings.
(https://teamleader.slack.com/archives/C2T7PPV9T/p1652185340106959)

lorgan3: I think the decision there was because introducing an enum would be a breaking change for the existing components.

qubis741 (dismissed):

JorenSaeyTL (dismissed):

ArnaudWeyts (approved):

ArnaudWeyts (changes requested):
Still missing a changelog entry

ArnaudWeyts (approved):

PR #2197, opened 2022-06-10
Add ability to remove the remove button from the first item in EmailSelector

stefaandevylder:
### Description

This PR adds a prop to `EmailSelector` so you can disable the removal of the first email address.

Screenshot 2022-06-10 at 10 00 42

#### Screenshot after this PR

Screenshot 2022-06-10 at 10 00 57

### DIFF
--- a/src/components/emailSelector/Label.tsx
+++ b/src/components/emailSelector/Label.tsx
@@ -32,6 +32,7 @@ interface LabelProps {
   onFinish?: (index: number, newLabel: Suggestion | { email: string }) => void;
   suggestions?: Suggestions;
   renderSuggestion?: React.ComponentType<React.ComponentProps<typeof EmailSuggestion>>;
+  disableRemovalOfFirst?: boolean;
 }
 
 const Label = ({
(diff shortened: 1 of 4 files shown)


qubis741 (dismissed):

qubis741:
MC: You can still delete first email with backspace button, is that okay?

stefaandevylder:
> MC: You can still delete first email with backspace button, is that okay?
This is indeed not okay, I will fix this 👍🏼

src/components/emailSelector/Label.tsx
```diff
@@ -147,7 +149,7 @@ const Label = ({
           break;
 
         case BACKSPACE:
-          if (!event.repeat && content === '' && onRemove) {
+          if (!event.repeat && content === '' && onRemove && !((index === 0 || index === 1) && disableRemovalOfFirst)) {
             onRemove(index);
```
ArnaudWeyts, line 153: Why not make that check part of the onRemove function?

stefaandevylder: Because the check is different for BACKSPACE & normal remove. The backpace checks also index 1 and the normal doesnt.

eniskraasniqi1 (approved):

PR #2205, opened 2022-06-14
[TIDAL-2039] Email selector menu can be the same width as input

BeirlaenAaron:
### Description

[Jira](https://teamleader.atlassian.net/browse/TIDAL-2039)

This PR allows the autocomplete menu of the email selector to take the same width as it's input
(Also makes the space between the input and the menu a little bit smaller as requested in the ticket)

image

#### Screenshot after this PR

If the `menuFullWidth` prop is on the email selector, it should look like this:
image

### Breaking changes

N/A
It's set as an optional prop to avoid breaking changes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ### Added
 
+- `EmailSelector`: Allow the autocomplete menu to take the same width as it's input with the `menuFullWidth` prop ([@BeirlaenAaron](https://github.com/BeirlaenAaron) in [#2205](https://github.com/teamleadercrm/ui/pull/2205))
+
 ### Changed
 
 ### Deprecated
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,7 +19,7 @@
 
 ## [14.6.1] - 2022-06-10
 
-### Fixed=======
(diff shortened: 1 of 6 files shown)


stefaandevylder (changes requested):
This needs to be added to the changelog

driesd (approved):

CHANGELOG.md
```diff
@@ -18,7 +19,7 @@
 
 ## [14.6.1] - 2022-06-10
 
-### Fixed
+### Fixed=======
```
lowiebenoot, line 22: I think you did a little oopsie here.

lowiebenoot:
In which case would it not look good? I would also opt for always giving it the full width, just like our Select component does.

CHANGELOG.md
BeirlaenAaron: That is indeed an oopsie 😅

qubis741 (approved):

lorgan3 (approved):

jelledc (approved):

lowiebenoot (approved):

stefaandevylder:
MC'ing so UI can be deployed

PR #2214, opened 2022-06-17
Grid: columnGap and rowGap props

BeirlaenAaron:
### Description

This adds two props to set the column and row gap separately

image

#### Screenshot after this PR

image

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ### Added
 
+- `Grid`: Add two props to set column and row gap separately ([@BeirlaenAaron](https://github.com/BeirlaenAaron) in [#2214](https://github.com/teamleadercrm/ui/pull/2214))
+
 ### Changed
 
 ### Deprecated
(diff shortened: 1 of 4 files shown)


jelledc (approved):

lorgan3 (approved):

stefaandevylder (approved):

lowiebenoot (approved):

JorenSaeyTL (approved):

ArnaudWeyts (approved):

ArnaudWeyts:
> PP: I think the `Gap` type might be too restrictive for these (I can see percentages or `em`/`rem` being useful as well), but I'm not sure because depending how you look at it, that restriction might be a good thing.

@jelledc Let's keep this strict for now, as it would be unfortunate if everyone starts using custom spacing, I haven't run into a use case where we wanted to use a percentage/rem/em margin

That being said, our spacer units should probably be defined in rem/em instead of pixels 🙈

PR #2233, opened 2022-06-28
[GUILD] Typescript `hoc`

BeirlaenAaron:
This PR converts `hoc` to TS


stefaandevylder (approved):

src/components/hoc/DocumentObjectProvider/DocumentObjectProvider.tsx
```diff
@@ -1,15 +1,17 @@
-import React, { PureComponent, createContext } from 'react';
+import React, { PureComponent, createContext, ReactNode, ComponentType } from 'react';
 
-export const Context = createContext();
+export const Context = createContext({});
 
-const DocumentObjectProvider = (WrappedComponent) => {
+const DocumentObjectProvider = (WrappedComponent: ComponentType, documentRef: ReactNode) => {
```
qubis741, line 5: I dont think `ReactNode` is right type, as it also allows `string` and `number`. Maybe `ReactElement` is better?

BeirlaenAaron: `ReactElement` isn't right, but there's a `Document` type which is the type of `node.ownerDocument` on line 23

qubis741 (approved):

stefaandevylder (approved):

PR #2247, opened 2022-06-30
Polymorphic `Box` component

jelledc:
Basically just a POC to implement polymorphic component typings on our `Box` component.
Based on https://www.freecodecamp.org/news/build-strongly-typed-polymorphic-components-with-react-and-typescript.


jelledc:
Closing because this was too complex and didn't seem to solve the issues we have with the current implementation.

PR #2255, opened 2022-07-01
[GUILD] Rewrite `SplitButton` in TS

KristofColpaert:
### Description

PR for the Typescript rewrite of the `SplitButton`.

### Manual check 

Please check Storybook.


src/components/utils/isReactElement.ts
```diff
@@ -0,0 +1,6 @@
+import { ReactElement } from 'react';
+
+export default (node: unknown): node is ReactElement => {
```
eniskraasniqi1, line 3: Q: Doesn't React.isValidElement already check this ?

src/components/splitButton/SplitButton.tsx
```diff
@@ -0,0 +1,117 @@
...
+import { Menu } from '../menu';
+import Popover from '../popover';
+import { IconChevronDownSmallOutline } from '@teamleader/ui-icons';
+import { BoxProps } from '../box/Box';
+import { GenericComponent } from '../../@types/types';
+import isReactElement from '../utils/isReactElement';
+
+export type Level = 'primary' | 'secondary' | 'destructive';
+export type Size = 'small' | 'medium' | 'large';
+interface SplitButtonProps extends Omit<BoxProps, 'children' | 'size'> {
```
eniskraasniqi1, line 14: size is redundant here I think

src/components/splitButton/SplitButton.tsx
```diff
@@ -0,0 +1,117 @@
...
+  onSecondaryButtonClick,
+  disabled,
+  ...others
+}) => {
+  const [buttonLabel, setButtonLabel] = useState<string>();
+  const [popoverActive, setPopoverActive] = useState<boolean>(false);
+
+  const popoverAnchorEl = useRef<HTMLElement | null>(null);
+
+  useEffect(() => {
```
stefaandevylder, line 43: Is it really needed to do a whole side effect for this?

src/components/utils/isReactElement.ts
jelledc: @eniskraasniqi1 This is basically a type guard, which serves a different purpose than `isValidElement`. I don't think that one can be used in the same way like this one.

KristofColpaert: Yes, you're right. Thanks for the advise! I've updated the PR ;)

src/components/splitButton/SplitButton.tsx
KristofColpaert: Size could be in BoxProps HTMLProps ;)

qubis741 (approved):

stefaandevylder (dismissed):

qubis741 (approved):

eniskraasniqi1:
MC`ing ⏳

eniskraasniqi1 (approved):
Manual check looks good ✅

PR #2256, opened 2022-07-01
[Guild] Fix Storybook rendering of string | number union type

KristofColpaert:
### Description

PR that defines how Storybook should render the string | number union type in input.stories.tsx.

### Manual check

Check if value field is text input instead of object input.


qubis741 (approved):

stefaandevylder (approved):

eniskraasniqi1 (approved):
MC`d ✅

PR #2273, opened 2022-07-11
Added exception for ckEditor elements in useFocusTrap

qubis741:
https://teamleader.atlassian.net/browse/FRAF-764

### Changed

- `useFocusTrap`: added exception for `ckEditor` elements to allow focusing 

CkEditors `LinkDialog` elements weren't focusable because of `focusTrap` used in `Dialog`.

### Manual check 
N/A

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,8 @@
 
 ### Changed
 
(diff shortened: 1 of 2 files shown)


farazatarodi (approved):

lowiebenoot (approved):
I guess there's not really another option since we don't have control over ckeditor itself? 😞

qubis741:
> I guess there's not really another option since we don't have control over ckeditor itself? 😞

No, its ugly af, but I haven't come up with other way. Someone does, please let me know

driesd (approved):

PR #2281, opened 2022-07-12
Fix bug on popover

eniskraasniqi1:
### Description
This is a bug introduced in the guild week, now timetracking is not working on core.

`getPosition` function was never using the `direction` prop. I checked history before TS conversion, it was always getting `position` as first argument instead of supposed `direction`.

##### Screenshot of Popover in javascript

Screenshot 2022-07-13 at 00 27 54

##### `isPositionPossible` function (which was used only on the getPosition func):

Screenshot 2022-07-13 at 00 28 38
This condition always returned false, so the isVerticalDirectionPositionPossible was never used.

### Manual check

- link this branch locally with core
- time tracking on the top right should now work

### DIFF
--- a/src/components/popover/positionCalculation.ts
+++ b/src/components/popover/positionCalculation.ts
@@ -234,24 +234,6 @@ const getDirection = (direction: Direction, anchorPosition: PositionValues, popo
   return direction;
 };
 
-const isVerticalDirectionPositionPossible = (
-  position: Position,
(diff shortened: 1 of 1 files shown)


qubis741 (approved):

qubis741:
Manually checking :hourglass:

qubis741:

lorgan3 (approved):
💡

PR #2286, opened 2022-07-15
[FRAF-772] Add message prop in Select for replacing default "No options" label

eniskraasniqi1:
### Description

The default "No options" Label is not translatable, so I added a prop which accepts a string and replaces the default one.

### DIFF
--- a/src/components/select/Select.js
+++ b/src/components/select/Select.js
@@ -363,8 +363,19 @@ class Select extends PureComponent {
   });
 
   render() {
-    const { components, creatable, error, inverse, helpText, size, success, warning, forwardedRef, ...otherProps } =
-      this.props;
+    const {
+      components,
+      noOptionsMessage = 'No options',
+      creatable,
+      error,
+      inverse,
+      helpText,
+      size,
+      success,
+      warning,
+      forwardedRef,
+      ...otherProps
+    } = this.props;
 
     const boxProps = pickBoxProps(otherProps);
     const restProps = omitBoxProps(otherProps);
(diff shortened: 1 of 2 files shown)


lorgan3 (changes requested):
B: All rest props are already passed to the select so this was already possible.
It's fine to explicitly document this but you shouldn't change the type (from function to string)

You can find some examples here https://github.com/search?q=org%3Ateamleadercrm+noOptionsMessage&type=code

eniskraasniqi1:
> B: All rest props are already passed to the select so this was already possible. It's fine to explicitly document this but you shouldn't change the type (from function to string)
> 
> You can find some examples here https://github.com/search?q=org%3Ateamleadercrm+noOptionsMessage&type=code

Oh shit, you're absolutely right... Thank you, will close this one 🙏

src/components/select/Select.js
```diff
@@ -363,8 +363,19 @@ class Select extends PureComponent {
   });
 
   render() {
-    const { components, creatable, error, inverse, helpText, size, success, warning, forwardedRef, ...otherProps } =
-      this.props;
+    const {
+      components,
+      noOptionsMessage = 'No options',
```
jelledc, line 368: PP: I would drop the default value as it is already the default in `react-select`.

```suggestion
      noOptionsMessage,
```

jelledc (approved):
~Looks good, did a manual check and it works as intended.~ Ticket was closed in the mean time :see_no_evil: .

PR #2314, opened 2022-08-08
[FRAF-873] - Pass onClick to the wrapper of checkbox

eniskraasniqi1:
### Description

When using checkbox on a Datagrid Row, currently when clicking the checkbox triggers the row to be clicked also. 
This is happening because we're not passing the onClick to the container, so when trying stopPropagation it does nothing. 

### Manual check
- Link ui locally with core [PR](https://github.com/teamleadercrm/core/pull/28409)
- Go to one of the new company detail pages, and create a time tracking in the unbilled time widget
- Click Checkbox, it shouldn't trigger row click

### DIFF
--- a/src/components/checkbox/Checkbox.tsx
+++ b/src/components/checkbox/Checkbox.tsx
@@ -55,7 +55,7 @@ const Checkbox: GenericComponent<CheckboxProps> = forwardRef<HTMLInputElement, C
     const IconCheckmark = size === 'large' ? IconCheckmarkMediumOutline : IconCheckmarkSmallOutline;
 
     const restProps = omit(others, ['onChange']);
-    const boxProps = pickBoxProps(restProps);
+    const boxProps = { ...pickBoxProps(restProps), onClick: restProps.onClick };
     const inputProps = omitBoxProps(restProps);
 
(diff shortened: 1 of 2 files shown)


src/components/checkbox/Checkbox.tsx
```diff
@@ -55,7 +55,7 @@ const Checkbox: GenericComponent<CheckboxProps> = forwardRef<HTMLInputElement, C
     const IconCheckmark = size === 'large' ? IconCheckmarkMediumOutline : IconCheckmarkSmallOutline;
 
     const restProps = omit(others, ['onChange']);
-    const boxProps = pickBoxProps(restProps);
+    const boxProps = { ...pickBoxProps(restProps), onClick: restProps.onClick };
```
lowiebenoot, line 58: onClick will still also be in the inputProps. But I guess we can ignore that, since the input isn't actually visible/clickable.

lowiebenoot (approved):

JorenSaeyTL (approved):

KristofColpaert:
Doing the MC

qubis741 (approved):

PR #2339, opened 2022-08-26
[FRAF-791] - Pass other props from DialogBase to Overlay

eniskraasniqi1:
### Issue

https://teamleader.atlassian.net/browse/FRAF-791

### Description

There has been introduced a bug ([Commit](https://github.com/teamleadercrm/ui/commit/cccf75a10e89c7605917d420c274dfe1176621de)) with the overlapping of z-index between new dialog and legacy one. 
Since we don't have z-index stack order for this, and changing it back will bring the issue which Arnaud mentioned on the commit, I went for passing the style prop individually in case if needed.

This PR allows other props being passed on Dialog -> Overlay, so then custom z-index can be passed.

before

#### Screenshot after this PR

After

### DIFF
--- a/src/components/dialog/DialogBase.tsx
+++ b/src/components/dialog/DialogBase.tsx
@@ -60,6 +60,7 @@ export const DialogBase: DialogBaseComponent = ({
   dragHandleRef,
   form,
   onSubmit,
+  ...others
 }) => {
   const { ref, FocusRing } = useFocusTrap({ active, initialFocusRef });
   useDraggable({ active, dragTargetRef: ref, dragHandleRef });
@@ -86,6 +87,7 @@ export const DialogBase: DialogBaseComponent = ({
             className={theme['overlay']}
             onOverlayClick={onOverlayClick}
             onEscKeyDown={onEscKeyDown}
+            {...others}
           >
             <FocusRing>
               <Box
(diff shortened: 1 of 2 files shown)


farazatarodi:
Q: Can't you pass a z-index to the file popup itself?

eniskraasniqi1:
> Q: Can't you pass a z-index to the file popup itself?

Hmm maybe I could, but since its used everywhere thats why I didn't go for it. Unless you think its better to use a big number on it 🤔

lowiebenoot:
Why are those 2 popups visible at the same time?

eniskraasniqi1:
> Why are those 2 popups visible at the same time?

I guess the file popup opens up on top of everything, so it was like that until now as well. Do you suggest that we dispatch a close event for the new dialogs if the file popup is opened ?

farazatarodi:
> > Why are those 2 popups visible at the same time?
> 
> I guess the file popup opens up on top of everything, so it was like that until now as well. Do you suggest that we dispatch a close event for the new dialogs if the file popup is opened ?

If the popup can be closed with a dispatch that would be the better option. The best would be to disable the file drag functionality when a popup is open.

eniskraasniqi1:
> > > Why are those 2 popups visible at the same time?
> > 
> > 
> > I guess the file popup opens up on top of everything, so it was like that until now as well. Do you suggest that we dispatch a close event for the new dialogs if the file popup is opened ?
> 
> If the popup can be closed with a dispatch that would be the better option. The best would be to disable the file drag functionality when a popup is open.

Yeah these two seem more of a valid decision, so fixing it individually one by one doesn't seem like an option right now. Thank you guys, will close this for now!

PR #2344, opened 2022-09-04
[FRAF-909] - Fix bug (first item getting focused on menu item)

eniskraasniqi1:
#### Link - https://teamleader.atlassian.net/browse/FRAF-909

Menu item was getting focused on initial render, since the focus selector wasn't working correctly. According to this [stackoverflow answer](https://stackoverflow.com/a/49329527/10006363), in order for focus to work correctly on an `li` element it should have `tabIndex="-1"`.

(EDIT) - Went for the invisible element solution.

Screen Shot 2022-09-04 at 18 39 47

#### Screenshot after this PR

Screen Shot 2022-09-04 at 18 41 22

### Manual check
- Link core with ui locally
- Go to an invoice detail page
- Click the 3 dots icon, the first item should not be focused.

### DIFF
--- a/src/components/menu/MenuItem.tsx
+++ b/src/components/menu/MenuItem.tsx
@@ -68,7 +68,7 @@ const MenuItem: GenericComponent<MenuItemProps> = ({
   const restProps = omitBoxProps(others);
 
   return (
-    <Box {...boxProps} data-teamleader-ui="menu-item" display="flex" element="li">
+    <Box {...boxProps} data-teamleader-ui="menu-item" display="flex" element="li" tabIndex={-1}>
       <Box
         onClick={handleClick}
         alignItems="center"
--- a/src/components/menu/MenuItem.tsx
+++ b/src/components/menu/MenuItem.tsx
@@ -68,7 +68,7 @@ const MenuItem: GenericComponent<MenuItemProps> = ({
   const restProps = omitBoxProps(others);
 
   return (
-    <Box {...boxProps} data-teamleader-ui="menu-item" display="flex" element="li" tabIndex={-1}>
+    <Box {...boxProps} data-teamleader-ui="menu-item" display="flex" element="li">
       <Box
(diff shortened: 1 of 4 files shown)


qubis741 (dismissed):

src/components/menu/MenuItem.tsx
```diff
@@ -68,7 +68,7 @@ const MenuItem: GenericComponent<MenuItemProps> = ({
   const restProps = omitBoxProps(others);
 
   return (
-    <Box {...boxProps} data-teamleader-ui="menu-item" display="flex" element="li">
+    <Box {...boxProps} data-teamleader-ui="menu-item" display="flex" element="li" tabIndex={-1}>
```
lorgan3, line 71: I'm not 100% convinced this is the correct solution. I don't think the `li` element needs to be focussable, just the buttons themselves which is already the case.

Is this perhaps an issue that could be solved by having different styling between hovered and focussed menu items?
Or maybe it is possible to make the `Menu` component (or an invisible component before the first menu item if that causes the menu items to be skipped) focussable. That way we keep the correct ux and we also aren't forced to focus the first button by default.

eniskraasniqi1: For the first suggestion, wasn't able to find a way with different styling for the menu items. But I could definitely try the second approach, and figure out a way if possible 👍.

eniskraasniqi1: Adding an invisible item before the other menu items, it does the trick. Will update the PR with the changes now. Thank you for the remark.

lorgan3 (approved):

stefaandevylder (approved):

lowiebenoot (approved):

lorgan3:
✅ checked
Also gave it a try by adding `tabIndex="-1"` to the Box in Menu.tsx because I was curious but it has the same effect as the invisible item.

PR #2351, opened 2022-09-09
Fix close button alignment issue

lorgan3:
### Description

Make the close-button css more specific in order to solve an issue that the transformed css is not in the same order as in development.

Doesn't seem like this requires a changelog entry since I think this is coupled to our build process in core but happy to add one if necessary.

Sharing before and after screenshot in a google doc in case this contains info that shouldn't be public yet
https://docs.google.com/document/d/1B8PC6TQWwYPaxgWwpn3X7QRabuLdqTy-nJkHo6QaOKA/edit?usp=sharing

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,7 +24,7 @@
 
 ### Dependency updates
 
-- `storybook`: Bump storybook from 6.5.9 to 6.5.10 ([@KristofColpaert](https://github.com/KristofColpaert)) in  [#2327](https://github.com/teamleadercrm/ui/pull/2327)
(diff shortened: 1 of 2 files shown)


qubis741 (approved):

JorenSaeyTL (dismissed):

jelledc (dismissed):

farazatarodi (dismissed):

BeirlaenAaron (approved):

stefaandevylder (dismissed):

lowiebenoot (dismissed):

lorgan3:
I'll have to add a changelog entry after all because I want to create a release with this fix soon 😅

eniskraasniqi1 (approved):

qubis741 (approved):

PR #2355, opened 2022-09-13
Toggle with tooltip

qubis741:
### Added

- `Toggle`: `tooltip` and `tooltipProps` props

This change was done to have matching tooltip with `Label`. Plus its reusable.

### Manual Check
 - check story


src/components/toggle/Toggle.tsx
```diff
@@ -18,6 +22,8 @@ export interface ToggleProps extends Omit<BoxProps, 'className' | 'size'> {
   label?: ReactNode;
   onChange?: (event: ChangeEvent<HTMLInputElement>) => void;
   size?: Exclude<typeof SIZES[number], 'tiny' | 'fullscreen' | 'smallest' | 'hero'>;
+  tooltip?: ReactNode;
```
farazatarodi, line 25: Should be `string` probably. Also `tooltipLabel` would be a more suitable name I think.

src/components/toggle/Toggle.tsx
```diff
@@ -82,6 +90,18 @@ const Toggle: GenericComponent<ToggleProps> = ({
...
           )}
           {children}
+          {tooltip && (
+            <TooltippedIcon
+              tooltip={<TextSmall>{tooltip}</TextSmall>}
+              tooltipSize="small"
+              color="teal"
+              tint="darkest"
+              marginLeft={1}
+              {...tooltipProps}
```
farazatarodi, line 100: I think you are already setting all of the tooltip props. Do we need the `tooltipProps`?

src/components/toggle/Toggle.tsx
qubis741: We use `ReactNode` for tooltips as there may be elements/components like links inside

src/components/toggle/Toggle.tsx
```diff
@@ -18,6 +22,8 @@ export interface ToggleProps extends Omit<BoxProps, 'className' | 'size'> {
   label?: ReactNode;
   onChange?: (event: ChangeEvent<HTMLInputElement>) => void;
   size?: Exclude<typeof SIZES[number], 'tiny' | 'fullscreen' | 'smallest' | 'hero'>;
+  tooltip?: ReactNode;
+  tooltipProps?: Record<string, any>;
```
farazatarodi, line 26: Should be the allowed tooltip prop types.

src/components/toggle/Toggle.tsx
qubis741: that is for customisation, in case we need it in codebase

src/components/toggle/Toggle.tsx
qubis741: I dont understand

src/components/toggle/toggle.stories.tsx
```diff
@@ -33,26 +28,5 @@ export const DefaultStory = (args: ToggleProps) => {
...
-        <Box display="flex">
-          Label
-          <TooltippedIcon tooltip={<TextSmall>Tooltip</TextSmall>} marginLeft={2}>
-            <IconInfoBadgedSmallFilled />
-          </TooltippedIcon>
-        </Box>
-      }
-    />
-  );
+  tooltip: 'This is the label tooltip text',
```
farazatarodi: Q: Why is the tooltip added to the default story?

qubis741: Because storybook controls set is as object, where when you want to edit it, it throws error and whole page crashes. Setting default value prevent this.

src/components/toggle/Toggle.tsx
lorgan3: I think Faraz means that this should basically be what's currently in `TooltippedComponent.propTypes` (and I agree)

I guess that means ideally `Tooltip` is also converted to typescript

src/components/toggle/Toggle.tsx
farazatarodi: Tooltips can't have links inside of them.

qubis741: for example Company Dialog needs to have it as `ReactNode` to display `<ul>`

src/components/toggle/Toggle.tsx
farazatarodi: `TooltipProps` from the tooltip component.

src/components/toggle/Toggle.tsx
farazatarodi: Ah, HTML tags. Makes sense 👍

src/components/toggle/Toggle.tsx
farazatarodi: I think if we are already setting the icon and its color, we will not be customizing the tooltip itself. Might be worth a check with the design team.

src/components/toggle/toggle.stories.tsx
farazatarodi: Got it 👍 Can we have the extra story without the tooltip then?

src/components/toggle/Toggle.tsx
qubis741: There are still lot of props that can be customized:
https://github.com/teamleadercrm/ui/blob/next-release/src/components/tooltip/Tooltip.js#L261

src/components/toggle/toggle.stories.tsx
qubis741: I dont see much of a point having another story without `tooltip` when user can control main story. In that case we should create story for each prop.

src/components/toggle/Toggle.tsx
farazatarodi: I get that and that is what I'm saying. If we are implementing the tooltip directly into the toggle, it is for consistency purposes. Then we should limit the customization because that would be against the consistency.
Now the only difference between this implementation and the original one is the fixed icon. Doesn't look like a big enough reason to change the implementation IMO. And based on this pattern I would argue that in this situation we should hardly ever change the color of the tooltip or show it on click.

qubis741: I will limit customisation props in that case, to keep consistency but allow tweaks that are used in codebase and makes sense.

qubis741:
On hold until https://github.com/teamleadercrm/ui/pull/2356/ is resolved

src/components/toggle/toggle.stories.tsx
farazatarodi: The point is that there were two stories in the beginning. And we always do showcase some of the important use cases in the stories. That is the whole point of stories actually 😅

lowiebenoot (dismissed):

src/components/toggle/toggle.stories.tsx
qubis741: I did some research and you are right, we should do more stories to show different cases, generally in whole library. Thanks for bringing that up.

KristofColpaert (approved):

stefaandevylder (approved):

eniskraasniqi1 (approved):

BeirlaenAaron (approved):

lorgan3 (approved):
✅ Looks good 

image

PR #2356, opened 2022-09-13
Guild/typescript tooltip

qubis741:
Conversion of `Tooltip` to TS

### Manual Check
Stories works


src/components/tooltip/Tooltip.tsx
```diff
@@ -0,0 +1,306 @@
...
+  tooltipShowOnClick?: boolean;
+  tooltipSize?: AllowedSize;
+  documentObject: Document;
+  tooltipShowDelay?: number;
+  /** The z-index of the Tooltip */
+  zIndex?: number;
+  tooltipActive?: boolean;
+  ComposedComponent: React.ElementType;
+  // Because its HOC, it needs to pass props of child component
+  [key: string]: any;
```
lorgan3, line 67: After many attempts, it seems like if you remove this and change the type of the `Tooltip` component you can have proper/better type checking/autocomplete

(Had to create signature for react elements and html elements)
```
function Tooltip<E extends keyof JSX.IntrinsicElements>(
  ComposedComponent: E,
): React.ComponentType<JSX.IntrinsicElements[E] & TooltipProps>;
function Tooltip<P>(ComposedComponent: React.ElementType<P>): React.ComponentType<P & TooltipProps>;

function Tooltip(ComposedComponent: React.ElementType) {
  return DocumentObjectProvider((props) => {
    ...
```

src/components/tooltip/Tooltip.tsx
```diff
@@ -0,0 +1,306 @@
...
+  tooltipActive = null,
+  ComposedComponent,
+  ...other
+}) => {
+  const tooltipRoot = useMemo(() => documentObject.createElement('div'), []);
+  const ref = useRef(null);
+  const [active, setActive] = useState(false);
+  const [position, setPosition] = useState<PositionState>({ position: tooltipPosition, top: 'auto', left: 'auto' });
+
+  const activate = (position: PositionState | undefined) => {
```
lorgan3, line 98: It's a bit hard to review since so much has changed but you had to add all the handling for null/undefined right? Would it be an option to use casts (or `!`) to remove null/undefined since apparently it never happened in the original code?

src/components/hoc/DocumentObjectProvider/DocumentObjectProvider.tsx
```diff
@@ -1,9 +1,12 @@
 import React, { PureComponent, createContext, ComponentType } from 'react';
 
-export const Context = createContext({});
+export const Context = createContext<Document | Record<string, any>>({});
```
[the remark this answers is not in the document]
qubis741: I still need to use `as Document` in this case, but this typing makes more sense than `Record`

src/components/tooltip/Tooltip.tsx
qubis741: Awesome 🔥

src/components/tooltip/Tooltip.tsx
qubis741: yes, it wasn't correctly re-implemented

src/components/avatar/Avatar.tsx
```diff
@@ -151,7 +152,8 @@ const Avatar = ({
   );
 
   const enableTooltip = tooltip && typeof fullName === 'string' && fullName.length > 0;
-  const Component = enableTooltip ? TooltippedBox : Box;
+  // @ts-ignore TS complains about non-overlapping props
+  const Component: GenericComponent<BoxProps | TooltipProps> = enableTooltip ? TooltippedBox : Box;
```
lorgan3, line 156: PP: It seems the root of the problem is the type of `tooltipProps`. I prefer casting over using `@ts-ignore`. It seems like for all the components in this commit you can solve it by removing the `@ts-ignore` and `GenericComponent<BoxProps | TooltipProps> `
and when spreading the tooltipProps `{...(tooltipProps as TooltipProps)}`

src/components/tooltip/Tooltip.tsx
```diff
@@ -92,7 +92,7 @@ const TooltippedComponent: GenericComponent<TooltippedComponentProps> = ({
   const [active, setActive] = useState(false);
   const [position, setPosition] = useState<PositionState>({ position: tooltipPosition, top: 'auto', left: 'auto' });
 
-  const activate = (position: PositionState | undefined) => {
+  const activate = (position: PositionState) => {
     if (!position) {
       return false;
     }
```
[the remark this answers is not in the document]
qubis741: facepalm

lorgan3 (approved):
🎉

qubis741:
> 🎉

Thanks for your valuable feedback and help

JorenSaeyTL (approved):

JorenSaeyTL:
MCing ⌛ ...

JorenSaeyTL:
Tooltip stories still work ✅

stefaandevylder (approved):

PR #2368, opened 2022-09-19
[TIDAL-2600] Increase marketing dialog text width

BeirlaenAaron:
### Description

The translations for the Customer Meeting interstitial dialog in core are too long to fit in the primary button.
Think the best way here is to just increase the width.
Translation should fit in the button in every language now.

image

#### Screenshot after this PR

image

### Breaking changes

N/A

### DIFF
--- a/src/components/marketingDialog/theme.css
+++ b/src/components/marketingDialog/theme.css
@@ -1,8 +1,8 @@
 .marketing-dialog {
-  width: 900px;
+  width: 1000px;
 
   &-text {
-    width: 300px;
+    width: 400px;
     display: flex;
     flex-direction: column;
   }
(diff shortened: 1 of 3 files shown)


farazatarodi (approved):

KristofColpaert (approved):

stefaandevylder (approved):

adognameddog (dismissed):

src/components/marketingDialog/theme.css
```diff
@@ -1,8 +1,8 @@
 .marketing-dialog {
-  width: 900px;
+  width: 1000px;
```
lorgan3, line 2: Q: Can we use the `hero` width (985px) instead of the override or is it still too small?

lorgan3 (approved):

src/components/marketingDialog/theme.css
BeirlaenAaron: Yes it fits :+1:

PR #2371, opened 2022-09-19
Make MarketingDialog draggable

BeirlaenAaron:
### Description

Allow the Dialog component to have no footer and convert MarketingDialog to it so it's draggable

image

#### Screenshot after this PR

[Screencast from 19-09-22 16:10:29.webm](https://user-images.githubusercontent.com/55881661/191038282-73944581-1a43-4a06-965e-d5661001c69f.webm)

### Breaking changes

N/A

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,8 @@
 ### Changed
 
 - `Tooltip`: `children` prop is optional ([@qubis741](https://github.com/qubis741)) in [#2364](https://github.com/teamleadercrm/ui/pull/2364))
+- `Dialog`: make `primaryAction` optional and show no footer if there are no action props ([@BeirlaenAaron](https://github.com/BeirlaenAaron)) in [#2371](https://github.com/teamleadercrm/ui/pull/2371))
+- `MarketingDialog`: convert from `DialogBase` to `Dialog` to make it draggable ([@BeirlaenAaron](https://github.com/BeirlaenAaron)) in [#2371](https://github.com/teamleadercrm/ui/pull/2371))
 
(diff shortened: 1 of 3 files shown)


lorgan3 (commented):
This a mostly personal preference but I would rather reimplement the dragging functionality in the marketing dialog because now we're making the `Dialog` more like the `DialogBase` which is not a good thing I think. The way I see it, the `Dialog` component is supposed to force you to be consistent (Which the marketing dialog is not)

Let's see what the others think though.

stefaandevylder:
Personally I don't see the reason why you would need to be able to drag a marketing dialog, either you interact with it or click it away 🤔 But thats just my opinion

BeirlaenAaron:
> Personally I don't see the reason why you would need to be able to drag a marketing dialog, either you interact with it or click it away thinking Any other functional dialog makes sense tho. But thats just my opinion Edit: This might be a good question for a marketeer

I noticed it was draggable in the [design](https://www.figma.com/file/aPFtL8DksaB0Js4g7m5RMk/Vectera-MVP-designs-(September-implementation)?node-id=1%3A21546) and thought I'd change it while I was at it

adognameddog (approved):

JorenSaeyTL (approved):

jelledc:
> This a mostly personal preference but I would rather reimplement the dragging functionality in the marketing dialog because now we're making the `Dialog` more like the `DialogBase` which is not a good thing I think. The way I see it, the `Dialog` component is supposed to force you to be consistent (Which the marketing dialog is not)
> 
> Let's see what the others think though.

I agree.

PR #2374, opened 2022-09-22
Fix tooltip tooltipActive prop

lorgan3:
### Description

When the tooltip was converted to a functional component there was a small change in behaviour regarding the `tooltipActive` prop. When setting it to `undefined`/`null` it used close and this was no longer the case.
I've also remove the default `null` since it's weird that `null` is not a valid type for it, but internally it can be.

https://user-images.githubusercontent.com/1833617/191789281-32313c03-6676-4779-a342-48c839b84fdc.mov

#### Screenshot after this PR

https://user-images.githubusercontent.com/1833617/191789430-70050c5e-e950-40e6-acbe-72ff5e3227ff.mov

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,8 @@
 
 ### Fixed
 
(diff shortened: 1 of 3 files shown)


qubis741 (approved):

qubis741:

BeirlaenAaron (approved):

PR #2380, opened 2022-09-26
Numeric input stepper customisation

jelledc:
### Description

Allow overriding a stepper button's `disabled` state and hooking into their `mouseDown` events. Allows fine-grained control over disabling decrease and increase stepper buttons.

https://user-images.githubusercontent.com/984441/192465833-062d4c63-3651-42e5-8cfa-4b38cb758808.mov

A manual check can be done in storybook under the playground's forms section.

### DIFF
--- a/src/components/input/NumericInput.tsx
+++ b/src/components/input/NumericInput.tsx
@@ -66,6 +66,10 @@ export interface NumericInputProps
   connectedLeft?: ReactElement;
   /** Element stuck to the right hand side of the component. */
   connectedRight?: ReactElement;
+  /** Boolean indicating whether the decrease button should be disabled. If not set, this will be derived from the min value. */
+  decreaseDisabled?: boolean;
+  /** Boolean indicating whether the increase button should be disabled. If not set, this will be derived from the max value */
+  increaseDisabled?: boolean;
   /** Boolean indicating whether the input should render as inverse. */
   inverse?: boolean;
   /** The maximum value that can be inputted. */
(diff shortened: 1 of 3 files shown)


src/components/input/NumericInput.tsx
```diff
@@ -176,6 +182,14 @@ const NumericInput: GenericComponent<NumericInputProps> = forwardRef<HTMLElement
     };
 
     const handleDecreaseMouseDown = () => {
+      if (onDecreaseMouseDown) {
+        onDecreaseMouseDown();
+
+        if (isMinReached()) {
+          return;
+        }
```
lorgan3, line 205: Q: why does this only need to happen if `onDecreaseMouseDown` is passed?

lorgan3 (approved):

src/components/input/NumericInput.tsx
jelledc: Good question. It's because in all other cases, this is already covered by the buttons becoming disabled and the min/max boundaries present in `updateStep`. Only if we hook into this by passing `onDecreaseMouseDown`, and the min/max has been reached, we should no longer do an increase or decrease, so the return is mainly there to prevent the rest of the function being executed.

src/components/input/NumericInput.tsx
```diff
@@ -220,7 +226,7 @@ const NumericInput: GenericComponent<NumericInputProps> = forwardRef<HTMLElement
       if (stepper === 'connected') {
         return (
           <Button
-            disabled={isMaxReached()}
+            disabled={typeof increaseDisabled !== 'undefined' ? increaseDisabled : isMaxReached()}
```
lowiebenoot, line 266: fyi: you could do `increaseDisabled ?? isMaxReached()` (i often forget the existence of this as well).

lowiebenoot (dismissed):

src/components/input/NumericInput.tsx
jelledc: Changed it, thanks for reminding me!

jelledc:
Rebased onto the latest next-release.

qubis741 (approved):

qubis741:

mrtus (approved):

PR #2392, opened 2022-09-29
Added smaller bottom margin for small sized label

qubis741:
### Description

`Label`: reduced bottom margin for small sized 

#### Manual check

- label story, small size has 9px margin bottom, others 12px

### DIFF
--- a/src/components/label/theme.css
+++ b/src/components/label/theme.css
@@ -1,4 +1,3 @@
-
 @import '@teamleader/ui-utilities';
 
 .is-small {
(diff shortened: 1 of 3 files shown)


src/components/label/theme.css
```diff
@@ -0,0 +1,14 @@
+
```
lorgan3, line 1: I think this file could use some linting

lorgan3 (approved):

eniskraasniqi1 (approved):

PR #2394, opened 2022-09-30
Fix escape key closing on overlays

BeirlaenAaron:
### Description

Fixed bug where pressing escape to close would no longer work after clicking inside a Dialog/Popover or any overlay component

### Breaking changes

N/A

### Manual check

Open a dialog and click inside of it, pressing escape should now close it

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,8 @@
 
 ### Fixed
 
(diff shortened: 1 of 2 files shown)


driesd (approved):

lorgan3 (dismissed):

qubis741 (dismissed):

lorgan3 (approved):
✅ seems to work

driesd (approved):

PR #2403, opened 2022-10-07
Guild/typescript interfaces and touchups

qubis741:
### Description

- all components have exported their interfaces
- Select props are temporary set as `any`, until adaptation to core is ready (there is around 40 TS errors with proper TS prop of Select). This is for case if there is need for releasing UI library until PR for adapting those changes is ready.
- Updated Select props to be easier to use 

### Manual Check 
Smoke test stories

### DIFF
--- a/src/components/select/Select.tsx
+++ b/src/components/select/Select.tsx
@@ -33,7 +33,18 @@ const minHeightBySizeMap: Record<string, number> = {
   medium: 36,
   large: 48,
 };
+
 export interface SelectProps<
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars
+  Option extends OptionType = OptionType,
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars
+  IsMulti extends boolean = false,
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars
+  IsClearable extends boolean = false,
+> {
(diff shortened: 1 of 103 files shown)


KristofColpaert (approved):

src/components/select/Select.tsx
```diff
@@ -26,18 +27,24 @@ import ValidationText from '../validationText';
...
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars
+  Option extends OptionType = OptionType,
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars
+  IsMulti extends boolean = false,
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars
+  IsClearable extends boolean = false,
+> {
+  [key: string]: any;
+}
+export interface SelectPropsFull<
```
lorgan3, line 47: Q: when does this stop being temporary? I also don't think `SelectPropsFull` is being used now

lorgan3: Ah right I read the description 😄

lorgan3 (approved):

farazatarodi (approved):

JorenSaeyTL (approved):

driesd (approved):

lowiebenoot (approved):

PR #2408, opened 2022-10-18
Label margin bottom set by default and not by size

qubis741:
### Changed

- `Label`: `marginBottom` size is same for all sizes and further customisable 

Previous implementation was blocking further customisation of `Label` `marginBottom`, for example when we want it to be 0. `9px` for small sized Labels is not needed anymore, we will stick to `Margin` sizes everywhere,

### Manual check
- all Label sizes have same marginBottom


BeirlaenAaron (approved):

lorgan3 (approved):

stefaandevylder (approved):

lowiebenoot (approved):

eniskraasniqi1 (approved):

KristofColpaert (approved):

eniskraasniqi1:
Checked ✅

PR #2409, opened 2022-10-18
Allow the left action of dialogs to be a toggle

BeirlaenAaron:
### Description

Allow the left action of dialogs to be a toggle

image

#### Screenshot after this PR
image

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ### Added
 
+- `Dialog`: added `isToggle` prop to leftAction to allow it to be a toggle ([@BeirlaenAaron](https://github.com/BeirlaenAaron)) in [#2409](https://github.com/teamleadercrm/ui/pull/2409))
+
 ### Changed
 
 ### Deprecated
--- a/src/components/dialog/Dialog.tsx
+++ b/src/components/dialog/Dialog.tsx
@@ -4,10 +4,14 @@ import omit from 'lodash.omit';
 import React, { ReactNode, useCallback, useEffect, useRef, useState } from 'react';
 import { useResizeDetector } from 'react-resize-detector';
 import { GenericComponent } from '../../@types/types';
-import { Button, ButtonGroup, DialogBase, Heading3 } from '../../index';
+import { Button, ButtonGroup, ButtonProps, DialogBase, Heading3, Toggle, ToggleProps } from '../../index';
 import { DialogBaseProps } from './DialogBase';
 import theme from './theme.css';
 import { SIZES } from '../../constants';
+
+interface LeftAction extends Omit<ButtonProps | ToggleProps, 'fullWidth' | 'marginTop'> {
+  isToggle?: boolean;
+}
 
 export interface DialogProps extends Omit<DialogBaseProps, 'ref'> {
   /** If true, the dialog will show on screen. */
@@ -17,7 +21,7 @@ export interface DialogProps extends Omit<DialogBaseProps, 'ref'> {
   /** A class name for the wrapper to apply custom styles. */
   className?: string;
   /** Object containing the the props of the action on the left (a Button). */
-  leftAction?: object;
+  leftAction?: LeftAction;
   /** Callback function that is fired when the close icon (in the header) is clicked. */
   onCloseClick?: () => void;
   /** Object containing the props of the primary action (a Button, with level prop set to 'primary'). */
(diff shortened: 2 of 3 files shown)


lorgan3 (approved):
I hadn't considered that design placed the toggle in the footer 🙈 

Q: how will you check whether the toggle is checked or not?

BeirlaenAaron:
> Q: how will you check whether the toggle is checked or not?

You can just use the `onClick` and `checked` props in leftAction

src/components/dialog/dialog.stories.tsx
```diff
@@ -45,8 +45,9 @@ export const DefaultStory: ComponentStory<typeof Dialog> = (args) => {
 
 DefaultStory.args = {
   leftAction: {
-    label: 'Remove',
-    level: 'destructive',
+    label: 'Toggle me',
```
lowiebenoot, line 48: I wouldn't change this default story as it is quite rare and more usual to use a destructive button there.

Maybe just add another story?

lowiebenoot (approved):

KristofColpaert (approved):

lorgan3 (approved):

eniskraasniqi1 (approved):

CHANGELOG.md
```diff
@@ -2,6 +2,8 @@
 
 ### Added
 
+- `Dialog`: added `isToggle` prop to leftAction to allow it to be a toggle ([@BeirlaenAaron](https://github.com/BeirlaenAaron)) in [#2409](https://github.com/teamleadercrm/ui/pull/2409))
```
JorenSaeyTL, line 5: Add this point we could just allow for left action to be any custom component we want and have the implementation take care of how it should behave

JorenSaeyTL (approved):
Left a comment, but this works as well... just wondering what we are gonna do if we have to add a checkbox next 🙈

src/components/dialog/Dialog.tsx
```diff
@@ -4,11 +4,15 @@ import omit from 'lodash.omit';
...
 import { GenericComponent } from '../../@types/types';
-import { Button, ButtonGroup, DialogBase, Heading3 } from '../../index';
+import { Button, ButtonGroup, ButtonProps, DialogBase, Heading3, Toggle, ToggleProps } from '../../index';
 import { DialogBaseProps } from './DialogBase';
 import theme from './theme.css';
 import { SIZES } from '../../constants';
 
+interface LeftAction extends Omit<ButtonProps | ToggleProps, 'fullWidth' | 'marginTop'> {
+  isToggle?: boolean;
+}
```
jelledc, line 14: PP: feels a bit weird to have the `isToggle` together with the button/toggle props.

src/components/dialog/Dialog.tsx
```diff
@@ -93,9 +97,10 @@ const Dialog: GenericComponent<DialogProps> = ({
       <DialogBase.Footer
         display="flex"
         justifyContent={leftAction ? 'space-between' : 'flex-end'}
+        alignItems="center"
         className={cx({ [theme['scroll-shadow']]: !reachedScrollEnd && showScrollShadow })}
       >
-        {leftAction && <Button {...leftAction} />}
+        {leftAction && (leftAction.isToggle ? <Toggle {...leftAction} /> : <Button {...leftAction} />)}
```
jelledc, line 103: Q: I guess typescript will complain about this, right? Because `isToggle` is not a valid prop?

jelledc, line 103: PP: I'd make a render prop out of these, because like @JorenSaeyTL already said, that would make it more future-proof.

```suggestion
        {renderLeftAction ? renderLeftAction(leftAction) : <Button {...leftAction} />)}
```

jelledc (approved):

src/components/dialog/Dialog.tsx
lorgan3: Maybe we should first loop this back to design to see what is the purpose of the `leftAction` actually.
I'm not a big fan of just allowing you to render anything in this dialog which is supposed to force consistency.

jelledc: Edited my suggestion a bit, but I do agree with your point too.

lowiebenoot: I agree, I don't think we should just allow anything.

BeirlaenAaron:
Closing this PR, design decided we won't be adding the toggle in the footer anymore.
We'll just add it inside the dialog

PR #2412, opened 2022-10-26
Fix dialog being to big on Safari

stefaandevylder:
### Description

Bug: When multiple tabs are open in safari, the save button of the new modal company add isn't visible.
This was semi-reproducible on Safari Macbook. It indeed cut off a part of the screen when opening a new tab.

The cause of the bug is because we are using `vh` and `vw` as a measurelent in CSS. This type of measurement is differently calculated on WebKit.

Screenshot 2022-10-26 at 16 37 52

#### Screenshot after this PR

Screenshot 2022-10-26 at 16 38 30

### DIFF
--- a/src/components/dialog/theme.css
+++ b/src/components/dialog/theme.css
@@ -20,7 +20,7 @@
   border-radius: var(--dialog-border-radius);
   display: flex;
   flex-direction: column;
-  max-height: 96vh;
+  max-height: calc(96vh - 20px);
   max-width: 90vw;
   opacity: 0;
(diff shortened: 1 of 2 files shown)


lorgan3 (approved):

eniskraasniqi1 (approved):

BeirlaenAaron (approved):

qubis741 (approved):

qubis741:
Manually checking :hourglass:

qubis741 (approved):

PR #2418, opened 2022-10-28
Adjustment of Wysiwyg props and export of SelectRef

qubis741:
### Description

- adjusted `Wysiwyg` props to better match reality
- exported `SelectRef` of `Select` so it's easier to work with it in codebase

### Manuel Check
N/A

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ### Added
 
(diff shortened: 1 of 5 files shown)


JorenSaeyTL (approved):
👍

driesd (approved):

PR #2430, opened 2022-11-04
Making DatePickerInput typeable (again)

qubis741:
https://teamleader.atlassian.net/browse/FRAF-1154

### Added

- `Popover`: `withFocusTrap` prop to allow control over focus ([@qubis741](https://github.com/qubis741)) in [#2430](https://github.com/teamleadercrm/ui/pull/2430))
- `DatePickerInput`: `typeable` and `errorText` props to allow typing date ([@qubis741](https://github.com/qubis741)) in [#2430](https://github.com/teamleadercrm/ui/pull/2430))

### Manual Check
Popover story
- [x] `withFocusRing` on -> focus stays in popover
- [x] `withFocusRing` off -> focus goes wherever he wants

allowedFormats = ['dd/MM/yyyy', 'd/M/yyyy', 'dd.MM.yyyy', 'd.M.yyyy', 'dd-MM-yyyy', 'd-M-yyyy']
DatePickerInput story
- [x] you can type in all formats, on blur date is selected and converted to `/` format
- [x] have selected date, select first 2 digits with mouse, try to type day `30` -> works
- [x] paste some date in allowed formats -> works
- [x] you enter invalid date, on blur -> red border
- [x] set error text in props, enter invalid date and blur -> red border + text
- [x] overall everything works


JorenSaeyTL (approved):

KristofColpaert (dismissed):

src/components/datepicker/localeUtils.ts
```diff
@@ -87,6 +87,15 @@ export const formatDate = (
 export const parseDate = (string: string, format = DateTime.DATE_SHORT as string, locale = defaultLocale) =>
   DateTime.fromFormat(string, format, { locale }).toJSDate();
 
+export const parseMultiFormatsDate = (value: string, formats: string[], locale: string) =>
+  formats.reduce((previous: Date | null, current) => {
```
lowiebenoot, line 91: A bit weird to use a reduce here. Why not use a normal `for` loop and do an early return once you have a valid date?

lowiebenoot (changes requested):
It has some weird behaviour if you want to change the date by typing, I assume because you are setting the input value (not sure).

Let's say the current filled in date is 07/11/2022. If I want to change the day and go with my cursor to the day in the input field, change 07 to 30 for example, then the cursor jumps to the end after pressing 3.

stefaandevylder (approved):

qubis741:
> It has some weird behaviour if you want to change the date by typing, I assume because you are setting the input value (not sure).
> 
> Let's say the current filled in date is 07/11/2022. If I want to change the day and go with my cursor to the day in the input field, change 07 to 30 for example, then the cursor jumps to the end after pressing 3.
> 
>  Nov-07-2022.17-38-41.mp4

thanks for thorough review. That issue is because we allow both with and without 0 padding format. So in you case, you wanted to write `30`, but since month and year was already filled and we allow both formats, after typing "3" it was `3/09/1990` which is valid and it got selected. I will try to keep cursor at same place or figure something out

stefaandevylder (dismissed):

stefaandevylder:
All comments that @lowiebenoot  left were checked and work now. ✅

src/components/datepicker/DatePickerInput.tsx
```diff
@@ -123,22 +123,19 @@ const DatePickerInput: GenericComponent<DatePickerInputProps> = ({
...
-    const value = event.target.value;
-    const date = parseMultiFormatsDate(value, ALLOWED_DATE_FORMATS, locale);
     setDisplayError(false);
-    if (date) {
-      setSelectedDate(date);
-      setInputValue(getFormattedDateString(date));
-    } else {
-      setInputValue(value);
-    }
+    const value = event.target.value;
```
lowiebenoot, line 127: Was thinking that it would be nice to keep doing `setSelectedDate(date)` in here when it's a valid date, so that you see the selected date changing in the picker while typing, and you don't need to blur it. 

```ts
  const handleInputChange = (event: React.ChangeEvent<HTMLInputElement | HTMLTextAreaElement>) => {
    setDisplayError(false);
    const value = event.target.value;
    setInputValue(value);

    const date = parseMultiFormatsDate(value, ALLOWED_DATE_FORMATS, locale);

    if (date) {
      setSelectedDate(date);
    }
  };
```

But this is not working at the moment, because you removed this part in DatePicker. 
```ts
  useEffect(() => {
    setSelectedDate(others.selectedDate);
    setSelectedMonth(others.selectedDate);
  }, [others.selectedDate]);
```

What's the reason that you removed this?

qubis741: Because setting date in `inputChange` was causing that bug you recorded on video, why it was happening is written [here](https://github.com/teamleadercrm/ui/pull/2430#issuecomment-1305492973). We want to support all allowed formats (in description of PR) + also legacy datepicker input  works similar, so those are the reason we chose this approach, together with design. Of course best way would be to support changing of date on `inputChange`, but then user has that bug, which is a no go :/

lowiebenoot: The exact reason that it was happening is not because you were setting the date, but because you were setting the inputValue. More specifically, you were setting it to something else than what the input actually had, which causes the cursor to be moved. For example, if the input value was `2/11/2022`, you parsed the date and then set the inputValue to `02/11/2022`. 

If you don't do that parsing on inputChange, it actually works fine (I have it working with the code from my previous comment). We need to be careful with executing the `onChange` handler though, I didn't look into that yet.


https://user-images.githubusercontent.com/330765/200489954-69754e78-52d8-4d7c-a6a3-e47e4805f67c.mp4

qubis741: Sorry I have expressed it incorrectly in previous comment. It is indeed because of setting inputValue, however that is desired behaviour, since we accept many formats but once date is set/picked, we want to displat it only in default format. I will try to go one more round with design, maybe we find some compromise there. Thanks

src/components/datepicker/localeUtils.ts
```diff
@@ -87,6 +87,19 @@ export const formatDate = (
 export const parseDate = (string: string, format = DateTime.DATE_SHORT as string, locale = defaultLocale) =>
   DateTime.fromFormat(string, format, { locale }).toJSDate();
 
+export const parseMultiFormatsDate = (value: string, formats: string[], locale: string) => {
+  let output = null;
+  formats.some((format) => {
```
lorgan3, line 92: Because you you need access to the parsed values I don't think `some` (or any other higher order function) works very well here, I agree with Lowie, a simple for loop with early return would actually make most sense here.

lorgan3 (dismissed):

src/components/datepicker/localeUtils.ts
qubis741: `some` has early return functionality, once `true` is returned and I am setting `output` variable inside that loop, (which I would also do in `for` loop), so what is advantage of using `for` instead of `some` here?

lowiebenoot: I'm ok with using `some`, but I also agree with @lorgan3. 
You're not really using the return value of `some`,  but you're just using `return true` to stop the loop, and set a variable that is scoped outside of the loop.

Something like this makes more sense to me:

```ts
for (let i = 0; i < formats.length; i++) {
  const parsedDate = DateTime.fromFormat(value, formats[i], { locale });
  if (parsedDate.isValid) {
    return parsedDate.toJSDate();
  }
}

return null; // or just no return, if returning undefined is also fine
```

lowiebenoot (approved):

lowiebenoot (approved):

stefaandevylder (approved):

lorgan3 (approved):

lowiebenoot (changes requested):
Found a bug: if you just type something incorrectly (e.g. `foobar`), and then select a date with your mouse, then it picks the date, but it's still showing the error border.


https://user-images.githubusercontent.com/330765/200737903-2374eea5-29e8-41fe-83f0-5531b9ec3a66.mp4

lorgan3 (dismissed):

KristofColpaert (approved):

eniskraasniqi1 (approved):

jelledc (approved):

jelledc:
Seems to work as intended. Small remark(s):

image

There's a small bit of whitespace underneath the calendar, which is caused by a `display: inline-block` on the `date-picker` class. Seems like something that has been in the datepicker code for 4 years now, so maybe we should just set it to flex if we're rendering it in a popover?

Also, the class attribute starts with `undefined`, but that shouldn't cause any issues.

qubis741:
@jelledc thanks for review. `undefined` is from `classnames` function, where we are passing `className` prop that can be also undefined

lowiebenoot (approved):

lorgan3 (approved):

PR #2441, opened 2022-11-14
Fix DatePickerInput inability to click on MonthPicker in Dialog

qubis741:
### Description

Fixed `DatePickerInput` inability to click on `MonthPicker` while inside `Dialog`. Also improved typing for `DatePickerInput`, so `formatDate` is not allowed when `typeable=true` as it's in conflict.

### Manual check

We can do it over my shoulder, as its part of adaptation PR

- link UI library to core
- [x] you have TS errors about `formatDate` not allowed in `DatePickerInput` 
- fix those errors in `client/src/application/Companies/Detail/GenerateReport/CompanyGenerateReportDialog.tsx`
- [ ] new company detail page -> kebab menu -> Download report -> Datepicker are typeable, Month can be selected
- [ ] other selects still work


src/components/select/Select.tsx
```diff
@@ -466,6 +467,7 @@ function Select<Option extends OptionType, IsMulti extends boolean, IsClearable
           ClearIndicator,
           DropdownIndicator,
           IndicatorSeparator: null,
+          Input: (inputProps) => <ReactSelectComponents.Input data-is-select="true" {...inputProps} />,
```
lorgan3, line 470: B?: Components should be defined outside of the render function, this will cause the input to rerender more than neccesary.
https://react-select.com/components#defining-components

lorgan3 (dismissed):

JorenSaeyTL (dismissed):

JorenSaeyTL:
Q: Do the type issues make it a breaking change?

JorenSaeyTL:
MCing ⌛ ...

qubis741:
Adding on HOLD, I need to also validate whether date is allowed

qubis741:
> Q: Do the type issues make it a breaking change?

Actually you are right, as `typeable` is true by default and it would turn "off" custom formatting. Thx

JorenSaeyTL:
Will MC with Stefaan, cause can't seem to link it properly 🙈

lorgan3:
> > Q: Do the type issues make it a breaking change?
> 
> Actually you are right, as `typeable` is true by default and it would turn "off" custom formatting. Thx

Since we barely mentioned typescript changes before I wouldn't consider this a 'real' breaking change though

lorgan3:
I see, yeah since the thing it breaks is just cosmetic I'm inclined to not consider this a breaking change but I can definitely see arguments for making it one as well. I also understand why we're making `typeable` true by default.

jelledc (approved):

jest.config.json
```diff
@@ -0,0 +1,26 @@
...
+  "rootDir": "./",
+  "testMatch": [
+    "**/src/**/?(*.)(spec|test).(ts)?(x)"
+  ],
+  "testPathIgnorePatterns": [
+    "<rootDir>/src($|/.*)(.stories.spec.tsx)$"
+  ],
+  "modulePaths": ["<rootDir>/node_modules"],
+  "coveragePathIgnorePatterns": ["<rootDir>/node_modules/","<rootDir>/es/","<rootDir>/cjs/"],
+  "moduleFileExtensions": ["js", "jsx", "ts", "tsx", "json"],
```
lorgan3, line 11: `js` and `jsx` can be removed right?

jest.config.json
```diff
@@ -0,0 +1,26 @@
...
+  "moduleNameMapper": {
+    "\\.(css|scss)$": "identity-obj-proxy"
+  },
+  "globals": {
+    "__DEBUG__": false,
+    "__DEV__": false,
+    "environment": ""
```
lorgan3, line 18: Are these being set or is this because it's copied from somewhere?

src/components/datepicker/utils.spec.ts
```diff
@@ -0,0 +1,29 @@
+import { isAllowedDate } from './utils';
```
lorgan3, line 1: PP put tests in a `__test(s)__` subfolder

src/components/datepicker/utils.spec.ts
```diff
@@ -0,0 +1,29 @@
...
+      { disabledDays: { before: new Date(2022, 10, 16) }, output: false },
+      { disabledDays: { after: new Date(2022, 10, 16) }, output: true },
+      { disabledDays: { after: new Date(2022, 10, 14) }, output: false },
+      { disabledDays: { before: new Date(2022, 10, 14), after: new Date(2022, 10, 16) }, output: true },
+      { disabledDays: { before: new Date(2022, 10, 16), after: new Date(2022, 10, 17) }, output: false },
+      { disabledDays: { daysOfWeek: [1, 3, 4] }, output: true },
+      { disabledDays: { daysOfWeek: [1, 2, 3, 4] }, output: false },
+      { disabledDays: (_date: Date) => false, output: true },
+      { disabledDays: (_date: Date) => true, output: false },
+    ])('works with disabled days: `$disabledDays', ({ disabledDays, output }) => {
```
lorgan3, line 25: I think you're missing a backtick after `$disabledDays `

src/components/datepicker/utils.ts
```diff
@@ -18,3 +19,53 @@ export const isSelectingFirstDay = (from: Date, to: Date, day: Date) => {
   const isRangeSelected = from && to;
   return !from || isBeforeFirstDay || isRangeSelected;
 };
+
+export const isAllowedDate = (date: Date, disabledDays?: AllowedDisabledDays) => {
+  if (disabledDays) {
```
lorgan3, line 24: PP early return
```
if (!disabledDays) {
    return true;
}

// other checks here
```

src/components/datepicker/utils.ts
```diff
@@ -18,3 +19,53 @@ export const isSelectingFirstDay = (from: Date, to: Date, day: Date) => {
...
+    date.setHours(0, 0, 0, 0);
```
lorgan3, line 25: Looks like a utility function like
```
const getDay = (date: DateTime) => date.setHours(0, 0, 0, 0);
```

Could simplify this code a bunch (it also already returns a timestamp so no need to use `getTime` either)

\* Naming could be better 😅

ci/pipeline.yaml
```diff
@@ -260,6 +260,48 @@ jobs:
         status: failure
         context: check-types
 
+  - name: pr-test
```
lorgan3, line 263: Maybe check with Yorick if we should do this with a github action instead

qubis741: yes I was about to when I am finished with preparations,thx

jest.config.json
qubis741: coppied from `core` it was still WIP but thanks

jest.config.json
qubis741: yes

qubis741:
MC over shoulder together with @eniskraasniqi1 @JorenSaeyTL

qubis741:
Adding `jest` will be tackled in separate PR

src/components/datepicker/utils.ts
```diff
@@ -32,9 +32,9 @@ export const isAllowedDate = (date: Date, disabledDays?: AllowedDisabledDays) =>
   if (disabledDays instanceof Date) {
     return dateTime !== getDateTimeAtStartOfDay(disabledDays);
   }
-  if (Array.isArray(disabledDays) && disabledDays.length > 0) {
+  if (Array.isArray(disabledDays) && disabledDays.length > 0 && disabledDays[0] instanceof Date) {
     return !disabledDays.some((disabledDay) => {
-      return dateTime === getDateTimeAtStartOfDay(disabledDay);
+      return disabledDay instanceof Date ? dateTime === getDateTimeAtStartOfDay(disabledDay) : false;
```
lorgan3, line 37: I guess it doesn't make sense to update a disabled test, but can you add a comment to it to remind us to update it for this case?

lorgan3 (approved):

src/components/datepicker/utils.ts
qubis741: done :)

lorgan3 (approved):

eniskraasniqi1 (approved):

JorenSaeyTL (approved):

PR #2267, opened 2022-07-06
[FRAF-752] - Fix menu not working correctly

eniskraasniqi1:
### Description

- Menu is not working correctly after being converted to typescript, just opening a blank square. Fixed that, now it should work as excpected.

## Manual check

1. Link locally with core branch **ui-library-type-updates**
2. Test the 3 dots menu if its working correctly

PR #2474, opened 2022-12-05
[FIX] Looping of menu item

stefaandevylder:
### Description

Fixes the possibility to have an infinite loop. This was introduced when converting from JS to Typescript. Bug was found with Sentry.

This is a shot in the dark, but most likely the fix.

Sentry: https://sentry.io/organizations/teamleader/issues/3788415931/?project=1211390&referrer=slack

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,7 +12,13 @@
 
 ### Dependency updates
 
-## [18.1.0] - 2022-12-05
+## [18.1.1] - 2022-12-06
+
+### Fixed
+
+- `Menu`: Fix possibility to have an infinite loop ([@stefaandevylder](http://github.com/stefaandevylder)) in [#2474](https://github.com/teamleadercrm/ui/pull/2474)
+
+## [18.1.0] - 2022-12-0
 
 ### Added
 
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "name": "@teamleader/ui",
   "description": "Teamleader UI library",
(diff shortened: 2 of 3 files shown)


lorgan3 (commented):
Q: Do we need to to anything to make sure to not undo this fix? https://github.com/teamleadercrm/ui/pull/2267

@eniskraasniqi1 maybe you remember when the menu opened as an empty square?

qubis741 (dismissed):

KristofColpaert (approved):

eniskraasniqi1:
> 

Not really 😅, but I'm going to test those changes locally (bring back some lines), and see if I added this line for a good reason.

jelledc (dismissed):

lowiebenoot (commented):
Doesn't seem like a good solution to me. The dependency that you now use is a ref, so it will never update, so the effect will never run again.

stefaandevylder:
@lowiebenoot it actually does rerender, tested multiple times

stefaandevylder:
@lowiebenoot The item should not be rerendered so it is correct

qubis741 (approved):

BeirlaenAaron (approved):
Sorry didn't see this, could've probably been added in the same release :see_no_evil:

lowiebenoot (approved):

eniskraasniqi1 (approved):

JorenSaeyTL:
Group MCed with @eniskraasniqi1 and @stefaandevylder ✅

PR #2479, opened 2022-12-06
🐛 Fix looping on Menu

stefaandevylder:
### Description

Previous PR brought a new bug where some pages could not open it. This has been fixed. I removed state as it was not needed.

### DIFF
--- a/src/components/menu/Menu.tsx
+++ b/src/components/menu/Menu.tsx
@@ -1,15 +1,6 @@
 import uiUtilities from '@teamleader/ui-utilities';
 import cx from 'classnames';
-import React, {
-  ReactElement,
-  ReactNode,
-  SyntheticEvent,
-  useCallback,
-  useEffect,
-  useLayoutEffect,
-  useRef,
-  useState,
-} from 'react';
+import React, { ReactElement, ReactNode, SyntheticEvent, useCallback, useEffect, useRef, useState } from 'react';
 import { GenericComponent } from '../../@types/types';
 import Box, { pickBoxProps } from '../box';
 import { BoxProps } from '../box/Box';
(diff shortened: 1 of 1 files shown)


qubis741 (approved):

BeirlaenAaron (approved):

lowiebenoot (changes requested):
It's bugged now. It will render the first time with an undefined width and height. If it doesn't re-render, width and height will stay undefined.

For example: in the Menu story, the outline is not visible because of this issue. It's rendering it with a width and height of 0.

Screenshot 2022-12-06 at 22 38 01

Doing a `getBoundingClientRect()` on each render is not a good idea because
* it's measuring the result of the previous render, which might not be correct
* it's somewhat of an "expensive" execution to do during a render

stefaandevylder:
I feel like this whole thing is just built wrong

stefaandevylder:
Also, how come this is not bugged when I link this with core locally? 🤔🤔

gdegirmenci:
The issue is not specifically related to company/company detail page, happening in other places as well, also it seems first seen was 10 months ago. 

We created ticket to debug in detailed FYI @joaoinacio 

https://teamleader.atlassian.net/browse/ATL-1822

lowiebenoot:
Do I understand it correctly now that previous fix (#2474) actually introduced a different bug and actually broke the component? Cause in that case our release would be blocked.

lowiebenoot:
> I feel like this whole thing is just built wrong

can you also elaborate why @stefaandevylder ?

stefaandevylder:
@lowiebenoot We shouldn't keep state managed like this together with refs in a useEffect

lowiebenoot:
> 



> @lowiebenoot We shouldn't keep state managed like this together with refs in a useEffect

@stefaandevylder that's still very vague. Elaborate more please.

stefaandevylder:
@lowiebenoot I just mean that using a ref inside a useEffect is not best practice

PR #2483, opened 2022-12-08
Add `tooltipProps` to `Avatar` to allow a custom tooltip

BeirlaenAaron:
### Description

We would like to add a custom tooltip to the avatars, which is "Unassigned" in our case
Added a prop `tooltipProps` to Avatar

image

### DIFF
--- a/src/components/avatar/Avatar.tsx
+++ b/src/components/avatar/Avatar.tsx
@@ -48,6 +48,8 @@ export interface AvatarProps extends Omit<BoxProps, 'size' | 'ref'> {
   team?: boolean;
   /** If true, the name will be shown in a tooltip on hover. */
   tooltip?: boolean;
+  /** The tooltip props for the avatar */
+  tooltipProps?: TooltipProps;
 }
 
 type AvatarInternalComponentProps = { size: Exclude<typeof SIZES[number], 'fullscreen' | 'smallest'> } & Pick<
@@ -138,6 +140,7 @@ const Avatar = ({
   id,
   onImageChange,
   team,
+  tooltipProps,
   ...others
 }: AvatarProps) => {
   const avatarClassNames = cx(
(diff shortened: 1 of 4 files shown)


KristofColpaert (approved):

src/components/avatar/Avatar.tsx
```diff
@@ -151,10 +154,10 @@ const Avatar = ({
...
 
-  const enableTooltip = tooltip && typeof fullName === 'string' && fullName.length > 0;
+  const enableTooltip = tooltip && ((typeof fullName === 'string' && fullName.length > 0) || tooltipProps);
 
   const Component = enableTooltip ? TooltippedBox : Box;
-  const tooltipProps = enableTooltip
+  const defaultToolTipProps = enableTooltip
     ? {
         tooltip: <TextBodyCompact>{fullName}</TextBodyCompact>,
         tooltipColor: 'white',
```
lorgan3, line 163: Maybe add these to the tooltipProps too so it's easier to make it look consistent
```
{
  tooltipColor: 'white',
  tooltipPosition: 'top',
  tooltipSize: 'small',
  ...tooltipProps
}
```

lorgan3 (approved):

lowiebenoot (dismissed):

src/components/avatar/Avatar.tsx
```diff
@@ -151,23 +154,24 @@ const Avatar = ({
     className,
   );
 
-  const enableTooltip = tooltip && typeof fullName === 'string' && fullName.length > 0;
+  const enableTooltip = tooltip && ((typeof fullName === 'string' && fullName.length > 0) || tooltipProps);
```
lorgan3, line 157: Nice!, I didn't even think about adding the props also in the original case but it makes a lot of sense.
I have 1 tiny suggestion for a further improvement now though 😅 
This way when overwriting the default styles it won't show an empty tooltip if there's no fullName

```suggestion
  const enableTooltip = tooltip && ((typeof fullName === 'string' && fullName.length > 0) || tooltipProps?.tooltip);
```

lorgan3 (approved):

lorgan3 (approved):
✅  Also manually checked

farazatarodi (approved):

qubis741 (approved):

PR #2488, opened 2022-12-14
[FRONT-95] Unit testing

lowiebenoot:
### Description

Implemented unit testing with testing-library and jest.
I think it would be nice to add a very basic snapshot test for each component, just so that we know it still renders.
And adding more and more tests gradually based on specific props and interactions for those components.
I still need to add it to CI as well, will do that as a follow up.

Currently I added a few snapshot tests and some more specific tests for `Button` and `Dialog` already.

What do you think?


lowiebenoot:
(there's a TS error which I fixed in a separate PR. I thought changing the base branch would re-run CI but apparently not)

jest.config.js
```diff
@@ -0,0 +1,12 @@
+module.exports = {
+  setupFilesAfterEnv: ['./tests/setupTests.ts'],
+  moduleNameMapper: {
+    '\\.(jpg|jpeg|png|gif|eot|otf|webp|svg|ttf|woff|woff2|mp4|webm|wav|mp3|m4a|aac|oga)$':
+      '<rootDir>/tests/__mocks__/fileMock.js',
+    '\\.(css|less|scss|sass)$': 'identity-obj-proxy',
+    '^@teamleader/ui-utilities': '<rootDir>/tests/__mocks__/filemock.js',
+    '^@teamleader/ui-typography': '<rootDir>/tests/__mocks__/filemock.js',
```
lorgan3, line 8: We can also make this a ts file right?

src/components/button/__tests__/Button.spec.tsx
```diff
@@ -0,0 +1,75 @@
+import React from 'react';
+import { render, fireEvent } from '@testing-library/react';
+import Button from '../Button';
+
+describe('Component - Button', () => {
+  test('it renders', async () => {
```
lorgan3, line 6: pp use `it` instead of `test` 
```suggestion
  it('renders', async () => {
```

src/components/dialog/__tests__/Dialog.spec.tsx
```diff
@@ -26,4 +27,109 @@ describe('Component - Dialog', () => {
...
+      >
+        Foobar
+      </Dialog>,
+    );
+    expect(screen.baseElement).toMatchSnapshot();
+  });
+
+  test('it calls the onClose handler when clicking the close button', async () => {
+    const handleOnClose = jest.fn();
+    const user = userEvent.setup();
```
lorgan3, line 48: We should probably use this for all tests instead of fireEvent then

lorgan3 (dismissed):

eniskraasniqi1 (approved):

stefaandevylder (approved):

BeirlaenAaron (approved):

jelledc (approved):

JorenSaeyTL (approved):

farazatarodi (approved):

PR #2489, opened 2022-12-14
`Dialog`: add the event parameter to the `Dialog` `onSubmit` type

lowiebenoot:
### Description

The `onSubmit` handler can have an `event` parameter.


lorgan3 (approved):

JorenSaeyTL (approved):

JorenSaeyTL:
I don't think the `frontenders` group does what is expected, cause I think I now approved for all frontenders

lowiebenoot:
@JorenSaeyTL yeah, I thought it was updated for all teams, but apparently we only updated the setting for 1 team.

PR #2493, opened 2022-12-16
Add GitHub action for running unit tests

lowiebenoot:
### Description

Added a GitHub action for running unit tests.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -1,0 +1,12 @@
+name: 'Test'
+on: push
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - name: Install dependencies
+        run: yarn
+      - name: Run tests
+        run: yarn test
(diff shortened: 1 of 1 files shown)


.github/workflows/test.yml
```diff
@@ -0,0 +1,12 @@
+name: 'Test'
+on: push
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
```
YorickH, line 6:
```suggestion
    runs-on: self-hosted
```

.github/workflows/test.yml
```diff
@@ -0,0 +1,12 @@
...
+    steps:
+      - uses: actions/checkout@v2
+      - name: Install dependencies
+        run: yarn
```
YorickH, line 10: Add caching for `yarn`

.github/workflows/test.yml
lowiebenoot: Why actually?

YorickH: You can leave it, these small ones are not a real issue

tiagobastos (approved):

YorickH (approved):

PR #2506, opened 2023-01-02
[FRONT-354] Fix GridItem

farazatarodi:
### Description

- Apparently `GridItem` was not exported. I made it so it will be exported through the `Grid` component itself.
- Now you can pass class names to the `Flex` and `Grid` components.

```
<GridItem area="header" />
```

#### Screenshot after this PR

```
<Grid.Item area="header" />
```

### Breaking changes

NA

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,9 @@
 
 ### Added
 
+- `Flex`: Allow class names
+- `Grid`: Allow class names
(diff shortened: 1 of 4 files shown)


mrtus (approved):
Yay! Thank you @farazatarodi

lorgan3 (approved):

PR #2511, opened 2023-01-03
`AdvancedCollapsible` - add `defaultIsCollapsed` prop

BeirlaenAaron:
### Description

Add `defaultIsCollapsed` prop to be able to set it's default state

### DIFF
--- a/src/components/advancedCollapsible/AdvancedCollapsible.tsx
+++ b/src/components/advancedCollapsible/AdvancedCollapsible.tsx
@@ -18,6 +18,7 @@ export interface AdvancedCollapsibleProps extends Omit<BoxProps, 'size'> {
   children: ReactNode;
   title: string;
   size?: AllowedAdvancedCollapsibleSize;
+  isCollapsed?: boolean;
 }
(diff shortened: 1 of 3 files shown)


driesd (approved):

src/components/advancedCollapsible/AdvancedCollapsible.tsx
```diff
@@ -18,16 +18,18 @@ export interface AdvancedCollapsibleProps extends Omit<BoxProps, 'size'> {
   children: ReactNode;
   title: string;
   size?: AllowedAdvancedCollapsibleSize;
+  isCollapsed?: boolean;
```
lorgan3, line 21: NTH: Call this `defaultIsCollapsed` to indicate that changing this prop after the initial render won't change the state anymore

(or alternatively make it so that changing this prop also changes the state)

stefaandevylder (approved):

lowiebenoot (approved):

lorgan3 (approved):

PR #2544, opened 2023-01-27
`MarketingStatusLabel` - make `Icon` optional and configurable

KristofColpaert:
### Description

`MarketingStatusLabel`:
- It's also possible to use this component without an `Icon`
- You can configure the `Icon`. It doesn't default to `IconLockSmallFilled` anymore

Screenshot 2023-01-27 at 10 18 54

#### Screenshot after this PR

Screenshot 2023-01-27 at 10 19 38
Screenshot 2023-01-27 at 10 19 43

### Breaking changes

- The default `Icon (IconLockSmallFilled)` in `MarketingStatusLabel` became optional. Now the default hasn't got an `Icon`. According to GitHub, there aren't any usages of the `MarketingStatusLabel` (except for `payments-frontend`, but this is the one who would like to modify).


package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "@teamleader/ui",
   "description": "Teamleader UI library",
-  "version": "18.6.2",
+  "version": "18.6.3",
```
qubis741, line 4: if its breaking change then you should go with major update

KristofColpaert: Knocks like a bus! 🚌 

JK, you're right! I will update this ;)

src/components/marketingStatusLabel/MarketingStatusLabel.tsx
```diff
@@ -31,18 +32,16 @@ const MarketingStatusLabel: GenericComponent<MarketingStatusLabelProps> = ({
...
       className={classNames}
       paddingHorizontal={2}
     >
       <TextElement className={theme['text']} marginRight={2}>
         {children}
       </TextElement>
-      <Icon className={theme['icon']}>
-        <IconLockSmallFilled />
-      </Icon>
+      <Icon className={theme['icon']}>{icon}</Icon>
```
jelledc, line 44:
```suggestion
      {icon && <Icon className={theme['icon']}>{icon}</Icon>}
```

KristofColpaert: Done 😉

jelledc (approved):

jelledc:
MC done 🚢

BeirlaenAaron (approved):

lorgan3 (approved):
Would you ever want to show another icon than the lock one?
I can't really think of a use case for that and if we were to add a prop instead to hide the lock we could make this a non breaking change 🤔

PR #2575, opened 2023-02-22
[FRONT-123] remove automatic blurring

lowiebenoot:
### Description

Some of our components contain automatic blurring when clicking them, to avoid having a focus state after clicking. This is not necessary anymore these days, because we now just style the focus state via `:focus-visible` which is only used when focus state is reached via keyboard (kind of, in reality it's a bit more complicated than that).

The automatic blurring was also causing some issues because the focus completely resets when doing the blur.

---

### Changed

- Focus states are now styled through `:focus-visible` instead of `:focus`. ([@lowiebenoot](https://github.com/lowiebenoot)) in [#2575](https://github.com/teamleadercrm/ui/pull/2575)

### Fixed

- `Toggle`: the component is now focusable. ([@lowiebenoot](https://github.com/lowiebenoot)) in [#2575](https://github.com/teamleadercrm/ui/pull/2575)

### Dependency updates

- `postcss-preset-env`: bump peer dependency expectation to `^8.0.0`, otherwise the `:focus-visible` styling will not be included. ([@lowiebenoot](https://github.com/lowiebenoot)) in [#2575](https://github.com/teamleadercrm/ui/pull/2575)

---

### Manual check

Check focus states in storybook. Clicking on them shouldn't show a focus state, but navigating with keyboard 

- [ ] in storybook
- [ ] linking it into your local project (I tested on `service-auth-frontend`. Couldn't get the linking working on `core`, even after linking react)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,9 +16,21 @@
 
 ## [20.0.0] - 2023-02-13
 
+### Changed
+
+- Focus states are now styled through `:focus-visible` instead of `:focus`. ([@lowiebenoot](https://github.com/lowiebenoot)) in [#2575](https://github.com/teamleadercrm/ui/pull/2575)
+
+### Fixed
+
+- `Toggle`: the component is now focusable. ([@lowiebenoot](https://github.com/lowiebenoot)) in [#2575](https://github.com/teamleadercrm/ui/pull/2575)
+
 ### Removed
 
 - `Label`: Remove option for 'tiny' label size ([@JorenSaeyTL](https://github.com/JorenSaeyTL)) in ([#2560](https://github.com/teamleadercrm/ui/pull/2560))
+
+### Dependency updates
+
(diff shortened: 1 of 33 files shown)


driesd (dismissed):

farazatarodi (dismissed):

lowiebenoot:
I somehow fucked something up with the link components while doing incremental commits, but it was caught by chromatic 💪

driesd (approved):

jelledc (approved):

farazatarodi (approved):

KristofColpaert (approved):

stefaandevylder (approved):

lorgan3 (approved):
Did you make this a breaking change because of the dependency update?

farazatarodi:
@lorgan3 No it is a breaking change because of 'tiny' and 'Section'

farazatarodi:
🕵️‍♂️

qubis741 (approved):

lowiebenoot:
@lorgan3 yeah, I just added my changes to a version that was not released yet.

Already adding a version to the changelog without making a release is really confusing me. I think we should stop doing it and do it the way we did before.

(And soon I want to look into making automatic changelogs where you just have to add a file with your changes, and later we can make a changelog automatically based on those files. There's a tool for that but I can't find it anymore, so if someone knows it, let me know 😄).

eniskraasniqi1 (approved):

PR #2582, opened 2023-02-28
[FRONT-443] Avoid DataGrid error when hovering

lowiebenoot:
### Description

DataGrid can crash under certain circumstances and then hovering over a DataGrid row. It seems to be a weird occasion when the amount of rows changes, but I can't reproduce it inside storybook (but I'm consistently having the issue in workorders-frontend when updated to v20). 

The issue is caused by storing a ref as null, but in a calculation we don't expect it to be null. The ref is stored as null, because it's an inline function, which means it gets called with null, and then with the DOM element.

From the [react docs](https://reactjs.org/docs/refs-and-the-dom.html#callback-refs): 

> If the ref callback is defined as an inline function, it will get called twice during updates, first with null and then again with the DOM element. This is because a new instance of the function is created with each render, so React needs to clear the old ref and set up the new one.

Screenshot 2023-02-28 at 14 18 51

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,12 @@
 
 ### Dependency updates
 
+## [20.0.1] - 2023-02-28
+
+### Changed
+
+- `DataGrid` Avoid error when hovering over a row. ([@lowiebenoot](https://github.com/lowiebenoot)) in [#2582](https://github.com/teamleadercrm/ui/pull/2582)
+
 ## [20.0.0] - 2023-02-27
 
 ### Changed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,7 +14,7 @@
 
 ## [20.0.1] - 2023-02-28
 
-### Changed
+### Fixed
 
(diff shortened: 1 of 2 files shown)


eniskraasniqi1 (approved):

CHANGELOG.md
```diff
@@ -12,6 +12,12 @@
 
 ### Dependency updates
 
+## [20.0.1] - 2023-02-28
+
+### Changed
+
+- `DataGrid` Avoid error when hovering over a row. ([@lowiebenoot](https://github.com/lowiebenoot)) in [#2582](https://github.com/teamleadercrm/ui/pull/2582)
```
lorgan3, line 19: `fixed` ? 😅

lorgan3 (approved):

CHANGELOG.md
lowiebenoot: ugh 😅

JorenSaeyTL (approved):

jelledc (approved):

PR #2618, opened 2023-04-06
[FRONT-476] make Menu scrollable

farazatarodi:
### Description

I meant to make `Menu` scrollable, but I basically rewrote it.

```
### Added

- `Menu`: It is now scrollable when the available space is less than the height

### Changed

- [BREAKING] `Menu`: State management now should happen in the parent component.
- `Menu`: Shadow and border now use the values from the design system.
- [BREAKING] `Menu`: The `onHide` prop is now called `onInactive`.

### Removed

- [BREAKING] `Menu`: The `onShow` property is removed as the state management is now moved to the parent component.

### Fixed

- [BREAKING] `Menu`: It now requires an anchor element for positioning when it is not static. Previously it was based on the parent element, which caused layout bugs in `flex` elements.
```

image
image

#### Screenshot after this PR

image
image

### Breaking changes

- `Menu`: State management now should happen in the parent component.
- `Menu`: The `onHide` prop is now called `onInactive`.
- `Menu`: The `onShow` property is removed as the state management is now moved to the parent component.
- `Menu`: It now requires an anchor element for positioning when it is not static. Previously it was based on the parent element, which caused layout bugs in `flex` elements.


lowiebenoot (commented):
Should there be a recalculation of max height (and repositioning when auto position?) on window resize?

When trying the Menu and IconMenu component out in story book, i get some weird things when changing the position, but I'm not sure if it's an issue with the component or with the storybook container that is not ideal to show off the component.

When I change the position to bottom left on the Menu component:
Screenshot 2023-04-07 at 11 56 12

IconMenu component being clipped off

Screenshot 2023-04-07 at 11 26 57

src/components/menu/Menu.tsx
```diff
@@ -51,6 +42,8 @@ export interface MenuProps<S = any> extends Omit<BoxProps, 'children' | 'classNa
   selected?: S;
 }
 
+type Position = Exclude<MenuProps['position'], undefined>;
```
lorgan3, line 53: This does not look right. (Could just be `type Position = MenuProps['position']` I guess)

PP:
```
type Position = 'auto' | 'static' | 'top-left' | 'top-right' | 'bottom-left' | 'bottom-right';

export interface MenuProps<S = any> extends Omit<BoxProps, 'children' | 'className'> {
  ...
  position?: Position
  ...
}
```

src/components/menu/Menu.tsx
```diff
@@ -64,17 +57,15 @@ const Menu = <S,>({
   selected,
   ...others
 }: MenuProps<S>): ReactElement<any, any> | null => {
-  const [stateWidth, setStateWidth] = useState<number | undefined>(0);
-  const [stateHeight, setStateHeight] = useState<number | undefined>(0);
-  const [statePosition, setPosition] = useState<string | undefined>(position);
+  const [positionState, setPositionState] = useState<Position>(position);
```
lorgan3, line 68: PP: changing `position` after the component was mounted won't do anything. Either name it something like `initialPosition` to clarify or add a `useEffect` to update the state when the prop changes.

src/components/menu/Menu.tsx
farazatarodi: @LennertBlommaert the type is correct, but your suggestion makes more sense.

JorenSaeyTL (approved):

farazatarodi:
@lowiebenoot I don't see it necessary, but in the case that we want to add it, I would add it in another PR. This one is huge enough.

lorgan3 (dismissed):

lowiebenoot:
> @lowiebenoot I don't see it necessary, but in the case that we want to add it, I would add it in another PR. This one is huge enough.

I thought we had it in the current Menu component, but apparently not, so nevermind. It's probably in popover that we have it.

lowiebenoot (dismissed):

CHANGELOG.md
```diff
@@ -2,16 +2,25 @@
...
 
+- `Menu`: It is now scrollable when the available space is less than the height ([@farazatarodi](https://https://github.com/farazatarodi) in [#2618](https://github.com/teamleadercrm/ui/pull/2618))
+
 ### Changed
 
-- `Container`: remove 72px on xl viewport ([@farazatarodi](https://https://github.com/farazatarodi) in [#2615](https://github.com/teamleadercrm/ui/pull/2615))
+- `Container`: Removed 72px padding on xl viewport ([@farazatarodi](https://https://github.com/farazatarodi) in [#2618](https://github.com/teamleadercrm/ui/pull/2618))
+- [BREAKING] `Menu`: State management now should happen in the parent component ([@farazatarodi](https://https://github.com/farazatarodi) in [#2618](https://github.com/teamleadercrm/ui/pull/2618)).
+- `Menu`: Shadow and border now use the values from the design system ([@farazatarodi](https://https://github.com/farazatarodi) in [#2618](https://github.com/teamleadercrm/ui/pull/2618)).
+- [BREAKING] `Menu`: The `onHide` prop is now called `onInactive` ([@farazatarodi](https://https://github.com/farazatarodi) in [#2618](https://github.com/teamleadercrm/ui/pull/2618)).
```
lowiebenoot, line 12: What's the reason for renaming this actually?

lowiebenoot: I find onHide a better name.

farazatarodi: @lowiebenoot because I initially replaced it with `setActive` and then again with `onInactive`. Sounded like a more suitable name.

src/components/menu/Menu.tsx
```diff
@@ -204,6 +204,19 @@ const Menu = <S,>({
     }
   }, [active, handleDocumentClick, position]);
 
+  useEffect(() => {
+    const menuNode = menuRef.current;
+    if (!menuNode) {
+      return;
+    }
+
+    menuNode.addEventListener('focusout', handleDocumentClick);
```
lorgan3, line 213: Q: should this also only happen for static position like the `click` event handler?

lorgan3 (dismissed):

src/components/menu/Menu.tsx
lorgan3: (And do you need both this one and the click event handler?)
I suppose this will close the menu if you click on a disabled menu item too which might not be desired

farazatarodi: @lorgan3 I removed the event listener. The problem before was not the event listener on the document, it was the callback.

lowiebenoot (approved):

PR #2620, opened 2023-04-13
[FRONT-481] Fix multiple variable imports

lowiebenoot:
### Description

We noticed browser inspectors were becoming very slow. This was caused by the many imports of css variables (aka custom properties). In (almost) every css file we were importing the variables (colors, utilities, typography, animations), which was causing the variables being output in the css multiple times. 

Since `@teamleader/ui-colors@1.0.0` the colors were split in multiple variables, and has 5 variables for each colors, which seems to have been slowing down the inspector a lot, in combination with importing it so many times.

To fix this, I removed all variable imports in css files and disabled the `custom-properties` in the postcss config, so we will use custom properties "natively" instead of having them transpiled during the build process.

Because the variables are not imported in the css files anymore, you need to import them yourself. You can do that by importing them into your JS or CSS.

JS:
```js
import '@teamleader/ui/es/index.css';
```

CSS:
```css
@import url('@teamleader/ui/es/index.css');
```

I also removed/changed/refactored some css to make sure we don't have duplicate variable names for different components.

Using custom properties natively means that we can use them in a different way than we are used to. You can for example do something like this:

```css
.button {
  background: var(--button-color);
}

.primary  {
  --button-color: green;
}

.secondary  {
  --button-color: grey;
}
```

### Manual check

- run story book
- inspect a component
=> you should see almost no duplicate variables in the css inspector anymore. `ui-typography` and `ui-utilities` still seems to be duplicate, because they are both imported in js as well, which seems to add the variables to the css as well.

### Breaking changes

- [BREAKING]: CSS custom properties are not transpiled anymore. Before, the custom properties were replaced with the actual values in the built CSS, but now they are not replaced as we want to actually use custom properties natively. So you need to make sure that you import the variables into your project (e.g. `import '@teamleader/ui/es/index.css';` in JS or `@import url('@teamleader/ui/es/index.css)` in css). ([@lowiebenoot](https://https://github.com/lowiebenoot) in [#2620](https://github.com/teamleadercrm/ui/pull/2620))
- [BREAKING] `@teamleader/ui-colors` has been updated to version 2.0.0 which uses the newer comma-free syntax for hsl colors. This means that you need to update the syntax where you use the hsl value in combination with alhpa (e.g. `hsl(var(--color-teal), 50%)` should be changed to `hsl(var(--color-teal) /50%)`). ([@lowiebenoot](https://https://github.com/lowiebenoot) in [#2620](https://github.com/teamleadercrm/ui/pull/2620))

### DIFF
--- a/src/components/counter/theme.css
+++ b/src/components/counter/theme.css
@@ -1,9 +1,5 @@
-:root {
-  --border-radius: 4px;
-}
-
 .counter {
-  border-radius: var(--border-radius);
+  border-radius: 4px;
   color: var(--color-white);
   display: inline-flex;
   align-items: center;
--- a/src/components/counter/theme.css
+++ b/src/components/counter/theme.css
@@ -1,7 +1,3 @@
-@import '@teamleader/ui-colors';
-@import '@teamleader/ui-utilities';
-@import '@teamleader/ui-typography';
-
 :root {
   --border-radius: 4px;
 }
--- a/src/components/counter/theme.css
+++ b/src/components/counter/theme.css
@@ -1,5 +1,5 @@
 .counter {
-  border-radius: 4px;
+  border-radius: var(--border-radius-medium);
   color: var(--color-white);
   display: inline-flex;
   align-items: center;
--- a/.storybook/preview.ts
+++ b/.storybook/preview.ts
@@ -1,6 +1,7 @@
 import { addDecorator, addParameters } from '@storybook/react';
 import { withDesign } from 'storybook-addon-designs';
 import { COLOR } from '../src/constants';
+import '../src/index.css';
 
 import theme from './theme';
 import {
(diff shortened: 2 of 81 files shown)


src/components/counter/theme.css
```diff
@@ -1,9 +1,5 @@
-:root {
-  --border-radius: 4px;
-}
-
 .counter {
-  border-radius: var(--border-radius);
+  border-radius: 4px;
```
lorgan3, line 2: This can be `--border-radius-medium` from ui-utilities

lorgan3 (approved):
🎉

lorgan3 (approved):

qubis741 (dismissed):

stefaandevylder (approved):

BeirlaenAaron (approved):

driesd (approved):

farazatarodi:
🕵️‍♂️

farazatarodi (approved):

eniskraasniqi1 (approved):

PR #2642, opened 2023-05-05
Make cursor a pointer on select inputs

BeirlaenAaron:
### Description

Changed cursor into a pointer when hovering over select inputs except for when they're disabled, and a text cursor if they are searchable
Also for the `FilterSelection` component

image

#### Screenshot after this PR

image

### Breaking changes

N/A


lorgan3:
I thought we would do this for the `FilterSelection` component because it's a button but not for select because it's more and input 🤔

BeirlaenAaron:
> I thought we would do this for the `FilterSelection` component because it's a button but not for select because it's more an input

@lorgan3 Select is also a button if it isn't searchable but not sure if it should be a pointer if it's searchable, maybe a text cursor like inputs have?

lorgan3:
> > I thought we would do this for the `FilterSelection` component because it's a button but not for select because it's more an input
> 
> @lorgan3 Select is also a button if it isn't searchable but not sure if it should be a pointer if it's searchable, maybe a text cursor like inputs have?

Best check with design 😉

BeirlaenAaron:
> > > I thought we would do this for the `FilterSelection` component because it's a button but not for select because it's more an input
> > 
> > 
> > @lorgan3 Select is also a button if it isn't searchable but not sure if it should be a pointer if it's searchable, maybe a text cursor like inputs have?
> 
> Best check with design wink

I asked, it should have a text cursor when it is searchable and if it is not a normal pointer

It actually already had a cursor, but it was a bit buggy :stuck_out_tongue: 

[Screencast from 05-05-23 16:52:51.webm](https://user-images.githubusercontent.com/55881661/236501917-f1a64875-df10-4f8b-aa79-19f98cd81bc8.webm)

CHANGELOG.md
```diff
@@ -12,6 +12,13 @@
 
 ### Dependency updates
 
+## [21.0.3] - 2023-05-05
```
lorgan3, line 15:
```suggestion
## [21.1.0] - 2023-05-08
```

I would go for a minor version, patch is only for `fixed` (you could also argue that this is fixed, not changed ofcourse)

lorgan3: Don't forget to update package.json if you were to change this

lorgan3 (dismissed):

lorgan3 (approved):

PR #2656, opened 2023-05-22
[FRONT-515] menu component portal

farazatarodi:
### Description

With the new implementation of the Menu component, it is using absolute position to position it according to the anchor element. This requires that the menu and the anchor element be wrapped inside a container with a non-static position.

This container and the elements can have problems with stack context (z index) and overflow.

After research and prototypes, the best solution would be to render the Menu component with a portal.

Other approaches considered:

- Including the container in the Menu itself. This will limit the component and difficult to implement.

- Using the dialog HTML tag. The top layer element that the dialog is rendered in can not be styled. This causes issues with window size and overflow.

### Screenshot before this PR

image

### Screenshot after this PR

image


jelledc (dismissed):

lorgan3 (approved):

eniskraasniqi1 (dismissed):

driesd (approved):

qubis741 (dismissed):

src/components/menu/Menu.tsx
```diff
@@ -247,6 +247,14 @@ const Menu = <S,>({
     }
   }, [calculatePosition, positionState, maxHeight]);
 
+  useEffect(() => {
+    if (active) {
+      document.body.style.overflow = 'hidden';
+    } else {
+      document.body.style.overflow = '';
```
lorgan3, line 254: I think you have to do something like https://github.com/teamleadercrm/ui/blob/next-release/src/components/overlay/Overlay.tsx#L63-L73 to prevent incorrectly unlocking the scroll

farazatarodi: @lorgan3 Isn't this the same? Do you mean the `lockscreen` prop?

driesd (approved):

lorgan3:
Did you do merge master by accident?

farazatarodi:
Creating a new PR because the base is not `next-release`

PR #2657, opened 2023-05-22
[FIX] react hooks

farazatarodi:
### Description

Adds the eslint plugin for React hook rules.

NA

#### Screenshot after this PR

NA

### Breaking changes

NA

### DIFF
--- a/src/components/datagrid/DataGrid.tsx
+++ b/src/components/datagrid/DataGrid.tsx
@@ -74,7 +74,7 @@ export const DataGrid: DatagridComponent = ({
             .reduce((accumulatedChildWidth, currentChildWidth) => accumulatedChildWidth + currentChildWidth, 0),
         )
         .reduce((maxRowWidth, currentRowWidth) => (currentRowWidth > maxRowWidth ? currentRowWidth : maxRowWidth), 0),
-    [rowNodes, rowNodes.size],
+    [rowNodes],
   );
   const childrenArray: (ReactElement | ReactElement[])[] = !Array.isArray(children) ? [children] : children;
   const bodyRowCount = (childrenArray.find((child) => Array.isArray(child)) as ReactElement[] | undefined)?.length;
@@ -139,6 +139,7 @@ export const DataGrid: DatagridComponent = ({
   useEffect(() => {
     setSelectedRows([]);
     handleSelectionChange([]);
+    // eslint-disable-next-line react-hooks/exhaustive-deps
   }, [comparableId]);
 
   const classNames = cx(
--- a/src/components/datagrid/DataGrid.tsx
+++ b/src/components/datagrid/DataGrid.tsx
@@ -1,6 +1,6 @@
 import cx from 'classnames';
 import omit from 'lodash.omit';
-import React, { ChangeEvent, ReactElement, ReactNode, useEffect, useMemo, useRef, useState } from 'react';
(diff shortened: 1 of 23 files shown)


jelledc (dismissed):

KristofColpaert (dismissed):

lorgan3:
How well is this tested? Changing the dependency arrays is very likely to change the behaviour of those components

farazatarodi:
@lorgan3 tested all of them. But an OTS MC is always welcome. I might have missed something.

JorenSaeyTL (approved):

lowiebenoot (approved):
I think this is quite risky to do though.

farazatarodi:
Reverted the commit that "fixes" the dependencies as it is impossible to guarantee the prevention of bugs.

JorenSaeyTL:
Good job but I think only the dependencies in the `useEffect`s are risky. I guess better safe then sorry 😉

JorenSaeyTL (approved):

src/components/datagrid/DataGrid.tsx
```diff
@@ -74,7 +74,7 @@ export const DataGrid: DatagridComponent = ({
             .reduce((accumulatedChildWidth, currentChildWidth) => accumulatedChildWidth + currentChildWidth, 0),
         )
         .reduce((maxRowWidth, currentRowWidth) => (currentRowWidth > maxRowWidth ? currentRowWidth : maxRowWidth), 0),
-    [rowNodes, rowNodes.size],
+    [rowNodes],
```
lowiebenoot, line 77: I think this useMemo will now run only once, as `rowNodes` never seems to be changed and is always the same Map. That's whay the `rowNodes.size` was added I guess.

src/components/tab/TabGroup.tsx
```diff
@@ -54,7 +54,9 @@ const TabGroup: GenericComponent<TabGroupProps> = ({ children, className, size,
   useEffect(() => {
     smoothScroll.polyfill();
     scrollContainerRef.current?.addEventListener('scroll', handleScroll);
+    // eslint-disable-next-line react-hooks/exhaustive-deps
```
lorgan3, line 57: This one can is probably not required

src/components/tooltip/Tooltip.tsx
```diff
@@ -87,6 +87,7 @@ const TooltippedComponent: GenericComponent<TooltippedComponentProps> = ({
   ComposedComponent,
   ...other
 }) => {
+  // eslint-disable-next-line react-hooks/exhaustive-deps
```
lorgan3, line 90: Same for this one

src/components/wysiwygEditor/WysiwygEditor.tsx
```diff
@@ -127,8 +127,10 @@ const WysiwygEditor: GenericComponent<WysiwygEditorProps> = ({
       if (!onKeyDown || !editorRef?.current?.wrapper) {
         return;
       }
+      // eslint-disable-next-line react-hooks/exhaustive-deps
```
lorgan3, line 130: And this one

src/components/tooltip/Tooltip.tsx
farazatarodi: @lorgan3 Why?

PR #2660, opened 2023-05-23
[FRONT-370] support all Box props on the Flex component

lowiebenoot:
### Description

Added support for all `Box` props on the `Flex` component (except `flexDirection`, which is just `direction` on the `Flex` component). Also allowed the reverse flex directions as well.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,9 @@
 
 ### Changed
 
+- `Flex`: Added support for all `Box` props on the `Flex` component ([@lowiebenoot](https://https://github.com/lowiebenoot) in [#2660](https://github.com/teamleadercrm/ui/pull/2660))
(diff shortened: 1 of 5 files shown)


stefaandevylder (approved):

eniskraasniqi1 (approved):

farazatarodi (approved):

qubis741 (approved):

BeirlaenAaron (approved):

lorgan3 (approved):
Now the only difference between flex and box is the `gap` property so I would kind of like to just add that to box and deprecate flex.

driesd (approved):

driesd:
> Now the only difference between flex and box is the `gap` property so I would kind of like to just add that to box and deprecate flex.

I'm all in favor of this ☝️

lowiebenoot:
> Now the only difference between flex and box is the `gap` property so I would kind of like to just add that to box and deprecate flex.

Box is not flex by default

KristofColpaert (approved):

PR #2661, opened 2023-05-23
[FRONT-515] menu component portal

farazatarodi:
OG PR: https://github.com/teamleadercrm/ui/pull/2656

### Description

With the new implementation of the Menu component, it is using absolute position to position it according to the anchor element. This requires that the menu and the anchor element be wrapped inside a container with a non-static position.
This container and the elements can have problems with stack context (z index) and overflow.
After research and prototypes, the best solution would be to render the Menu component with a portal.

Other approaches considered:
- Including the container in the Menu itself. This will limit the component and difficult to implement.
- Using the dialog HTML tag. The top layer element that the dialog is rendered in can not be styled. This causes issues with window size and overflow.

Due to the use of portals, the Menu component is no longer rendered as a sibling of its anchor element. This means that its position will not change if the page is scrolled. We can either implement a position listener for the anchor element or lock the scroll to solve this. Adding a listener proved difficult to implement and locking the scroll is chosen as the solution. Locking the scroll is only possible with an overlay as we do not know which element is scrollable.

### Screenshot before this PR

image

### Screenshot after this PR

image

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,10 +2,13 @@
 
 ### Added
 
+- `Menu`: If the child is a `Fragment` element, the `onClick` event handler will be passed to its children. ([@farazatarodi](https://https://github.com/farazatarodi) in [#2618](https://github.com/teamleadercrm/ui/pull/2656))
+
 ### Changed
 
 - `Flex`: Added support for all `Box` props on the `Flex` component ([@lowiebenoot](https://https://github.com/lowiebenoot) in [#2660](https://github.com/teamleadercrm/ui/pull/2660))
 - `Flex`: Allow reverse flex directions as well ([@lowiebenoot](https://https://github.com/lowiebenoot) in [#2660](https://github.com/teamleadercrm/ui/pull/2660))
+- `Menu`: Menu is now rendered through a portal if the position is not static ([@farazatarodi](https://https://github.com/farazatarodi) in [#2618](https://github.com/teamleadercrm/ui/pull/2656))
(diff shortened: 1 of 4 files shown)


CHANGELOG.md
```diff
@@ -2,10 +2,13 @@
 
 ### Added
 
+- `Menu`: If the child is a `Fragment` element, the `onClick` event handler will be passed to its children. ([@farazatarodi](https://https://github.com/farazatarodi) in [#2618](https://github.com/teamleadercrm/ui/pull/2656))
```
lorgan3, line 5: Pr references are wrong (also for changed and fixed)

lowiebenoot (dismissed):

stefaandevylder (approved):

lorgan3 (approved):

jelledc (dismissed):

driesd (approved):

KristofColpaert (dismissed):

stefaandevylder (approved):

lowiebenoot (approved):
it will not lock the scroll in projects that have a scrollable body I think. In our case (in focus-frontend) it's fine, because our body is not scrollable, but some other div is instead.

So actually we should still need the scroll locking like in the overlay as well.

driesd (approved):

farazatarodi:
@lowiebenoot if the body has scroll then the whole page (including the menu and the content) will scroll together. The point is that the menu sticks with the anchor element.

PR #2677, opened 2023-06-12
ci: [OPS-3327] move to github actions

MattiDeGrauwe:


env/production/config.json
```diff
@@ -0,0 +1,3 @@
+{
+  "env": "stg"
```
lowiebenoot, line 2: What is this for? And why is it stg?

MattiDeGrauwe: My bad, should be production, this provisions the environment on GitHub

lowiebenoot (approved):

PR #2710, opened 2023-07-11
Added labelPosition and fullWidth props to Toggle

qubis741:
### Added

`Toggle`: Props `labelPosition` and `fullWidth`. ([@qubis741](https://https://github.com/qubis741) in [#2710](https://github.com/teamleadercrm/ui/pull/2710))

### Design
https://www.figma.com/file/Aq9Ovc5xBe93MfuPUHTNW6/New-Quotations---designs?type=design&node-id=10451-283721&mode=design&t=hPowg87zssp1apqC-0

### DIFF
--- a/src/components/toggle/theme.css
+++ b/src/components/toggle/theme.css
@@ -29,10 +29,11 @@
     overflow: hidden;
     opacity: 0;
     margin: 0;
+    position: absolute;
   }
 
   .label {
-    margin-left: var(--spacer-small);
+    margin: 0 0 0 var(--spacer-small);
     display: flex;
   }
 
@@ -110,6 +111,18 @@
         background: var(--color-neutral-lightest);
       }
     }
+  }
+
+  &.label-on-left {
+    flex-direction: row-reverse;
+    .label {
+      margin: 0 var(--spacer-small) 0 0;
+    }
+  }
+
+  &.is-full-width {
+    justify-content: space-between;
+    width: 100%;
   }
 
   input:focus-visible + .track {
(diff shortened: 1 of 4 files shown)


src/components/toggle/theme.css
```diff
@@ -29,10 +29,11 @@
     overflow: hidden;
     opacity: 0;
     margin: 0;
+    position: absolute;
```
lorgan3, line 32: Q: shouldn't the toggle be `position: relative` then?

qubis741: input is hidden. I am setting it to `absolute` so it wont mess with `justify-content: space-between`

qubis741:
> Can we have a link to the design, please?

Added to description

src/components/toggle/theme.css
lorgan3: I see

lorgan3 (dismissed):

qubis741:
> Why not use the toggle with a `UIText` inside a `Flex`? All children of the flex container will stretch to full width by default. This is the implementation in the header customization menu of the new detail pages.

1. Its defined in Ahoy: https://ahoy.teamleader.design/3547946e7/p/812011-form-elements/b/40b1cb/t/281793
2. Its used atleast 3 times (project detail, quotation detail, now I need it in quote preview dialog)

farazatarodi:
Thanks for the clarification. I think we can link Ahoy documentation to our changes to speed up the process in the future.

farazatarodi (dismissed):

eniskraasniqi1 (approved):

farazatarodi (approved):

PR #2714, opened 2023-07-13
[FRONT-524] Upgrade storybook to version 7

lowiebenoot:
Upgraded storybook and addons to version 7.

There's still a warning about an addon that will probably be deprecated in the future, but couldn't get the changelog working without that plugin.


lowiebenoot:
Looking into the failing chromatic

lorgan3 (dismissed):

lorgan3:
Did a quick smoke test, seems to work 👍 
All chromatic stories are now visible in the root though

lowiebenoot:
> All chromatic stories are now visible in the root though

Ah didn't think of that.
Tried fixing it with an exclude pattern but that doesn't work because storybook prepends the patterns with `.storybook/` which breaks the pattern.
So fixed it by making the other pattern just more specific.

eniskraasniqi1 (approved):

qubis741 (approved):

jelledc (approved):

KristofColpaert (approved):

lorgan3 (approved):

PR #2788, opened 2023-10-18
`Select`: make sure there is only one select-overlay element

lowiebenoot:
When you run multiple instances of @teamleaer/ui, it will also make multiple select-overlay elements, which is giving us some issues in tests. We opted for only creating one, and never unmounting it instead.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,12 @@
 
 ### Dependency updates
 
+## [22.3.5] - 2023-10-18
+
+### Fixed
+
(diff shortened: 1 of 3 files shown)


farazatarodi (approved):

lorgan3 (approved):

driesd (approved):

eniskraasniqi1 (approved):

farazatarodi:
🕵️‍♂️ OTS MC

PR #2796, opened 2023-10-30
`Tooltip`: implement with Radix UI

lowiebenoot:
### Description

While trying to solve a bug in our tooltip positioning, I created another bug and realised that we would actually need to change and rewrite it completely in order to get it correct. Instead of implementing it ourself, we opted for using the tooltip from [Radix UI](https://www.radix-ui.com/). In the future we might use other Radix primitives as well for other components.

I still had to keep the hovering and active state inside our component (instead of letting radix handle it), because radix doesn't have the `tooltipHideOnClick`.

I removed the `horizontal` and `vertical` position, but it will still render to the opposite side in case there is not enough space on the side that you chose.

There is also no animation anymore when hiding the tooltip, because radix just removes it from the dom immediately.

### Breaking changes

`Tooltip`:  removed the `horizontal` and `vertical` position

### DIFF
--- a/src/components/tooltip/Tooltip.tsx
+++ b/src/components/tooltip/Tooltip.tsx
@@ -113,6 +113,12 @@ const TooltippedComponent: GenericComponent<TooltippedComponentProps> = ({
     }
   };
 
+  const handleOpenChange: RadixTooltip.TooltipProps['onOpenChange'] = (open) => {
+    if (open && onTooltipEntered) {
+      onTooltipEntered();
+    }
+  };
+
   useEffect(() => {
     if (tooltipActive && !active) {
       setActive(true);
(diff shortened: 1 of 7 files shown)


src/components/tooltip/Tooltip.tsx
```diff
@@ -231,47 +150,44 @@ const TooltippedComponent: GenericComponent<TooltippedComponentProps> = ({
...
+  // Using the radix tooltip component, but we only use it for rendering the tooltip
+  // we still manually implement the trigger with mouseover/leave/click and keep that in state.
+  // With a pure radix implementation we couldn't support our `tooltipHideOnClick` prop.
+  return (
+    <RadixTooltip.Provider delayDuration={tooltipShowDelay}>
+      <RadixTooltip.Root
+        onOpenChange={(open) => {
+          if (open && onTooltipEntered) {
+            onTooltipEntered();
+          }
```
lorgan3, line 162: PP no inline functions if it can be avoided

lorgan3 (approved):

src/components/tooltip/Tooltip.tsx
```diff
@@ -88,75 +80,11 @@ const TooltippedComponent: GenericComponent<TooltippedComponentProps> = ({
   ...other
 }) => {
   // eslint-disable-next-line react-hooks/exhaustive-deps
```
farazatarodi, line 82: I think this can be removed.

farazatarodi (approved):

JorenSaeyTL (dismissed):

KristofColpaert (approved):

lorgan3 (approved):

farazatarodi (approved):

PR #2803, opened 2023-11-06
Forward ref components

lowiebenoot:
### Description

Since the update to the new tooltip we are getting some errors because the tooltipped component is not implenting ref forwarding. Due to this the tooltip can't add event handlers on the trigger and errors.

I added ref forwarding to almost all components. Some components are a bit more difficult to add ref forwarding and will not be tooltipped anyway, so they're not a problem at the moment.

### DIFF
--- a/src/components/grid/GridItem.tsx
+++ b/src/components/grid/GridItem.tsx
@@ -3,11 +3,17 @@ import { GenericComponent } from '../../@types/types';
 
 export type GridItemProps = Partial<{ children: ReactNode; area: string }>;
 
-const GridItem: GenericComponent<GridItemProps> = forwardRef<HTMLDivElement, GridItemProps>(({ children, area }) => {
-  const gridItemStyles = { gridArea: area };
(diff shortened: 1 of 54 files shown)


src/components/grid/GridItem.tsx
```diff
@@ -1,12 +1,14 @@
-import React, { ReactNode } from 'react';
+import React, { ReactNode, forwardRef } from 'react';
 import { GenericComponent } from '../../@types/types';
 
 export type GridItemProps = Partial<{ children: ReactNode; area: string }>;
 
-const GridItem: GenericComponent<GridItemProps> = ({ children, area }) => {
+const GridItem: GenericComponent<GridItemProps> = forwardRef<HTMLDivElement, GridItemProps>(({ children, area }) => {
```
lorgan3, line 6: Looks like you forgot actually forwarding the ref here 👀

lorgan3 (approved):

qubis741 (approved):

driesd (approved):

PR #2824, opened 2023-12-05
Change element of select remove button

lorgan3:
### Description

Rendering a select in a label element causes the first remove button to get triggered when clicking on said label.
As a solution we'll make the remove button just a div (equivalent with what it was before)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,14 @@
 
 ### Removed
 
+### Fixed
+
+## [25.0.1]
+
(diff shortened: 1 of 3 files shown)


qubis741 (approved):

eniskraasniqi1:
Manual checked 🦅 ✅

- Removing one random tag doesn't remove the first one anymore

jelledc (approved):

farazatarodi (approved):

PR #2837, opened 2024-01-08
[FRONT-626] Fix popover positioning in Safari when zoomed by pinching

lowiebenoot:
### Description

Popover positioning is incorrect in Safari when you zoomed by pinching on the trackpad (or on iOS).

I found that floating-ui had the same [issue](https://github.com/floating-ui/floating-ui/pull/1099/files#r418947428) a long time ago and they solved it by checking the user agent. It's not great, but there's no other solution I'm afraid (Safari just does it different/wrong  🤷‍♂️).

I tried replacing our popover with the Radix-ui popover, but radix doesn't allow you to disable the focus trap on the popover, which we need to fully support our current popover.

### Manual check

To manual check this, you can't just open the popover story because that runs in an iframe. You need to open the actual iframe to test it out correctly. For example: http://localhost:3000/iframe.html?viewMode=docs&id=mid-level-blocks-popover--docs

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,8 @@
 
 ### Fixed
 
+- `Popover` and `Menu`: Fixed positioning in Safari when zoomed in by pinching ([@lowiebenoot](https://github.com/lowiebenoot)) in [#2837](https://github.com/teamleadercrm/ui/pull/2837)
+
 ## [25.0.3]
 
 ### Fixed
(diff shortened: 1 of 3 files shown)


lorgan3 (dismissed):

qubis741 (dismissed):

farazatarodi (dismissed):

jelledc (dismissed):

lowiebenoot:
Saw that we had the same issue for the Menu component, so fixed it there as well.

lorgan3 (approved):

farazatarodi (approved):

